PR #19, opened 2015-07-08
Require Java 7

ijuma:
Since Kafka will also require Java 7.

This opens the door to upgrading Jersey and Jetty and perhaps getting rid of JettyTestContainer.


ewencp:
add to whitelist

PR #21, opened 2015-07-14
Upgrade Jetty, Jersey, Jackson and JUnit to the latest released version

ijuma:
Jetty 9.2 includes a number of incompatible changes that we had to adapt
to.

Use `JettyTestContainer` and `JettyTestContainerFactory` from Jersey and
delete our copy (which needed some changes to work with Jetty 9.2).

This is still a WIP because we have lost the functionality from `MetricsSelectChannelConnector`. The way to do it in Jetty 9.2 seems to be very different. Example from `metrics`:

https://github.com/dropwizard/metrics/tree/master/metrics-jetty9/src/main/java/com/codahale/metrics/jetty9

@ewencp, shall we emulate what `metrics` does or do you have a better alternative? It's also worth noting that `StatisticsHandler` is required for graceful shutdown and it tracks a number of stats:

https://github.com/eclipse/jetty.project/blob/jetty-9.2.x/jetty-server/src/main/java/org/eclipse/jetty/server/handler/StatisticsHandler.java


ewencp:
@ijuma re: MetricsSelectChannelConnector, I think Jetty renamed/consolidated, but it seems like we should probably still be able to do something very similar. I think in Jetty 9 the different connector classes are just provided by a single ServerConnector class (probably because the nio support they need is available in all supported JDK versions now). I think what we probably want is to use NetworkTrafficServerConnector, which lets you hook a NetworkTrafficListener in, which gets notified of connection-level events.

The metrics approach doesn't seem directly related since it tracks different types of stats -- they are request-level rather than connection-level. That said, they may be tracking more info than we are collecting -- we have just the basic set here: http://docs.confluent.io/1.0/kafka-rest/docs/monitoring.html Tracking stats about error codes, for instance, might be a nice addition in a separate patch.

ijuma:
Good point about `NetworkTrafficListener`. I pushed a new version that uses it. It's still WIP though because it's not clear what to map to `accepted`. Documentation for `NetworkListener` (which confusingly says that `opened` is invoked when the connection has been `accepted`):

```
    /**
     * <p>Callback method invoked when a connection from a remote client has been accepted.</p>
     * <p>The {@code socket} parameter can be used to extract socket address information of
     * the remote client.</p>
     *
     * @param socket the socket associated with the remote client
     */
    public void opened(Socket socket);

    /**
     * <p>Callback method invoked when bytes sent by a remote client arrived on the server.</p>
     *
     * @param socket the socket associated with the remote client
     * @param bytes  the read-only buffer containing the incoming bytes
     */
    public void incoming(Socket socket, ByteBuffer bytes);

    /**
     * <p>Callback method invoked when bytes are sent to a remote client from the server.</p>
     * <p>This method is invoked after the bytes have been actually written to the remote client.</p>
     *
     * @param socket the socket associated with the remote client
     * @param bytes  the read-only buffer containing the outgoing bytes
     */
    public void outgoing(Socket socket, ByteBuffer bytes);

    /**
     * <p>Callback method invoked when a connection to a remote client has been closed.</p>
     * <p>The {@code socket} parameter is already closed when this method is called, so it
     * cannot be queried for socket address information of the remote client.<br />
     * However, the {@code socket} parameter is the same object passed to {@link #opened(Socket)},
     * so it is possible to map socket information in {@link #opened(Socket)} and retrieve it
     * in this method.
     *
     * @param socket the (closed) socket associated with the remote client
     */
    public void closed(Socket socket);
```

ewencp:
@ijuma Yeah, I think they just don't expose quite the same level of detail. In practice, `connections-accepted` and `connections-opened` are not really that different -- one is just a result of calling `accept()`, which is just waiting for the connection, `opened` is when the TCP connection between the client and server actually starts. These two events may be separated significantly in time, but in a given time window they should differ by at most 1. I exposed `accepted` mainly because it was included with the connection events -- maybe I should have just omitted it since the actual connections are what's more useful.

Since we don't have the same info anymore, maybe we can stay compatible by just using the `opened` rate for both since they are basically equivalent anyway. And then we could mark the metric deprecated (and should update corresponding docs, although those will be in the confluentinc/docs repo).

ijuma:
Did it as you suggested @ewencp. This PR is ready for review. Will do a separate PR in `confluentinc/docs`.

ijuma:
Created a PR for the documentation update. Note that it's actually in `confluentinc/kafka-rest/docs` instead of `confluentinc/docs`.

ewencp:
@ijuma LGTM, thanks for doing the update and cleanup!

Sorry for misdirecting you on the documentation. Unfortunately the same docs are duplicated in schema-registry because we need this documented everywhere we use rest-utils. Mind doing a quick patch there too? Feel free to just push directly instead of doing a PR since it'll be basically identical to the kafka-rest one.

ijuma:
Thanks @ewencp. No problem at all about the docs location, you were very helpful. :) I will push the change to schema-registry too.

PR #22, opened 2015-07-22
swagger documentation prototype

jendap:
DO NOT MERGE!!! at least not yet!

This is a prototype for adding [swagger.io](http://swagger.io) - apache license annotation driven documentation framework. See the other pull request confluentinc/kafka-rest#102

Run both from this branch and go to [localhost:8082/doc](http://localhost:8082/doc) and try it out! Let me know if you like it or not. If you like it I can clean / finish up this pull request. If not feel free to close it.

Also note that this thing can generate decent clients in many languages. Not a client library one would like to support for a long time but good enough for many folks in different languages to lower their barrier for the time being. If you want to try that out try:
```bash
git clone https://github.com/swagger-api/swagger-codegen.git
cd swagger-codegen
mvn package
java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i http://localhost:8082/swagger.json -l java -o kafka-rest-java
```

Note - do not waste much time reviewing the code - this is not intended to merge yet. It is not properly structured, the other pull request is not really exposing all the api, the webJarsResourceHandler is not nice, the beanConfig has everything hardcoded (even "localhost:8082"), etc. All those easy to fix. But it does not make sense to waste time on it if there would be no desire to merge it.

What do you think?


ewencp:
@jendap For something like this we need to figure out how it would fit in with the rest of our documentation and whether it's worth the cost of adding this extra tool (and the complexity of building). Based on what you mentioned already, some pros and cons:

Pros:
* Keeps up to date automatically with code, at least up to basic descriptions, API arguments and types, etc.
* Can generate temporary API clients for languages that don't already have custom wrapper libraries.

Cons:
* Requires some work to get it to fit in with the rest of the documentation (build process, styling, representing it all as a unified set of docs rather than regular docs vs api docs, etc).
* No example API requests? I'm having trouble figuring out if this is possible or not -- I didn't see any in the examples and the spec is difficult to navigate without any TOC (and "example" and "sample" show up a lot in other contexts). The ability to run your own examples in-browser is nice, but not a replacement for real examples and is oriented towards APIs that are provided as a service (e.g., we wouldn't have a good way to limit usage by API key since we don't have API keys).

jendap:
Thank you for taking a look at it!

Most of the concerns are about modifying the index.html - add custom css for style, add custom navigation into index.html.

But most importantly it seems that we should align our expectations here. A startup guy and kafka user would benefit from being able to show other teams what kafka is and how it can be used. Generate some ruby or python client to get data flowing to kafka. Unified look and company colours are usually a concern of that company ;-)

Those two interests can easily coexist independently - open source project may have something nerds will find useful even if it is not ready for enterprise sales pitch yet.



Answers:

**Examples** Great question! Where to store them? What I have done in the past was to save them as individual json files in the repo. File src/main/resources/examples/produce-avro.json:
```json
{
  "method": "POST",
  "path": "/topics/avrotest",
  "headers": {
    "Content-Type": "application/vnd.kafka.avro.v1+json"
  },
  "body": {"value_schema": "{\"type\": \"record\", \"name\": \"User\", \"fields\": [{\"name\": \"name\", \"type\": \"string\"}]}", "records": [{"value": {"name": "testUser"}}]}
}
```
Use "examples" folder...
1) Expose it as a static resource on /docs/examples to use by a few lines custom javascript in index.html
2) Run unit tests on it! No broken examples. Plus when you see in code review anybody touching these files you double check it for api breakage!

Hook up "/docs/examples"...
1) Modify the TOC / navigation menu in html to contain links to javascript action using those example urls. A few lines of javascript can then take the the static json example and fill it into the form.
2) Another idea - add an annotation in upstream project - `@ApiExample({"/examples/produce-avro.json"})`. Return it in swagger.json. The swagger-ui component could render a drop down box with examples per api call. Click on it would call javascript to fetch the example.json and fill in the fields. This functionality does not exists yet but I will try to add it.

**Rate limiting** Is that an issue? Forget the api key. That text field can be removed from index.html. Why would anybody want to rate limit browser differently?

**Building documentation** Two ideas for professional product:
1) [swagger-maven-plugin](https://github.com/kongchen/swagger-maven-plugin/), [swagger2markup](https://github.com/Swagger2Markup/swagger2markup) or [something else](http://swagger.io/open-source-integrations/) may be useful.
2) Alternatively - dump swagger.json in integration test to a file. Use maven-dependency-plugin to unzip the swagger-ui.jar, add the custom and css and index.html and the swagger.json. Done.


**Future** I will put here a commit (soon) to show how to do those things. I will try the "@ApiExample" idea upstream.

jendap:
@ewencp check out both repositories ...

It is updated with (very few) examples. And some left hand navigation and some colours. Well, I'm not a front end guy so it's really not great but it does the job.

Serious documentation one day can be, imho, done using https://github.com/redowl/swagger2markup-maven-plugin

How about that?

ewencp:
@jendap Sorry for sitting on this for so long. For now we've decided to stick with our current approach to documenting APIs -- while something automated would be great, the current approach integrates very well with the rest of our documentation which makes it much simpler to maintain.

jendap:
Sure.

btw: the current documentation is pretty nice :+1:

PR #23, opened 2015-09-16
added new config parameter to enable Cross-origin resource sharing

chizar:
Hi guys,

Have added new conf parameter to have ability to use Jetty's org.eclipse.jetty.servlets.CrossOriginFilter filter. 

The filter enables Cross-origin resource sharing (https://en.wikipedia.org/wiki/Cross-origin_resource_sharing), which is very convenient if I want to use kafka-rest api directly from client browser's javascript code. 

Please review.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -59,6 +59,11 @@
             <version>${jetty.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-servlets</artifactId>
+            <version>${jetty.version}</version>
+        </dependency>
+        <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
             <version>${jackson.version}</version>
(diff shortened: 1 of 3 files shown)


ewencp:
ok to test

ewencp:
@chizar This looks good to me. 2 quick questions.

First, is this functionality unit testable? Right now there's nothing in the automated tests that validates that this works.

Second, I noticed that the filter is on `/*`. Is there any good use case for more granular control. I'd rather get all the flexibility we need in now rather than having to support multiple configs.

chizar:
1. It is possible to check is filter really added to context if property is defined.But not sure that it is so critical. If you insist - will add test for Application class.

2. I could not imagine any case where user want to change filter mapping.

ewencp:
@chizar Ok, thanks! I think there could possibly be cases where people want more control over some settings (for instance, restricting to GET instead of the default GET and POST would permit read-only access to the cluster). But I think we can leave those improvements to a future patch since this is really useful as-is for providing internal web apps that utilize REST proxy functionality.

PR #25, opened 2015-11-04
Use x.y.z versioning scheme (i.e. 2.0.0-SNAPSHOT)

miguno:


miguno:
Note: The Jenkins build above fails because the associated PR for the versioning change of `common` isn't merged yet: https://github.com/confluentinc/common/pull/23

PR #29, opened 2016-01-06
Patching for Issue 143

lukemao:
Based on requirements specified in KAFKA Rest API (http://docs.confluent.io/1.0.1/kafka-rest/docs/api.html#consumers):

`Request JSON Array of Objects:
    records – A list of records to produce to the topic.
    records[i].key (object) – The message key, formatted according to the embedded format, or null to omit a key (optional)
    records[i].value (object) – The message value, formatted according to the embedded format
    records[i].partition (int) – Partition to store the message in (optional)
`

We can not provide duplicated keys when producing messages. 
I updated the `rest-utils` to restrict number of key to be provided in message producing.

### DIFF
--- a/core/src/main/java/io/confluent/rest/validation/JacksonMessageBodyProvider.java
+++ b/core/src/main/java/io/confluent/rest/validation/JacksonMessageBodyProvider.java
@@ -15,12 +15,14 @@
  */
 package io.confluent.rest.validation;
 
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
 import com.fasterxml.jackson.jaxrs.json.JacksonJaxbJsonProvider;
 
 import javax.validation.ConstraintViolationException;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.ext.Provider;
@@ -40,10 +42,13 @@ import java.lang.reflect.Type;
 public class JacksonMessageBodyProvider extends JacksonJaxbJsonProvider {
 
   public JacksonMessageBodyProvider() {
-    setMapper(new ObjectMapper());
+	ObjectMapper mapper = new ObjectMapper();
+	mapper.enable(DeserializationFeature.FAIL_ON_READING_DUP_TREE_KEY);
+    setMapper(mapper);
   }
 
   public JacksonMessageBodyProvider(ObjectMapper mapper) {
+	mapper.enable(DeserializationFeature.FAIL_ON_READING_DUP_TREE_KEY);
     setMapper(mapper);
   }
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/confluent/rest/validation/JacksonMessageBodyProvider.java
```diff
@@ -40,10 +42,13 @@
...
 
   public JacksonMessageBodyProvider() {
-    setMapper(new ObjectMapper());
+	ObjectMapper mapper = new ObjectMapper();
+	mapper.enable(DeserializationFeature.FAIL_ON_READING_DUP_TREE_KEY);
+    setMapper(mapper);
   }
 
   public JacksonMessageBodyProvider(ObjectMapper mapper) {
+	mapper.enable(DeserializationFeature.FAIL_ON_READING_DUP_TREE_KEY);
```
ewencp: Forcibly overriding this seems to be counter to the goal of this constructor, which is to allow the user to get control over the exact settings of the `ObjectMapper`. Is there any reason this needs to be overridden here? If the caller chose to leave `FAIL_ON_READING_DUP_TREE_KEY` off, is there any reason not to respect that?

This might require adjustments to kafka-rest to address the issue that triggered this PR, but keep in mind this class is used across a few projects and some might not need to be as restrictive and might prefer being a bit more liberal with the input they accept.

core/src/main/java/io/confluent/rest/validation/JacksonMessageBodyProvider.java
```diff
@@ -77,6 +82,8 @@ public Object readFrom(Class<Object> type,
       Throwable cause = e.getCause();
       if (cause instanceof ConstraintViolationException) {
         throw (ConstraintViolationException) cause;
+      }else if(cause instanceof WebApplicationException){
```
ewencp: Why do we need to extract this type of exception now? And do we . Without a test that triggers the specific error, I'm having trouble understanding why we would get a `WebApplicationException` as the cause of the `JsonMappingException` when a key is duplicated. I wouldn't have expected it since `WebApplicationException` is a Jersey error, not a Jackson error, but maybe I'm missing something about the exception handling in this case?

I know I mentioned `WebApplicationException` in the kafka-rest issue, but this wasn't quite what I had in mind. I just meant that somewhere in this `catch` block we might need to handle the specific exception thrown in response to this type of error and convert it to the right type (e.g. that's how invalid constraints found during bean validation are handled by extracting the `ConstraintViolationException`, which is in turn handled elsewhere by an ExceptionMapper to convert it to the appropriate `WebApplicationException`).

core/src/main/java/io/confluent/rest/validation/JacksonMessageBodyProvider.java
lukemao: Sorry I was thinking having a single key per record to be produced is the requirement, as defined in the API reference page. Key should be uniquely identify a record. However, I think other REST function may also use this to parse their request, which may allow duplicated meta-data. I will change it back.

core/src/main/java/io/confluent/rest/validation/JacksonMessageBodyProvider.java
lukemao: Sorry I mis-understood your comment. The online Jackson doc specified that FAIL_ON_READING_DUP_TREE_KEY may trigger the JsonMappingException only. They did not mention specific cause type.

lukemao:
Just made another commit. Also changed the `Application` class to use the default constructor in `JacksonMessageBodyProvider`.

ewencp:
Can we provide a test that this actually behaves as expected? I'm not sure, but maybe it could fit into `ExceptionHandlingTest` since this should trigger an exception during parsing of the entity?

PR #34, opened 2016-04-11
allow cors on any HTTP verb available.

norwood:
by default CrossOriginFilter will only allow cors for GET/POST/HEAD. 

we will now allow users to config a comma separated list of VERBs via `access.control.allowed.methods`


ewencp:
LGTM There's basically equivalent code in kafka for Connect's REST API since we couldn't use rest-utils there. Are you porting this there or should we file a bug for someone else to follow up on it? (There are even unit tests over there, so this could actually be validated properly.)

norwood:
sure. i can add this there too.

PR #38, opened 2016-05-10
Supporting HTTPS.

alexlod:
My selfish reason for this PR is to support https in both the schema registry and REST proxy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,12 @@ See the [`examples/`](examples/) directory for a simple demo service.
 
 Security
 --------
-The REST `Application` can support both http and https. https is disabled by default. Three modes are supported:
-http, https, and http+https. In the last mode, http and https are supported on different ports.
+The REST `Application` can support both http and https. https is disabled by default. Use the `listeners` configuration
+parameter to control which protocol is used.
 
 The https implementation is similar to Kafka, where the server specifies a keystore and trust store. When SSL
 client auth is configured, the client must authenticate -- the key in the client's keystore must be trusted by
-the server's trust store, either because the client cert is in the server's trust store, or because the CA that signed
-the client's cert is in the server's trust store.
+the server's trust store.
 
 Metrics
 -------
--- a/README.md
+++ b/README.md
@@ -6,6 +6,23 @@ REST APIs using Jersey, Jackson, Jetty, and Hibernate Validator.
 
 See the [`examples/`](examples/) directory for a simple demo service.
 
+Security
+--------
+The REST `Application` can support both http and https. https is disabled by default. Three modes are supported:
+http, https, and http+https. In the last mode, http and https are supported on different ports.
+
+The https implementation is similar to Kafka, where the server specifies a keystore and trust store. When SSL
+client auth is configured, the client must authenticate -- the key in the client's keystore must be trusted by
+the server's trust store, either because the client cert is in the server's trust store, or because the CA that signed
+the client's cert is in the server's trust store.
+
+Metrics
+-------
+When both http and https protocols are enabled, two metrics groups are created, `jetty-metrics` and
+`jetty-https-metrics`. However, `jetty-https-metrics` is a duplicate of `jetty-metrics`. Put differently, each
+metrics group will have the same value for all metrics, and those values will be influenced by both http and
+https requests.
+
 Contribute
 ----------
 
--- a/README.md
+++ b/README.md
@@ -17,10 +17,8 @@ the server's trust store.
 
 Metrics
 -------
-When both http and https protocols are enabled, two metrics groups are created, `jetty-metrics` and
-`jetty-https-metrics`. However, `jetty-https-metrics` is a duplicate of `jetty-metrics`. Put differently, each
-metrics group will have the same value for all metrics, and those values will be influenced by both http and
-https requests.
+When both http and https and enabled, metrics are not separated per protocol. Meaning, when observing metrics, the
+values of each metric are a combination of both http and https.
 
 Contribute
 ----------
--- a/README.md
+++ b/README.md
@@ -17,8 +17,10 @@ the server's trust store.
 
 Metrics
 -------
-When both http and https and enabled, metrics are not separated per protocol. Meaning, when observing metrics, the
-values of each metric are a combination of both http and https.
+One metrics group exists that spans all listeners. Meaning, when using multiple listeners and observing metrics, the
+values of each metric are a combination of all listeners. For example, if the maximum latency for a http listener
+is 50ms and the maximum latency for a https listener is 100ms, the `request-latency-max` metric will be 100ms, the max
+across all listeners.
 
 Contribute
 ----------
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -18,6 +18,7 @@ package io.confluent.rest;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.jaxrs.base.JsonParseExceptionMapper;
 
+import io.confluent.common.config.ConfigException;
 import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.NetworkTrafficServerConnector;
 import org.eclipse.jetty.server.Server;
@@ -36,6 +37,9 @@ import org.glassfish.jersey.server.validation.ValidationFeature;
 import org.glassfish.jersey.servlet.ServletContainer;
 
 import java.util.EnumSet;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -125,76 +129,73 @@ public abstract class Application<T extends RestConfig> {
       }
     };
 
-    // http
-    if (config.getString(RestConfig.REST_PROTOCOL_CONFIG).equals(RestConfig.REST_PROTOCOL_HTTP) ||
-        config.getString(RestConfig.REST_PROTOCOL_CONFIG).equals(RestConfig.REST_PROTOCOL_HTTP_PLUS_HTTPS)) {
-      NetworkTrafficServerConnector connector = new NetworkTrafficServerConnector(server);
-      connector.addNetworkTrafficListener(new MetricsListener(metrics, "jetty", metricTags));
-      connector.setPort(getConfiguration().getInt(RestConfig.PORT_CONFIG));
-      server.addConnector(connector);
-      log.info("Added http connector.");
-    }
+    List<URI> listeners = parseListeners(config.getString(RestConfig.LISTENERS_CONFIG),
+            config.getInt(RestConfig.PORT_CONFIG));
+    for (URI listener : listeners) {
+      log.info("Adding listener: " + listener.toString());
 
-    // https
-    if (config.getString(RestConfig.REST_PROTOCOL_CONFIG).equals(RestConfig.REST_PROTOCOL_HTTPS) ||
-        config.getString(RestConfig.REST_PROTOCOL_CONFIG).equals(RestConfig.REST_PROTOCOL_HTTP_PLUS_HTTPS)) {
-      SslContextFactory sslContextFactory = new SslContextFactory();
-      // IMPORTANT: the key's CN, stored in the keystore, must match the FQDN. This is a Jetty requirement.
-      if (!config.getString(RestConfig.SSL_KEYSTORE_LOCATION_CONFIG).isEmpty()) {
-        sslContextFactory.setKeyStorePath(config.getString(RestConfig.SSL_KEYSTORE_LOCATION_CONFIG));
-        sslContextFactory.setKeyStorePassword(config.getString(RestConfig.SSL_KEYSTORE_PASSWORD_CONFIG));
-        sslContextFactory.setKeyManagerPassword(config.getString(RestConfig.SSL_KEY_PASSWORD_CONFIG));
-        sslContextFactory.setKeyStoreType(config.getString(RestConfig.SSL_KEYSTORE_TYPE_CONFIG));
+      // http
+      if (listener.getScheme().equals("http")) {
+        NetworkTrafficServerConnector connector = new NetworkTrafficServerConnector(server);
+        connector.addNetworkTrafficListener(new MetricsListener(metrics, "jetty", metricTags));
+        connector.setPort(listener.getPort());
+        connector.setHost(listener.getHost());
+        server.addConnector(connector);
+      }
 
-        if (!config.getString(RestConfig.SSL_KEYMANAGER_ALGORITHM_CONFIG).isEmpty()) {
-          sslContextFactory.setSslKeyManagerFactoryAlgorithm(
-                  config.getString(RestConfig.SSL_KEYMANAGER_ALGORITHM_CONFIG));
+      // https
+      if (listener.getScheme().equals("https")) {
+        SslContextFactory sslContextFactory = new SslContextFactory();
+        // IMPORTANT: the key's CN, stored in the keystore, must match the FQDN. This is a Jetty requirement.
+        if (!config.getString(RestConfig.SSL_KEYSTORE_LOCATION_CONFIG).isEmpty()) {
+          sslContextFactory.setKeyStorePath(config.getString(RestConfig.SSL_KEYSTORE_LOCATION_CONFIG));
+          sslContextFactory.setKeyStorePassword(config.getString(RestConfig.SSL_KEYSTORE_PASSWORD_CONFIG));
+          sslContextFactory.setKeyManagerPassword(config.getString(RestConfig.SSL_KEY_PASSWORD_CONFIG));
+          sslContextFactory.setKeyStoreType(config.getString(RestConfig.SSL_KEYSTORE_TYPE_CONFIG));
+
+          if (!config.getString(RestConfig.SSL_KEYMANAGER_ALGORITHM_CONFIG).isEmpty()) {
+            sslContextFactory.setSslKeyManagerFactoryAlgorithm(
+                    config.getString(RestConfig.SSL_KEYMANAGER_ALGORITHM_CONFIG));
+          }
         }
-      }
 
-      sslContextFactory.setNeedClientAuth(config.getBoolean(RestConfig.SSL_CLIENT_AUTH_CONFIG));
+        sslContextFactory.setNeedClientAuth(config.getBoolean(RestConfig.SSL_CLIENT_AUTH_CONFIG));
 
-      if (!config.getString(RestConfig.SSL_ENABLED_PROTOCOLS_CONFIG).isEmpty()) {
-        sslContextFactory.setIncludeProtocols(config.getString(RestConfig.SSL_ENABLED_PROTOCOLS_CONFIG).split(","));
-      }
-
-      if (!config.getString(RestConfig.SSL_CIPHER_SUITES_CONFIG).isEmpty()) {
-        sslContextFactory.setIncludeCipherSuites(config.getString(RestConfig.SSL_CIPHER_SUITES_CONFIG).split(","));
-      }
-
-      if (!config.getString(RestConfig.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG).isEmpty()) {
-        sslContextFactory.setEndpointIdentificationAlgorithm(
-                config.getString(RestConfig.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG));
-      }
-
-      if (!config.getString(RestConfig.SSL_TRUSTSTORE_LOCATION_CONFIG).isEmpty()) {
-        sslContextFactory.setTrustStorePath(config.getString(RestConfig.SSL_TRUSTSTORE_LOCATION_CONFIG));
-        sslContextFactory.setTrustStorePassword(config.getString(RestConfig.SSL_TRUSTSTORE_PASSWORD_CONFIG));
-        sslContextFactory.setTrustStoreType(config.getString(RestConfig.SSL_TRUSTSTORE_TYPE_CONFIG));
-
-        if (!config.getString(RestConfig.SSL_TRUSTMANAGER_ALGORITHM_CONFIG).isEmpty()) {
-          sslContextFactory.setTrustManagerFactoryAlgorithm(
-                  config.getString(RestConfig.SSL_TRUSTMANAGER_ALGORITHM_CONFIG));
+        if (!config.getString(RestConfig.SSL_ENABLED_PROTOCOLS_CONFIG).isEmpty()) {
+          sslContextFactory.setIncludeProtocols(config.getString(RestConfig.SSL_ENABLED_PROTOCOLS_CONFIG).split(","));
         }
+
+        if (!config.getString(RestConfig.SSL_CIPHER_SUITES_CONFIG).isEmpty()) {
+          sslContextFactory.setIncludeCipherSuites(config.getString(RestConfig.SSL_CIPHER_SUITES_CONFIG).split(","));
+        }
+
+        if (!config.getString(RestConfig.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG).isEmpty()) {
+          sslContextFactory.setEndpointIdentificationAlgorithm(
+                  config.getString(RestConfig.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG));
+        }
+
+        if (!config.getString(RestConfig.SSL_TRUSTSTORE_LOCATION_CONFIG).isEmpty()) {
+          sslContextFactory.setTrustStorePath(config.getString(RestConfig.SSL_TRUSTSTORE_LOCATION_CONFIG));
+          sslContextFactory.setTrustStorePassword(config.getString(RestConfig.SSL_TRUSTSTORE_PASSWORD_CONFIG));
+          sslContextFactory.setTrustStoreType(config.getString(RestConfig.SSL_TRUSTSTORE_TYPE_CONFIG));
+
+          if (!config.getString(RestConfig.SSL_TRUSTMANAGER_ALGORITHM_CONFIG).isEmpty()) {
+            sslContextFactory.setTrustManagerFactoryAlgorithm(
+                    config.getString(RestConfig.SSL_TRUSTMANAGER_ALGORITHM_CONFIG));
+          }
+        }
+
+        sslContextFactory.setProtocol(config.getString(RestConfig.SSL_PROTOCOL_CONFIG));
+        if (!config.getString(RestConfig.SSL_PROVIDER_CONFIG).isEmpty()) {
+          sslContextFactory.setProtocol(config.getString(RestConfig.SSL_PROVIDER_CONFIG));
+        }
+
+        NetworkTrafficServerConnector sslConnector = new NetworkTrafficServerConnector(server, sslContextFactory);
+        sslConnector.addNetworkTrafficListener(new MetricsListener(metrics, "jetty-https", metricTags));
+        sslConnector.setPort(listener.getPort());
+        sslConnector.setHost(listener.getHost());
+        server.addConnector(sslConnector);
       }
-
-      sslContextFactory.setProtocol(config.getString(RestConfig.SSL_PROTOCOL_CONFIG));
-      if (!config.getString(RestConfig.SSL_PROVIDER_CONFIG).isEmpty()) {
-        sslContextFactory.setProtocol(config.getString(RestConfig.SSL_PROVIDER_CONFIG));
-      }
-
-      NetworkTrafficServerConnector sslConnector = new NetworkTrafficServerConnector(server, sslContextFactory);
-      sslConnector.addNetworkTrafficListener(new MetricsListener(metrics, "jetty-https", metricTags));
-      sslConnector.setPort(getConfiguration().getInt(RestConfig.PORT_HTTPS_CONFIG));
-      server.addConnector(sslConnector);
-      log.info("Added https connector.");
-    }
-
-    if (server.getConnectors().length < 1) {
-      log.error("No connectors were added to Jetty, meaning http and https are disabled. " +
-               "This is likely a configuration error. " + RestConfig.REST_PROTOCOL_CONFIG +
-               " is likely not configured correctly. Current value: " +
-               config.getString(RestConfig.REST_PROTOCOL_CONFIG));
     }
 
     ServletContextHandler context = new ServletContextHandler(ServletContextHandler.SESSIONS);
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -6,6 +6,23 @@ REST APIs using Jersey, Jackson, Jetty, and Hibernate Validator.
 
 See the [`examples/`](examples/) directory for a simple demo service.
 
+Security
+--------
+The REST `Application` can support both http and https. https is disabled by default. Three modes are supported:
+http, https, and http+https. In the last mode, http and https are supported on different ports.
```
ewencp: Still scanning through the rest of the patch, but an initial thought: Kafka switched to a `listeners` strategy for specifying what protocol and port it would listen on. Would this make sense here as well? I would be worried about the potential for combinatoric explosion in the set of config values we'd need to check here if we add more variants (e.g. HTTP Basic Auth). Admittedly it's probably not that bad as we can always parse these (in fact, we could convert this setting to a `LIST` type instead and just separate with `,`). But then once you start adding in customization of the port being listened on, etc., it seems like it could get messier than just using the `listeners` config approach.

README.md
```diff
@@ -6,6 +6,23 @@ REST APIs using Jersey, Jackson, Jetty, and Hibernate Validator.
...
+The https implementation is similar to Kafka, where the server specifies a keystore and trust store. When SSL
+client auth is configured, the client must authenticate -- the key in the client's keystore must be trusted by
+the server's trust store, either because the client cert is in the server's trust store, or because the CA that signed
+the client's cert is in the server's trust store.
+
+Metrics
+-------
+When both http and https protocols are enabled, two metrics groups are created, `jetty-metrics` and
+`jetty-https-metrics`. However, `jetty-https-metrics` is a duplicate of `jetty-metrics`. Put differently, each
+metrics group will have the same value for all metrics, and those values will be influenced by both http and
```
ewencp: Hmm, not sure I understand the purpose of doing this? If they are identical, couldn't we just stick with one? I would get the second group if we split the accounting of traffic between them, but this setup seems... wasteful?

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -121,10 +125,77 @@ protected void doStop() throws Exception {
...
+        }
+      }
+
+      sslContextFactory.setProtocol(config.getString(RestConfig.SSL_PROTOCOL_CONFIG));
+      if (!config.getString(RestConfig.SSL_PROVIDER_CONFIG).isEmpty()) {
+        sslContextFactory.setProtocol(config.getString(RestConfig.SSL_PROVIDER_CONFIG));
+      }
+
+      NetworkTrafficServerConnector sslConnector = new NetworkTrafficServerConnector(server, sslContextFactory);
+      sslConnector.addNetworkTrafficListener(new MetricsListener(metrics, "jetty-https", metricTags));
```
ewencp: Hmm, now I am confused by the earlier comment re: duplicating the metrics. This looks like it does generate a new set of metrics dedicated to this `NetworkTrafficServerConnector`? Are the two sets of stats actually identical? I haven't dug into the details of jetty, but this looks like they should be independent even if you use `http+https`.

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -93,13 +107,80 @@
       "<code>MetricReporter</code> interface allows plugging in classes that will be notified " +
       "of new metric creation. The JmxReporter is always included to register JMX statistics.";
   protected static final String METRICS_REPORTER_CLASSES_DEFAULT = "";
+  public static final String SSL_KEYSTORE_LOCATION_CONFIG = "ssl.keystore.location";
```
ewencp: Out of curiosity (and maybe @ijuma has a good idea here as well), how likely is it that we'd need to add any more configs here? I am looking at this list and wondering how much maintenance pain this could end up being between Kafka, this, the schema-registry kafkastore-prefixed versions, et al. In this repo, we don't depend on anything Kafka-related except for the tests you're adding here, but if there's a chance we need to add stuff, update docstrings, etc, we might want to figure out a strategy to manage all that info centrally and then provide it to the N different projects that we're now adding SSL support to.

core/src/test/java/io/confluent/rest/SslTest.java
```diff
@@ -0,0 +1,327 @@
...
+      // create a new client cert that isn't in the server's trust store.
+      File untrustedClient = File.createTempFile("SslTest-client-keystore", ".jks");
+      Map<String, X509Certificate> certs = new HashMap<>();
+      createKeystoreWithCert(untrustedClient, "client", certs);
+
+      try {
+        makeGetRequestAndVerify200("https://localhost:" + (RestConfig.PORT_HTTPS_CONFIG_DEFAULT) + "/test",
+                untrustedClient.getAbsolutePath(), SSL_PASSWORD, SSL_PASSWORD);
+      } catch (SSLHandshakeException she) {
+        // expected.
```
ewencp: You can also specify expected exceptions in the `@Test` annotation so you don't need the try/catch block. I'm not sure whether we want to do this here or not given that a subsequent exception is also expected, but figured I'd mention it if it would simplify the test code.

core/src/test/java/io/confluent/rest/SslTest.java
```diff
@@ -0,0 +1,327 @@
...
+    configServerKeystore(props);
+    configServerTruststore(props);
+    enableSslClientAuth(props);
+    TestRestConfig config = new TestRestConfig(props);
+    SslTestApplication app = new SslTestApplication(config);
+    try {
+      app.start();
+
+      try {
+        makeGetRequestAndVerify200("https://localhost:" + (RestConfig.PORT_HTTPS_CONFIG_DEFAULT) + "/test");
```
ewencp: I realize you're just reusing code, but these calls might be confusing in tests that are expected to fail -- the name implies you'll check the 200 result, but in fact you won't since it throws an exception first, right?

core/src/test/resources/log4j.properties
```diff
@@ -14,5 +14,8 @@
 # limitations under the License.
 ##
 
-log4j.rootLogger=OFF
+log4j.rootLogger=INFO, stdout
```
ewencp: Is this just for debugging? I'm not sure why we added this change.

ewencp:
@alexlod Took a first pass, this looks great! Some questions re: how we want to do configuration. I'm also not the best filter for security-related stuff -- maybe if @ijuma has time he can scan through as well since he looked at the schema-registry patch, or maybe he can call in someone else to take a pass.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -121,10 +125,77 @@ protected void doStop() throws Exception {
...
+      connector.setPort(getConfiguration().getInt(RestConfig.PORT_CONFIG));
+      server.addConnector(connector);
+      log.info("Added http connector.");
+    }
+
+    // https
+    if (config.getString(RestConfig.REST_PROTOCOL_CONFIG).equals(RestConfig.REST_PROTOCOL_HTTPS) ||
+        config.getString(RestConfig.REST_PROTOCOL_CONFIG).equals(RestConfig.REST_PROTOCOL_HTTP_PLUS_HTTPS)) {
+      SslContextFactory sslContextFactory = new SslContextFactory();
+      // IMPORTANT: the key's CN, stored in the keystore, must match the FQDN. This is a Jetty requirement.
```
ijuma: Doesn't Jetty support `AltSubjectNames`? I would be surprised if it doesn't. We have been telling people about the CN and FQDN for Kafka too, but it turns out that this is actually no longer the recommended way. See https://issues.apache.org/jira/browse/KAFKA-3665?focusedCommentId=15274001&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15274001 for more details.

README.md
alexlod: The way rest-utils is currently implemented, it does not specify which host to listen on, so all network interfaces are listened on. But Jetty does support listening on specific interfaces via `NetworkTrafficServerConnector.setHost()`. I like the `listeners` model. Going this route will also let the end-user listen on specific interfaces. I'll make this change -- good suggestion, @ewencp!

README.md
alexlod: Currently, `io.confluent.common.metrics` does not support registering multiple metrics listeners with the same name. As such, there are a few options:

 1. My current implementation, where two different groups exist, but they have identical values.
 2. Change `io.confluent.common.metrics` to support adding multiple listeners with the same group name.
 3. Use a different group name, and different metrics tags, which means https and http would have different values.

I chose (1) because if I were operating a rest app, I would want to see the same metrics for http and https, yet I thought the change to `io.confluent.common.metrics` may be problematic because it would the current semantics quite a bit. Although, thinking about this more, I may actually prefer http and https values to be different, so get more monitoring granularity.

I agree (1) is odd and am happy to consider one of the other alternatives. I'm leaning towards option (3). I'd love to hear your thoughts on options (2) and (3).

core/src/main/java/io/confluent/rest/Application.java
alexlod: They are actually identical because they use the same underlying metrics tags LinkedList. I have a test ensuring that they're identical.

core/src/main/java/io/confluent/rest/RestConfig.java
alexlod: I'm interested to hear @ijuma's opinion, too.

I agree that having some sort of factory to make security more convenient is a fabulous idea. And seems feasible in theory, although I haven't tried it.

The switch to Apache's `ConfigDef` will simply this a lot because we can use `SslConfigs.addClientSslSupport()` and `SaslConfigs.addClientSaslSupport()`.

core/src/test/java/io/confluent/rest/SslTest.java
alexlod: Correct. I'll change this to be less confusing. Good suggestion!

core/src/test/java/io/confluent/rest/SslTest.java
alexlod: Initially I thought this wouldn't simplify the code. But with your suggestion below about the http get helper, I think it will. I'll go this route.

core/src/test/resources/log4j.properties
alexlod: I used it for debugging, yes. However, I thought enabling logging in general would be a good thing. I'm happy to change it back if you disagree.

core/src/main/java/io/confluent/rest/RestConfig.java
ijuma: I expect the number of configs to continue to grow as we flesh out support for things like certificate revocation. The number shouldn't be large, but having to update multiple different places may be a bit error-prone.

alexlod: @ijuma, do you know of any issues we may encounter if we made special helper methods for creating security configurations (with a prefix, eg in the schema registry, `kafkastore.ssl.*`)? Thinking about this more, I think this will be straightforward. Let me know if you disagree. Thanks!

core/src/main/java/io/confluent/rest/Application.java
alexlod: In my test, Jetty doesn't support SubjectAltNames. Here's the exception:

    Exception in thread "main" javax.net.ssl.SSLPeerUnverifiedException: Host name 'localhost' does not match the certificate subject provided by the peer (CN=Alex Loddengaard, OU=Services, O=Confluent, L=Palo Alto, ST=California, C=US)
    	at org.apache.http.conn.ssl.SSLConnectionSocketFactory.verifyHostname(SSLConnectionSocketFactory.java:465)
    	at org.apache.http.conn.ssl.SSLConnectionSocketFactory.createLayeredSocket(SSLConnectionSocketFactory.java:395)
    	at org.apache.http.conn.ssl.SSLConnectionSocketFactory.connectSocket(SSLConnectionSocketFactory.java:353)
    	at org.apache.http.impl.conn.DefaultHttpClientConnectionOperator.connect(DefaultHttpClientConnectionOperator.java:141)
    	at org.apache.http.impl.conn.PoolingHttpClientConnectionManager.connect(PoolingHttpClientConnectionManager.java:353)
    	at org.apache.http.impl.execchain.MainClientExec.establishRoute(MainClientExec.java:380)
    	at org.apache.http.impl.execchain.MainClientExec.execute(MainClientExec.java:236)
    	at org.apache.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:184)
    	at org.apache.http.impl.execchain.RetryExec.execute(RetryExec.java:88)
    	at org.apache.http.impl.execchain.RedirectExec.execute(RedirectExec.java:110)
    	at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:184)
    	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:82)
    	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:107)
    	at TestSchemaRegistryHttps.makeGetRequest(TestSchemaRegistryHttps.java:49)
    	at TestSchemaRegistryHttps.main(TestSchemaRegistryHttps.java:17)
    	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    	at java.lang.reflect.Method.invoke(Method.java:497)
    	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:144)

I generated the keystore with `-ext san=dns:localhost`.

I've had trouble finding this in the jetty docs to confirm.

alexlod:
@ewencp and @ijuma: I just pushed two new commits that address all your comments. Although some comments in this PR are still open and may or may not need to be addressed before we get this in.

core/src/main/java/io/confluent/rest/Application.java
alexlod: Actually, I may have generated the cert incorrectly. There may be a nasty bug with `keytool` where certain command-line args aren't forwarded using `genkey` instead of `genkeypair`. I'm investigating.

alexlod: I didn't generate the wrong cert. I got confused because the cert I generated had the right `san` set, but when the cert was signed by the CA, `ext` info is no longer printed to the screen. I'm having trouble finding a definitive answer, but I'm pretty sure the CA signing shouldn't remove the `san`.

alexlod:
I can't reproduce the failed test in my local environment. How can I check what JDK version Jenkins is on?

ewencp:
@alexlod It's JDK7. Do you need more specific version info than that? We test everything with the lowest version we support currently.

alexlod:
Thanks @ewencp. Don't need a more specific version number. I'll add another patch to get tests to pass on JDK7.

README.md
alexlod: @ewencp, I just pushed a new commit that still combines http and https metrics (I think this is right) but doesn't use separate group names. Now there's just one group. This is strictly better than my previous implementation, although perhaps we should discuss if combining metrics is the right approach or not.

(Also, note that the newest push will make this comment hidden.)

core/src/main/java/io/confluent/rest/Application.java
ijuma: OK, maybe file an issue to investigate this more later? It should not prevent this PR from getting in, but we are probably going to start recommending the usage of `SubjectAltNames` for Kafka and it would be good if the same approach worked across the platform. Jetty is used in Google Cloud and Google makes extensive use of `SubjectAltNames` for their `google.com` so it seems a bit weird that it would not be supported.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -121,10 +125,77 @@ protected void doStop() throws Exception {
...
+      if (!config.getString(RestConfig.SSL_KEYSTORE_LOCATION_CONFIG).isEmpty()) {
+        sslContextFactory.setKeyStorePath(config.getString(RestConfig.SSL_KEYSTORE_LOCATION_CONFIG));
+        sslContextFactory.setKeyStorePassword(config.getString(RestConfig.SSL_KEYSTORE_PASSWORD_CONFIG));
+        sslContextFactory.setKeyManagerPassword(config.getString(RestConfig.SSL_KEY_PASSWORD_CONFIG));
+        sslContextFactory.setKeyStoreType(config.getString(RestConfig.SSL_KEYSTORE_TYPE_CONFIG));
```
ijuma: Do we need to validate that the 3 parameters above are actually set?

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -121,10 +125,77 @@ protected void doStop() throws Exception {
...
+        if (!config.getString(RestConfig.SSL_KEYMANAGER_ALGORITHM_CONFIG).isEmpty()) {
+          sslContextFactory.setSslKeyManagerFactoryAlgorithm(
+                  config.getString(RestConfig.SSL_KEYMANAGER_ALGORITHM_CONFIG));
+        }
+      }
+
+      sslContextFactory.setNeedClientAuth(config.getBoolean(RestConfig.SSL_CLIENT_AUTH_CONFIG));
+
+      if (!config.getString(RestConfig.SSL_ENABLED_PROTOCOLS_CONFIG).isEmpty()) {
+        sslContextFactory.setIncludeProtocols(config.getString(RestConfig.SSL_ENABLED_PROTOCOLS_CONFIG).split(","));
```
ijuma: The forked config doesn't have `getList`?

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -121,10 +125,77 @@ protected void doStop() throws Exception {
...
+      sslConnector.setPort(getConfiguration().getInt(RestConfig.PORT_HTTPS_CONFIG));
+      server.addConnector(sslConnector);
+      log.info("Added https connector.");
+    }
+
+    if (server.getConnectors().length < 1) {
+      log.error("No connectors were added to Jetty, meaning http and https are disabled. " +
+               "This is likely a configuration error. " + RestConfig.REST_PROTOCOL_CONFIG +
+               " is likely not configured correctly. Current value: " +
+               config.getString(RestConfig.REST_PROTOCOL_CONFIG));
```
ijuma: We log at error level, but don't throw an exception?

ijuma: Normally, errors result in an exception, otherwise warn can be used. But maybe there's a good reason to do it differently here.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -121,10 +125,77 @@ protected void doStop() throws Exception {
...
+      }
+
+      NetworkTrafficServerConnector sslConnector = new NetworkTrafficServerConnector(server, sslContextFactory);
+      sslConnector.addNetworkTrafficListener(new MetricsListener(metrics, "jetty-https", metricTags));
+      sslConnector.setPort(getConfiguration().getInt(RestConfig.PORT_HTTPS_CONFIG));
+      server.addConnector(sslConnector);
+      log.info("Added https connector.");
+    }
+
+    if (server.getConnectors().length < 1) {
```
ijuma: Nitpick: isn't it clearer to say `length == 0`? Array lengths can never be negative after all.

core/src/test/java/io/confluent/rest/SslTest.java
```diff
@@ -0,0 +1,327 @@
...
+
+      httpclient = HttpClients.custom()
+              .setSSLSocketFactory(sslSf)
+              .build();
+    }
+
+    CloseableHttpResponse response = httpclient.execute(httpget);
+    assertEquals("Response status must be 200.", 200, response.getStatusLine().getStatusCode());
+    response.close();
+    httpclient.close();
```
ijuma: You probably want to have a finally block to ensure `response` and `httpclient` are closed if an exception is thrown

core/src/test/resources/log4j.properties
ijuma: You don't want tests to be too noisy when running them in your IDE, for example. `OFF` seems a bit extreme, but the right level for a given project varies so I'll defer to @ewencp as he knows the project better.

ewencp: I don't have a strong opinion here. I wouldn't worry much in this repo anyway -- rest-utils doesn't see a lot of changes these days.

alexlod: I suggest we leave it. I agree that noisy tests are difficult to parse, but these tests aren't too noisy and I find the output to be helpful when debugging.

core/src/main/java/io/confluent/rest/Application.java
alexlod: `SSL_KEYSTORE_TYPE_CONFIG` has a default, so that one doesn't need it. The passwords shouldn't be validated because they may actually be blank. And the location is already being validated in the `if` statement. So I think we're good. Unless I'm missing something?

core/src/main/java/io/confluent/rest/Application.java
alexlod: `ConfigDef` in Confluent common does not have getList. `AbstractConfig` does but it isn't implemented correctly (it doesn't `split()`).

core/src/main/java/io/confluent/rest/Application.java
alexlod: Good catch. An exception is better. My newest commit, with the listener logic, does this.

core/src/main/java/io/confluent/rest/Application.java
alexlod: Agreed! Good suggestion. My listener commit addresses this.

core/src/main/java/io/confluent/rest/Application.java
alexlod: Added a TODO.

alexlod:
@ijuma, thanks for your comments -- all great and helpful. However, you made them on an outdated diff. The newest diff addresses all your comments. However, it introduces a fairly big change -- the introduction of the listener config. The SSL code is not affected by this change, so perhaps @ewencp would be a better reviewer for the listener change?

core/src/main/java/io/confluent/rest/Application.java
ijuma: OK, thanks for explaining.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -168,6 +236,50 @@ protected void doStop() throws Exception {
     return server;
   }
 
+  private static String[] getList(String original) {
```
ewencp: Don't we have a `LIST` type in ConfigDef? Can we use that instead and get rid of this custom parsing code?

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -168,6 +236,50 @@ protected void doStop() throws Exception {
...
+    String[] parts = original.split(",");
+    for (int i = 0; i < parts.length; i++) {
+      parts[i] = parts[i].trim();
+    }
+    return parts;
+  }
+
+  // TODO: delete deprecatedPort parameter when `PORT_CONFIG` is deprecated. It's only used to support the deprecated
```
ewencp: Interesting, is this deprecation going to be logged anywhere yet? We'll definitely need to update any documentation for any projects that use rest-utils to indicate this transition would need to be made. If we're not logging anything yet, that might be fine too; maybe just suggest that using `listeners` is preferred, but `port` is still supported for compatibility.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -168,6 +236,50 @@ protected void doStop() throws Exception {
...
+  //       configuration.
+  static List<URI> parseListeners(String listenersStr, int deprecatedPort) {
+    String[] parts = getList(listenersStr);
+
+    // handle deprecated case, using PORT_CONFIG.
+    // TODO: remove this when `PORT_CONFIG` is deprecated, because LISTENER_CONFIG will have a default value which
+    //       includes the default port.
+    if (parts.length == 0 || parts[0].isEmpty()) {
```
ewencp: Related to above comment about `LIST` -- I can't remember the exact semantics, it might filter empty segments.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -168,6 +236,50 @@ protected void doStop() throws Exception {
...
+      parts = new String[] {"http://0.0.0.0:" + deprecatedPort};
```
ewencp: Do you need `0.0.0.0`? I suppose it is more explicit, although I don't really know if the preferred approach is to bind all interfaces, bind the FQDN, or maybe something else. Do you remember what Kafka does here? I think it doesn't even include an explicit IP, although maybe that's just a convenience it provides in its configs.

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -33,9 +33,16 @@
...
-  protected static final String PORT_CONFIG_DOC = "Port to listen on for new connections.";
+  protected static final String PORT_CONFIG_DOC = "DEPRECATED: port to listen on for new HTTP connections. Use " +
+      "listeners instead.";
   protected static final int PORT_CONFIG_DEFAULT = 8080;
 
+  public static final String LISTENERS_CONFIG = "listeners";
+  protected static final String LISTENERS_DOC = "List of listeners. http and https are supported. Each " +
+      "listener must include the protocol, hostname, and port. For example: http://myhost:8080," +
+      "https://0.0.0.0:8081";
+  protected static final String LISTENERS_DEFAULT = ""; // TODO: add a default value when `PORT_CONFIG` is deprecated.
```
ewencp: presumably you mean when it is removed since that is when it is safe? The alternative here is to change the default value of `port` immediately to remove the default and make `listeners` default to `http://:8080` and make any appearance of `port` force usage of the old approach. Not sure either approach is better -- only benefit I see to the alternative is that the new default is introduced immediately.

ewencp:
@alexlod Some questions, but the `listeners` updates look good to me. Note that if this gets merged we'll have config updates to be made on at least three downstream projects (I think possibly in the ConfigDef itself w/ overrides depending on what happens w/ default values, in the automatically generated config listing, and just general doc updates related to enabling SSL -- although we've not included the latter for CORS support, but I suspect SSL is in higher demand).

core/src/main/java/io/confluent/rest/Application.java
ijuma: In Kafka, we bind in the following way:

```scala
val socketAddress =
      if(host == null || host.trim.isEmpty)
        new InetSocketAddress(port)
      else
        new InetSocketAddress(host, port)
    val serverChannel = ServerSocketChannel.open()
    serverChannel.configureBlocking(false)
    serverChannel.socket().setReceiveBufferSize(recvBufferSize)
    try {
      serverChannel.socket.bind(socketAddress)
```

If the host is not present, we basically end up using:

```java
    /*
     * Returns the InetAddress representing anyLocalAddress
     * (typically 0.0.0.0 or ::0)
     */
    static InetAddress anyLocalAddress() {
        return impl.anyLocalAddress();
    }
```

When it comes to what we register to ZK though, we do:

```scala
if (endpoint.host == null || endpoint.host.trim.isEmpty)
  EndPoint(InetAddress.getLocalHost.getCanonicalHostName, endpoint.port, endpoint.protocolType)
```

So we use the `

core/src/main/java/io/confluent/rest/Application.java
alexlod: @ijuma suggested this earlier, too. I looked at the code for `getList` and it looked wrong to me. However, I just tested it and it works as expected. So I'll move to the `LIST` type.

core/src/main/java/io/confluent/rest/Application.java
alexlod: I think we should add a log. Will add one.

core/src/main/java/io/confluent/rest/Application.java
alexlod: @ijuma: thanks for the super detailed response!

Looks like Kafka binds to 0.0.0.0. Also, this is Jetty's default behavior when a host isn't provided (which is the current behavior). So I'm in favor of sticking with 0.0.0.0 as the default.

core/src/main/java/io/confluent/rest/RestConfig.java
alexlod: Yes, "deleted" is correct, not "deprecated."

As for your alternative approach, I think this will complicate the code, because I'll need to check if a port is configured, and if so, use that port instead of the port specified in the default listener. Unless I'm misunderstanding your suggestion? If not, I prefer my current approach. But thanks for your suggestion.

alexlod:
@ewencp: great suggestions! I addressed all of them in a new commit. Thanks.

core/src/main/java/io/confluent/rest/RestConfig.java
ijuma: No, I am not aware of any issues.

alexlod: Thanks, @ijuma! @ewencp, I suggest we move forward with this patch as is and we can introduce a helper to include security features later.

alexlod:
@ewencp, I made a new PR that has a single rebased commit, and is against the `3.x` branch. See here:

https://github.com/confluentinc/rest-utils/pull/40

alexlod:
Closing this in favor of https://github.com/confluentinc/rest-utils/pull/40 which is the same diff as this but in one commit and against the `3.x` branch.

AwesomeBobX64:
@ewencp It appears there is TLS support between the Schema Registry and Kafka, but not between the REST Proxy and Kafka. Would it be possible to add this soon?

ewencp:
@AwesomeBobX64 On the roadmap, but since all the new security support is contingent on using the new consumer the REST proxy will need to add new consumer support first. After that it should be a relatively small patch if added globally; if you need per-client support, things get more complicated. In any case, @alexlod is on the case!

AwesomeBobX64:
Maybe I can help add the new consumer support?

ewencp:
@AwesomeBobX64 Would be great if you could help contribute. I think the discussion is going on [here](https://github.com/confluentinc/kafka-rest/issues/204) and [here](https://github.com/confluentinc/kafka-rest/pull/216) although I think we determined it would make sense to start from a comprehensive API design and then work on implementation. I think @alexlod  can help fill in more details here -- we've been sketching out some API ideas, but not sure if anything has been formalized yet.

alexlod:
@AwesomeBobX64: we're working on a new REST Proxy consumer API for the new Java consumer. Once we have a spec for that, which will hopefully be soon, @GRpro plans to modify https://github.com/confluentinc/kafka-rest/pull/216 to both implement this new API and include the new consumer. Once that's done, we'll work on supporting TLS and other security features.

AwesomeBobX64:
That sounds awesome. In the meantime we're going to try out a workaround we came up with using our loopback interface.

PR #40, opened 2016-05-18
https support.

alexlod:
This is a copy of https://github.com/confluentinc/rest-utils/pull/38 with two changes: all commits rebased into one; made against the `3.x` branch instead of `master`.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -83,11 +83,37 @@
             <artifactId>jackson-annotations</artifactId>
             <version>${jackson.version}</version>
         </dependency>
-
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>${junit.version}</version>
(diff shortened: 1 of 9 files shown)


core/pom.xml
```diff
@@ -83,12 +83,38 @@
...
             <version>${jackson.version}</version>
         </dependency>
-
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>${junit.version}</version>
             <scope>test</scope>
         </dependency>
+        <!-- Kafka's SSL utilities are used is tests. -->
```
granders: Minor - typo: "is tests" -> "in tests"?

alexlod: Nice catch! I'll address this here (in a rebased commit) and in https://github.com/confluentinc/rest-utils/pull/38 in a new commit.

PR #44, opened 2016-06-29
add jaasy configs to rest-utils

norwood:


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -199,6 +210,25 @@ protected void doStop() throws Exception {
     ServletContextHandler context = new ServletContextHandler(ServletContextHandler.SESSIONS);
     context.setContextPath("/");
     context.addServlet(servletHolder, "/*");
+    String authMethod = getConfiguration().getString(RestConfig.AUTHENTICATION_METHOD);
+    if (!"NONE".equals(authMethod)) {
```
ewencp: Check known values and throw if its not one of the expected ones.

Also, is JAASLoginService the only one we want to support? If not, we'd need an additional config to switch on.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -199,6 +210,25 @@ protected void doStop() throws Exception {
...
+      List<String> roles = getConfiguration().getList(RestConfig.AUTHENTICATION_ROLES);
+      final ConstraintSecurityHandler securityHandler = new ConstraintSecurityHandler();
+      Constraint constraint = new Constraint();
+      constraint.setAuthenticate(true);
+      constraint.setRoles(roles.toArray(new String[roles.size()]));
+      ConstraintMapping constraintMapping = new ConstraintMapping();
+      constraintMapping.setConstraint(constraint);
+      constraintMapping.setMethod("*");
+      constraintMapping.setPathSpec("/*");
+      securityHandler.addConstraintMapping(constraintMapping);
```
ewencp: Won't most apps end up needing multiple constraint mappings? I'm not sure how we'd want to express this (or if there is a way to do it with annotations on the resource methods), but even in something like schema registry you might want read-only vs write access depending on the method type. In REST proxy when we have admin commands you'd want to lock those paths down with a different set of roles.

core/src/main/java/io/confluent/rest/Application.java
alexlod: SSL auth is currently supported via `ssl.client.auth` and applied with this line of code: `sslContextFactory.setNeedClientAuth(config.getBoolean(RestConfig.SSL_CLIENT_AUTH_CONFIG));`. This could be modified to be a part of an auth abstraction.

core/src/main/java/io/confluent/rest/Application.java
alexlod: I've had two customers request API ACL support in the schema registry. Perhaps what @norwood has here is a good starting point -- a single app-wide ACL. Then later we can make this pluggable so REST instances (eg Schema Registry) can define more fine-grain ACLs.

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -163,6 +164,13 @@
       "Leave blank to use Jetty's default.";
   protected static final String SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_DEFAULT = "";
 
+  public static final String AUTHENTICATION_METHOD = "authentication.method";
```
alexlod: Can you append `_CONFIG` to each of these config options to have consistent naming with other configuration parameters?

alexlod:
@norwood: looks good to me. Although writing tests will likely be annoying. Let me know if I can help when you get there -- I wrote SASL tests in the Schema Registry. Then, when you have tests, it'd be good to have @ijuma review.

core/src/main/java/io/confluent/rest/Application.java
ewencp: Yeah, I just want to make sure a) we don't get ourselves into an annoying compatibility mess and b) if we're going to express this via configs, that we *can* express this via configs. e.g. if we needed to annotate each resource handler and use introduce some additional info in the constraint somehow (e.g. "admin" vs "producer" vs "consumer" type resources in the REST proxy).

norwood: the way rest-utils does programatic configuration means we don't get the benefit(?) of the jetty xml configability: http://www.eclipse.org/jetty/documentation/9.4.x/configuring-security-authentication.html#_scoping_security_realms

to what @alexlod said, i was thinking of this as barebones what we need now and once we need per-path auth we would fight that battle

core/src/main/java/io/confluent/rest/Application.java
norwood: i think we can get away with just supporting jaas. you can use [jaas](http://www.eclipse.org/jetty/documentation/9.3.x/jaas-support.html#_a_closer_look_at_jaasloginservice) to config anything else with it (maybe not spnego).

i'll add a check for basic. the other possible value here would be `FORM`, but figuring out configs for that will be a pain equivalent to the constraint mappings http://www.eclipse.org/jetty/documentation/9.4.x/configuring-security-authentication.html#_configuring_an_authentication_mechanism

core/src/main/java/io/confluent/rest/Application.java
ewencp: @norwood If we can figure out how to shoehorn in XML configs as well, I think that's fine. There's certainly a ton of example configs out there that users would benefit from if we could handle arbitrary Jetty configs.

norwood: ok. i think we need to think that through though. it also probably has implications for everywhere else we config stuff for `/*` throughout this file.

one other option re: auth*, would be to only support authentication across `/*`, and then make users handle authorization themselves. right now if you allow all roles from your authentication source in, then you can just inject the `SecurityContext` in your endpoint and filter a role. (or do something like http://stackoverflow.com/questions/26777083/best-practice-for-rest-token-based-authentication-with-jax-rs-and-jersey)

ewencp: To follow up on offline discussion, I think this works fine -- use this to provide auth to the entire set of resources, but then have each resource manage authentication on its own. I think we want to remove the `AUTHENTICATION_ROLES_CONFIG` instead allowing this to be passed into the relevant methods (which I guess need to be made non-static). This allows the applications using these classes to figure out how to compute the union of all roles they want to allow into any resource, pass that in, then within each handler deal with adding the `SecurityContext` and filtering by the role appropriate for the handler on their own. This also allows them to figure out how best to express which resources/methods should be allowed for each role, e.g. adding higher level classifications rather than expressing them on a per-endpoint/method basis.

norwood:
@alexlod sorry for the delay. i'm not really sure what unit tests i could write for this other than making sure jetty is getting the right configs

ewencp:
@norwood I think @alexlod was referring to the integration tests used in schema reg that pulls up embedded zk/kafka/schema reg and starts making requests. There's an `EmbeddedServerTestHarness` in this repo that can help building tests like that.

alexlod:
What @ewencp said :)

norwood:
`EmbeddedServerTestHarness` doesn't ever run `Application.createServer`, so i'm not really following how to write a test for this.

norwood:
oh, i think i can maybe make something like what happens in `SslTest` work.

norwood:
retest this please

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -245,6 +259,28 @@ protected void doStop() throws Exception {
     return server;
   }
 
+  static boolean enableBasicAuth(String authMethod) {
+    RestConfig.AUTHENTICATION_METHOD_VALIDATOR.ensureValid(RestConfig.AUTHENTICATION_METHOD_CONFIG, authMethod);
```
ewencp: Why do we do this here? Shouldn't this have already been validated by `ConfigDef`/`AbstractConfig` during parsing?

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -21,7 +21,9 @@
 import io.confluent.common.config.ConfigDef.Importance;
 import io.confluent.common.utils.SystemTime;
 import io.confluent.common.utils.Time;
+import jersey.repackaged.com.google.common.collect.Lists;
```
ewencp: Do we need to rely on this? I'm actually not sure whether Jersey considers this public API? In particular, it looks like an `Arrays.asList` could work fine here? Or is this just because Guava folks know Guava and IntelliJ found the import for you?

core/src/test/java/io/confluent/rest/SaslTest.java
```diff
@@ -0,0 +1,262 @@
...
+                  + "};\n";
+    Files.write(
+        jaasFile.toPath(),
+        jaas.getBytes(),
+        StandardOpenOption.TRUNCATE_EXISTING
+    );
+
+    String loginProperties = "jay: kafka,Administrators\n"
+                             + "neha: akfak,Administrators\n"
+                             + "jun: kafka-\n";
```
ewencp: Not sure how @nehanarkhede is going to feel about being an Administrator of `akfak`...

ewencp:
@norwood Sorry, I think I was misleading because I have already forgotten what the test classes in `rest-utils` do. I had a couple of minor comments, but despite not really understanding this stuff, I believe the tests you added cover the necessary functionality. I'd wait for an LGTM from @alexlod/@ijuma, but LGTM

core/src/test/java/io/confluent/rest/SaslTest.java
norwood: actually akfak is her pw in this

core/src/main/java/io/confluent/rest/RestConfig.java
norwood: no. just living that tab-complete life

core/src/main/java/io/confluent/rest/Application.java
norwood: no. just a lazy refactor

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -163,6 +165,17 @@
       "Leave blank to use Jetty's default.";
   protected static final String SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_DEFAULT = "";
 
+  public static final String AUTHENTICATION_METHOD_CONFIG = "authentication.method";
+  public static final String AUTHENTICATION_METHOD_NONE = "NONE";
+  public static final String AUTHENTICATION_METHOD_BASIC = "BASIC";
+  public static final String AUTHENTICATION_METHOD_DOC = "Method of authentication. Must be BASIC to enable authentication. You must supply a valid JVM config file for java.security.auth.login.config for the appropriate authentication provider.";
+  public static final ConfigDef.ValidString AUTHENTICATION_METHOD_VALIDATOR = ConfigDef.ValidString.in(Arrays.asList(AUTHENTICATION_METHOD_NONE, AUTHENTICATION_METHOD_BASIC));
+  public static final String AUTHENTICATION_REALM_CONFIG = "authentication.realm";
+  public static final String AUTHENTICATION_REALM_DOC = "Securitry realm to be used in authentication.";
```
ijuma: Typo, `Security`

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -245,6 +259,27 @@ protected void doStop() throws Exception {
...
 
+  static boolean enableBasicAuth(String authMethod) {
+    return RestConfig.AUTHENTICATION_METHOD_BASIC.equals(authMethod);
+  }
+
+  static ConstraintSecurityHandler createSecurityHandler(String realm, List<String> roles) {
+    final ConstraintSecurityHandler securityHandler = new ConstraintSecurityHandler();
+    Constraint constraint = new Constraint();
+    constraint.setAuthenticate(true);
+    constraint.setRoles(roles.toArray(new String[roles.size()]));
```
ijuma: It turns out that passing `new String[0]` is more concise and faster (reference https://shipilev.net/blog/2016/arrays-wisdom-ancients/).

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -163,6 +165,17 @@
       "Leave blank to use Jetty's default.";
   protected static final String SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_DEFAULT = "";
 
+  public static final String AUTHENTICATION_METHOD_CONFIG = "authentication.method";
+  public static final String AUTHENTICATION_METHOD_NONE = "NONE";
+  public static final String AUTHENTICATION_METHOD_BASIC = "BASIC";
+  public static final String AUTHENTICATION_METHOD_DOC = "Method of authentication. Must be BASIC to enable authentication. You must supply a valid JVM config file for java.security.auth.login.config for the appropriate authentication provider.";
```
ijuma: Instead of JVM config, I think I would say JAAS config. And mention that `java.security.auth.login.config` is a system property.

core/src/test/java/io/confluent/rest/SaslTest.java
```diff
@@ -0,0 +1,262 @@
...
+    TestRestConfig config = new TestRestConfig(props);
+    app = new SaslTestApplication(config);
+    app.start();
+  }
+
+  @After
+  public void cleanup() throws Exception {
+    Configuration.setConfiguration(null);
+    if (previousAuthConfig != null) {
+      System.setProperty("java.security.auth.login.config", previousAuthConfig);
```
ijuma: Do we need this? Seems a bit weird for tests to rely on a system property set by another test.

core/src/test/java/io/confluent/rest/SaslTest.java
```diff
@@ -0,0 +1,262 @@
...
+    @Override
+    public void setupResources(Configurable<?> config, TestRestConfig appConfig) {
+      config.register(new SaslTestResource());
+    }
+
+    @Override
+    public Map<String, String> getMetricsTags() {
+      Map<String, String> tags = new LinkedHashMap<String, String>();
+      tags.put("instance-id", "1");
+      return tags;
```
ijuma: Nitpick: Collections.singletonMap is more concise.

ijuma:
Left some minor comments, seems fine otherwise (note that I have never used JAAS in the context of JAX-RS or Jetty).

core/src/test/java/io/confluent/rest/SaslTest.java
norwood: this is just undoing what we did in `setUp`, making sure this test has no side effects

ijuma: In Kafka, we typically just clear the system property instead of restoring it. We do it that way because it's simpler and no test should rely on that property being set already. A minor point, so you can leave it if you prefer it that way.

PR #45, opened 2016-07-22
Generalizing Application.parseListeners() to be used by the Schema Registry

alexlod:
This change is to support https://github.com/confluentinc/schema-registry/pull/371

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -37,12 +37,13 @@ import org.glassfish.jersey.server.validation.ValidationFeature;
 import org.glassfish.jersey.servlet.ServletContainer;
 
 import java.util.EnumSet;
-import java.net.URI;
-import java.net.URISyntaxException;
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -260,10 +262,14 @@ protected void doStop() throws Exception {
...
-      if (scheme != null && (scheme.equals("http") || scheme.equals("https"))) {
+      if (uri.getPort() == -1) {
+        throw new ConfigException("Found a listener without a port. All listeners must have a port. The " +
+          "listener without a port is: " + listenerStr);
+      }
+      if (scheme != null && supportedSchemes.contains(scheme)) {
         listeners.add(uri);
       } else {
-        log.warn("Found a listener with an unsupported scheme (ony http and https are supported). Ignoring " +
+        log.warn("Found a listener with an unsupported scheme (supported: " + supportedSchemes + "). Ignoring " +
```
ewencp: Should we be throwing a ConfigException if this else clause filters out all listeners?

ewencp:
@alexlod One minor question, but otherwise LGTM

core/src/main/java/io/confluent/rest/Application.java
alexlod: We are -- see line 277 just below this.

ewencp: Ah, apparently I cannot properly match braces.

alexlod: :). I'm gonna merge this thang.

PR #48, opened 2016-08-24
Remove unnecessary slf4j-log4j dependency that causes confusing warnings.

ewencp:


ewencp:
@miguno @ijuma @norwood @aayars @gwenshap same deal as confluentinc/common#35 (and depends on that to validate I believe since this may pull in the common deps indirectly). Probably still more downstream repos need changes (and for ones like schema-registry which include both libs and apps, it may get hairy), but the more of these we can remove from libraries, the better.

ijuma:
So, to be clear: you are not removing dependencies here, but adding them in test scope _because_ you removed them from common, right?

ewencp:
@ijuma Sort of right, some are ultimately removed because they seemed to be superfluous anyway. But for log4j you're correct, it's just moved to `test` scope so that logging works in tests (since they are standalone executables) but we don't need it for the library.

PR #51, opened 2016-09-19
we need more filters

xvrl:
@norwood @ewencp


ewencp:
@xvrl Not sure why that test error happened, but this addition seems fine to me.

PR #52, opened 2016-09-19
add support for gzip compression, fixes #18

### LINKED ISSUE #18: Add support for gzip compression
Embedded jetty should be able to configure gzip compression.

We might also want to figure out what other common features might need to be enabled as long as we are only embedding jetty and not running jetty directly as a servlet container with an xml config it can load to adjust things like this.

xvrl:


xvrl:
@norwood @ewencp

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -245,6 +244,14 @@ protected void doStop() throws Exception {
       context.setBaseResource(staticResources);
     }
 
+    if(config.getBoolean(RestConfig.ENABLE_GZIP_COMPRESSION_CONFIG)) {
```
ewencp: style nit: space after `if`

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -245,6 +244,14 @@ protected void doStop() throws Exception {
...
+      FilterHolder gzipFilter = new FilterHolder(AsyncGzipFilter.class);
+      // do not check if .gz file already exists for the requested resource
+      gzipFilter.setInitParameter("checkGzExists", "false");
```
ewencp: Is this never useful or just not important for the use case this is being added for? Presumably for static assets this can be useful to avoid re-compressing?

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -245,6 +244,14 @@ protected void doStop() throws Exception {
...
+      gzipFilter.setInitParameter("methods", "GET,POST");
+      context.addFilter(gzipFilter, "/*", null);
```
ewencp: Should there be a way to select what resources to apply compression to. More specifically, would it make sense for the config to be the resources to apply it to rather than a boolean?

core/src/main/java/io/confluent/rest/Application.java
xvrl: without this option the filter adds quite a bit of overhead checking for statically compressed resources. This is especially costly if serving static resources from within a jar file. Given that most of the requests are for dynamic content, it's better to turn if off

core/src/main/java/io/confluent/rest/Application.java
xvrl: I'm not sure if that level of granularity is necessary, do you have a particular use-case in mind that would warrant not turning it on? The only reason I can think of is if we served incompressible binary data from dynamic resources.

ewencp: Nothing specific, but since this covers everything it seems like it'd be potentially bad. For example, if you're serving static assets, gzipping a png is just going to be a waste of time. I had also been thinking of other already compressed data, e.g. in the REST proxy, but today the compressed form isn't exposed anyway, so maybe not worth worrying about that case.

Fine to leave it as is, was just calling out this as a potential limitation since it might change how we'd want to represent the config options.

xvrl: @ewencp static assets are already excluded based on mimeType so that shouldn't be a problem.

ewencp:
@xvrl Ok, seems like my issues were addressed -- LGTM

PR #57, opened 2016-12-20
bumping dependencies

norwood:
@ewencp


ewencp:
🕸 🚘

mhowlett:
thanks guys. i assume this was to fix the test failure i was looking into. was being slow 'cause I was using it as an exercise to get my head into the test framework. but seems there is plenty more opportunity for that coming up.

PR #59, opened 2017-02-28
rest-utils jackson version should match blueway at 2.8.4

cmccabe:


norwood:
i think you inadvertently have a commit from @aayars in this pr

PR #60, opened 2017-03-02
Pom madness

norwood:
dependent on: https://github.com/confluentinc/common/pull/43

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,10 +4,14 @@
 
     <modelVersion>4.0.0</modelVersion>
 
-    <groupId>io.confluent</groupId>
+    <parent>
+        <groupId>io.confluent</groupId>
+        <artifactId>common</artifactId>
+        <version>3.3.0-SNAPSHOT</version>
+    </parent>
+
     <artifactId>rest-utils-parent</artifactId>
     <packaging>pom</packaging>
-    <version>3.3.0-SNAPSHOT</version>
     <name>rest-utils-parent</name>
     <organization>
         <name>Confluent, Inc.</name>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -43,13 +47,11 @@
 
     <properties>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
-        <confluent.common.version>3.2.1-SNAPSHOT</confluent.common.version>
-        <slf4j.version>1.7.21</slf4j.version>
+        <confluent.common.version>${project.version}</confluent.common.version>
```
ewencp: Should we just get rid of this variable entirely since it's the same as the project version anyway, or do you still see value in having the level of indirection? Perhaps to make any subsequent transition to using separate version numbers for different components in the future?

I'm mildly embarrassed that just reusing the project version didn't occur to me. We'd probably significantly simplify the version bumping tooling by doing this across the board, at least as long as we stick with fully coordinated version numbers.

pom.xml
```diff
@@ -43,13 +47,11 @@
 
     <properties>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
-        <confluent.common.version>3.2.1-SNAPSHOT</confluent.common.version>
-        <slf4j.version>1.7.21</slf4j.version>
```
ewencp: Is this just the starting point? Or will other repos use a different parent pom (e.g. this one)? Things like jackson and the `kafka.scala.version` seem like they could be shared to avoid potential conflicts (e.g. differing jackson versions across connectors can be an issue).

pom.xml
```diff
@@ -64,28 +66,69 @@
...
+                <artifactId>jackson-annotations</artifactId>
+                <version>${jackson.version}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 
     <build>
         <plugins>
             <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
```
ewencp: Are build plugins inherited from the parent pom such that this is no longer required? This is a really important requirement for us so I don't want to mess it up (though I think our system tests would catch any issues since we're careful about only installing the earliest jdk we support).

test/pom.xml
```diff
@@ -24,7 +24,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>${junit.version}</version>
+            <scope>compile</scope>
```
ewencp: I don't understand why this change is needed? Isn't `compile` the default scope?

norwood: it is not the default since i moved the dep up to dependency manager in a parent pom. this needs to be reset to compile since test package is making a deployable for others to load

pom.xml
norwood: yeah, starting point. 

we could definitely have things like schema-registry and rest-proxy extend this pom so they can get jetty version (since i doubt anyone else needs that). and move jackson or something else shared up... though we can/will run in to conflicts there.

pom.xml
norwood: i think its fine to leave like this now. say at some point we move off of snapshot to independently versioned stuff

pom.xml
norwood: yes, here is the effective pom https://gist.github.com/norwood/16971fb2709d62997b75614cb1254bb4

norwood:
retest this please

norwood:
retest this please

PR #61, opened 2017-03-07
Code cleanup

norwood:


core/src/main/java/io/confluent/rest/exceptions/DebuggableExceptionMapper.java
```diff
@@ -27,6 +27,7 @@
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.PrintStream;
+import java.nio.charset.Charset;
```
xvrl: Doesn't seem to be used anywhere

core/src/test/java/io/confluent/rest/SaslTest.java
```diff
@@ -101,7 +102,7 @@ public void setUp() throws Exception {
                   + "};\n";
     Files.write(
         jaasFile.toPath(),
-        jaas.getBytes(),
+        jaas.getBytes(Charset.forName("UTF-8")),
```
xvrl: Might as well just use the string constant directly, using the charset has no benefit (see https://github.com/druid-io/druid/blob/master/java-util/src/main/java/io/druid/java/util/common/StringUtils.java)

core/src/test/java/io/confluent/rest/SaslTest.java
```diff
@@ -110,7 +111,7 @@ public void setUp() throws Exception {
                              + "jun: kafka-\n";
     Files.write(
         loginPropertiesFile.toPath(),
-        loginProperties.getBytes(),
+        loginProperties.getBytes(Charset.forName("UTF-8")),
```
xvrl: Same here

core/src/main/java/io/confluent/rest/exceptions/DebuggableExceptionMapper.java
```diff
@@ -59,11 +60,11 @@ public DebuggableExceptionMapper(RestConfig restConfig) {
       readableMessage += " " + exc.getClass().getName() + ": " + exc.getMessage();
       try {
         ByteArrayOutputStream os = new ByteArrayOutputStream();
-        PrintStream stream = new PrintStream(os);
+        PrintStream stream = new PrintStream(os, false, "UTF-8");
         exc.printStackTrace(stream);
         stream.close();
         os.close();
-        readableMessage += "\n" + os.toString();
+        readableMessage += "\n" + os.toString("UTF-8");
```
xvrl: Should we use the system line separator here while we're at it?

norwood: maybe? like `System.getProperty('line.separator')`?

xvrl: Just `System.lineSeparator()`

core/src/test/java/io/confluent/rest/SaslTest.java
```diff
@@ -16,55 +16,33 @@
 
 package io.confluent.rest;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
-
+import org.apache.commons.codec.Charsets;
```
ewencp: Use `StandardCharsets` that is built-in?

core/src/main/java/io/confluent/rest/exceptions/DebuggableExceptionMapper.java
```diff
@@ -59,11 +59,11 @@ public DebuggableExceptionMapper(RestConfig restConfig) {
       readableMessage += " " + exc.getClass().getName() + ": " + exc.getMessage();
       try {
         ByteArrayOutputStream os = new ByteArrayOutputStream();
-        PrintStream stream = new PrintStream(os);
+        PrintStream stream = new PrintStream(os, false, "UTF-8");
```
ewencp: I'm not that picky about it, but some people might prefer using `StandardCharsets.UTF_8.name()` to avoid repeating this string constant and making any issues w/ it a compile-time failure.

core/src/test/java/io/confluent/rest/StaticResourcesTest.java
```diff
@@ -46,10 +46,12 @@
 
   @Before
   public void setUp() throws Exception {
-    staticContent =
-        new BufferedReader(new InputStreamReader(
-            ClassLoader.getSystemResourceAsStream("static/index.html"), Charset.forName("utf-8")
-        )).readLine() + System.lineSeparator();
+    try (
+        InputStreamReader isr = new InputStreamReader(ClassLoader.getSystemResourceAsStream("static/index.html"), Charset.forName("utf-8"));
```
ewencp: We could replace `Charset.forName` here with one from `StandardCharsets`.

PR #62, opened 2017-03-20
Checkstyle

norwood:


norwood:
depends on https://github.com/confluentinc/common/pull/47

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -150,7 +158,7 @@ public Server createServer() throws RestConfigException {
 
     // Configure the servlet container
     ServletContainer servletContainer = new ServletContainer(resourceConfig);
-    FilterHolder servletHolder = new FilterHolder(servletContainer);
+    final FilterHolder servletHolder = new FilterHolder(servletContainer);
```
ewencp: Does this really complain about the `FilterHolder` but not the `ServletContainer` right above it? Or just unrelated cleanup?

norwood: it complains about this since `servletHolder` is not used near its declaration

norwood:
retest this please

PR #64, opened 2017-04-28
Decompose adding JSON provider into separate, protected method

C0urante:
Was running into an issue where an `ExceptionMapper` instance I'd registered for all `Throwable`s was getting overridden by the `GenericExceptionMapper` class that's registered in the `configureBaseApplication()` method. At the moment it's not preferable to just avoid calling `super.configureBaseApplication()` altogether, since I still want to register a body provider for JSON handling (current workaround is to not call `super.configureBaseApplication()` and just copy+paste the JSON body provider code into the subclass's `configureBaseApplication()` method). So, it'd be nice if there were a separate method for registering the body provider without also registering the `GenericExceptionMapper` instance or, for that matter, any unwanted `ExceptionMapper`s.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -395,7 +395,11 @@ public abstract class Application<T extends RestConfig> {
    * @param registerExceptionMapper Whether or not to register an additional exception mapper for
    *                                handling errors in (de)serialization
    */
-  protected void registerJsonProvider(Configurable<?> config, T restConfig, boolean registerExceptionMapper) {
+  protected void registerJsonProvider(
+      Configurable<?> config,
+      T restConfig,
+      boolean registerExceptionMapper
+  ) {
     ObjectMapper jsonMapper = getJsonMapper();
     JacksonMessageBodyProvider jsonProvider = new JacksonMessageBodyProvider(jsonMapper);
     config.register(jsonProvider);
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -375,9 +375,9 @@ public abstract class Application<T extends RestConfig> {
    * client.
    */
   public void configureBaseApplication(Configurable<?> config, Map<String, String> metricTags) {
-    RestConfig restConfig = getConfiguration();
-
     registerJsonProvider(config, true);
+
+    RestConfig restConfig = getConfiguration();
 
     config.register(ValidationFeature.class);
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -375,12 +375,9 @@ public void configureBaseApplication(Configurable<?> config) {
    * client.
    */
   public void configureBaseApplication(Configurable<?> config, Map<String, String> metricTags) {
-    RestConfig restConfig = getConfiguration();
+    registerJsonProvider(config, true);
```
ewencp: If we're going to provide methods for some of these bits, wouldn't it make sense to break out all of the substeps? e.g. `registerJsonProvider`, `registerFeatures`, `registerExceptionMappers`, etc?

C0urante: So something like this (but with Javadocs added)?

```
  public void configureBaseApplication(Configurable<?> config, Map<String, String> metricTags) {
    T restConfig = getConfiguration();

    registerJsonProvider(config, restConfig, true);
    registerFeatures(config, restConfig);
    registerExceptionMappers(config, restConfig);

    config.register(new MetricsResourceMethodApplicationListener(metrics, "jersey",
                                                                 metricTags, restConfig.getTime()));

    config.property(ServerProperties.BV_SEND_ERROR_IN_RESPONSE, true);
  }

  protected void registerJsonProvider(Configurable<?> config, T restConfig, boolean registerExceptionMapper) {
    ObjectMapper jsonMapper = getJsonMapper();
    JacksonMessageBodyProvider jsonProvider = new JacksonMessageBodyProvider(jsonMapper);
    config.register(jsonProvider);
    if (registerExceptionMapper) {
      config.register(JsonParseExceptionMapper.class);
    }
  }

  protected void registerFeatures(Configurable<?> config, T restConfig) {
    config.register(ValidationFeature.class);
  }

  protected void registerExceptionMappers(Configurable<?> config, T restConfig) {
    config.register(ConstraintViolationExceptionMapper.class);
    config.register(new WebApplicationExceptionMapper(restConfig));
    config.register(new GenericExceptionMapper(restConfig));
  }
```

ewencp: Yeah, that's what I had in mind. The exception mapper stuff split up a bit weirdly since the one for the `JsonProvider` is in `registerJsonProvider` instead of `registerExceptionMappers`, but otherwise breaking things down that way seems like a sane way to allow overriding the behavior without having to override everything.

PR #71, opened 2017-07-06
Upgrading jetty and jersey to the latest stable

The-Alchemist:
- small Jetty API changes
- jersey-hk2 is necessary because https://stackoverflow.com/a/44120415/423943
- SslTest fails because of a `javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure`,
  not sure why


The-Alchemist:
[clabot:check]

pom.xml
```diff
@@ -50,8 +50,8 @@
         <apache.httpclient.version>4.5.2</apache.httpclient.version>
         <bouncycastle.bcpkix.version>1.55</bouncycastle.bcpkix.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
-        <jersey.version>2.25</jersey.version>
-        <jetty.version>9.2.19.v20160908</jetty.version>
+        <jersey.version>2.26-b06</jersey.version>
+        <jetty.version>9.4.6.v20170531</jetty.version>
```
ewencp: We've held off on updating to 9.3+ because they remove Java 7 support and, unfortunately, there are quite a few folks still targeting JDK7. However, it looks like `9.2.22.v20170606` would be an updated version that would work here.

I'm also open to some maven profile magic that would allow people to build against newer versions if they're fine with a JDK8 requirement.

ewencp:
ok to test

pom.xml
The-Alchemist: Thanks, @ewencp .  Do you want separate Maven coordinates for JDK7 vs. JDK8 (`rest-utils-jdk7` and `rest-utils-jdk8`), or is a Maven classifier ok

I'm guessing the side effect of two different builds of `rest-util` is that there will be two builds of `kafka-rest`, the major downstream user of this lib.

So I believe there would be a similar build question for `kafka-rest`.

ewencp: @The-Alchemist Honestly not sure which would be preferable, I have generally tried to avoid multiple builds where possible. If we do a maven classifier, are we going to be able to get a default that will continue to work with jdk7? My biggest concern would be with compatibility -- it's not ideal if the maven coordinates change or downstream projects then also are required to add classifiers.

The-Alchemist: I understand... how about leaving the project as is for JDK7, and a `jdk8` classifier for JDK8?  (And the same for downstream projects.)

This way, nothing downstream has to change if they wanna keep using JDK7.

ewencp: @The-Alchemist yeah, that sounds totally reasonable to me. Will also hopefully make our lives easier when we can change the minimum JDK required!

core/pom.xml
```diff
@@ -96,5 +96,11 @@
             <classifier>test</classifier>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.glassfish.jersey.inject</groupId>
+            <artifactId>jersey-hk2</artifactId>
+            <scope>test</scope>
```
ijuma: Why is this only needed for tests?

core/src/test/java/io/confluent/rest/ApplicationTest.java
```diff
@@ -103,7 +102,7 @@ public void testAuthEnabledBASIC() {
 
   @Test
   public void testCreateSecurityHandlerWithNoRoles() {
-    ConstraintSecurityHandler securityHandler = Application.createSecurityHandler(REALM, Lists.<String>newArrayList());
+    ConstraintSecurityHandler securityHandler = Application.createSecurityHandler(REALM, new ArrayList<String>());
```
ijuma: Use `Collections.emptyList`?

core/src/test/java/io/confluent/rest/SslTest.java
```diff
@@ -84,7 +85,7 @@ public void setUp() throws Exception {
   private void createKeystoreWithCert(File file, String alias, Map<String, X509Certificate> certs) throws Exception {
     KeyPair keypair = TestSslUtils.generateKeyPair("RSA");
     // IMPORTANT: CN must be "localhost" because Jetty expects the server CN to be the FQDN.
-    X509Certificate cCert = TestSslUtils.generateCertificate("CN=localhost, O=A client", keypair, 30, "SHA1withRSA");
+    X509Certificate cCert = TestSslUtils.generateCertificate("CN=localhost, O=A client", keypair, 30, "SHA256WITHRSA");
```
ijuma: Why do we need this change?

core/src/test/java/io/confluent/rest/SslTest.java
```diff
@@ -276,10 +277,11 @@ private int makeGetRequest(String url, String clientKeystoreLocation, String cli
       SSLContext sslContext = sslContextBuilder.build();
 
       SSLConnectionSocketFactory sslSf = new SSLConnectionSocketFactory(sslContext, new String[]{"TLSv1"},
-              null, SSLConnectionSocketFactory.getDefaultHostnameVerifier());
+              null, NoopHostnameVerifier.INSTANCE);
```
ijuma: Why do we need this change?

core/src/test/java/io/confluent/rest/SslTest.java
```diff
@@ -276,10 +277,11 @@ private int makeGetRequest(String url, String clientKeystoreLocation, String cli
...
 
       httpclient = HttpClients.custom()
               .setSSLSocketFactory(sslSf)
+              .setSSLHostnameVerifier(NoopHostnameVerifier.INSTANCE)
```
ijuma: Why do we need this change?

pom.xml
```diff
@@ -50,8 +50,8 @@
         <apache.httpclient.version>4.5.2</apache.httpclient.version>
         <bouncycastle.bcpkix.version>1.55</bouncycastle.bcpkix.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
-        <jersey.version>2.25</jersey.version>
-        <jetty.version>9.2.19.v20160908</jetty.version>
+        <jersey.version>2.26-b06</jersey.version>
```
ijuma: 2.27 is out now.

pom.xml
ijuma: 9.4.10.v20180503 is now out.

ijuma:
@The-Alchemist, will you be able to update this PR within the next few days?

The-Alchemist:
Unfortunately not.  It's all in your hands.

IIRC, some of the changes had to do with version incompatibilities with
Java 8, and I could not get the SSL tests to pass, for some reason.

core/src/test/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListenerIntegrationTest.java
```diff
@@ -58,7 +58,6 @@ public void testListenerHandlesDispatchErrorsGracefully() {
         .request(MediaType.APPLICATION_JSON_TYPE)
         .get();
     assertEquals(500, response.getStatus());
-    assertTrue(response.readEntity(String.class).contains("Resource Java method invocation error"));
```
ijuma: What is the reason for removing this? Too brittle?

core/src/main/java/io/confluent/rest/logging/AbstractNCSARequestLog.java
```diff
@@ -156,7 +156,7 @@ public void log(Request request, Response response)
       buf.append("] \"");
       append(buf,request.getMethod());
       buf.append(' ');
-      append(buf, request.getUri().toString());
+      append(buf, request.getRequestURI());
```
ijuma: I was a bit unsure which of the methods `getOriginalURI` and `getRequestURI` is the one that replaced `getUri`.

core/src/test/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListenerIntegrationTest.java
The-Alchemist: I think so, but I can't remember, to be honest.

core/src/main/java/io/confluent/rest/logging/AbstractNCSARequestLog.java
The-Alchemist: IIRC, I checked the debugger to see what held the correct value, but it could use some more testing if the unit tests don't cover this.

ewencp: Actually, we can get rid of this entire file. This is still around from way back in v1.0 where JDK6 requirement meant we couldn't use a new enough version of Jetty that had support for this. We missed removing it in previous upgrades. I'll just get rid of it and replace with Jetty version.

PR #74, opened 2017-09-25
bump jackson version

norwood:


pom.xml
```diff
@@ -47,12 +47,14 @@
...
-        <bouncycastle.bcpkix.version>1.55</bouncycastle.bcpkix.version>
+        <apache.httpclient.version>4.5.3</apache.httpclient.version>
+        <bouncycastle.bcpkix.version>1.58</bouncycastle.bcpkix.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
+        <!-- can't upgrade jersey past 2.25 until we move to java 8 -->
         <jersey.version>2.25</jersey.version>
+        <!-- can't upgrade jetty past 9.2 until we move to java 8 -->
         <jetty.version>9.2.22.v20170606</jetty.version>
-        <jackson.version>2.8.4</jackson.version>
+        <jackson.version>2.9.1</jackson.version>
```
xvrl: anything specific we need in 2.9? is 2.8.9 not enough?

norwood: progress is always moving forward. iow i'd rather not let stuff rot on old versions

xvrl: true true, but if this is just about addressing CVEs then it might not be necessary, especially so close to the release. Jackson is actually really good about maintaining older versions and if they bump from 2.8 to 2.9 that usually indicates some change in behavior.

xvrl: looked a bit deeper into the 2.9 compatibility changes, and I *think* they should be fine unless we serialize `java.sql.Date`s somewhere.

ewencp:
The only place that would happen is in the converters for logical types like timestamp/date, but the converters should be translating them to their underlying type https://github.com/apache/kafka/blob/trunk/connect/json/src/main/java/org/apache/kafka/connect/json/JsonConverter.java#L225-L261

PR #78, opened 2017-12-15
Use common-docker as parent, to enable Docker builds in child projects

aayars:


aayars:
I narrowed the issues I was seeing down to clashing configurations for the assembly plugin.

PR #81, opened 2018-01-04
Ability to configure metrics tags

mageshn:
Dependent PR : https://github.com/confluentinc/common/pull/88

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -55,6 +55,7 @@ import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.EnumSet;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -152,8 +153,14 @@ public Server createServer() throws RestConfigException {
     ResourceConfig resourceConfig = new ResourceConfig();
 
     Map<String, String> metricTags = getMetricsTags();
+    Map<String, String> configuredTags = getConfiguration().getMap(RestConfig.METRICS_TAGS_CONFIG);
 
-    configureBaseApplication(resourceConfig, metricTags);
+    Map<String, String> combinedMetricsTags = new HashMap<>(metricTags);
```
edenhill: looks like metricTags is only used here, so maybe call getMetricsTags() directly here?

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -152,8 +153,14 @@ public Server createServer() throws RestConfigException {
...
+    combinedMetricsTags.putAll(configuredTags);
+
+    log.info("The metrics tags are " + combinedMetricsTags);
```
edenhill: Should this really be info, sounds like (conditional) debug?

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -105,6 +105,14 @@
       + "<code>MetricReporter</code> interface allows plugging in classes that will be notified "
       + "of new metric creation. The JmxReporter is always included to register JMX statistics.";
   protected static final String METRICS_REPORTER_CLASSES_DEFAULT = "";
+
+  public static final String METRICS_TAGS_CONFIG = "metrics.tags.map";
```
edenhill: double plural sounds weird, but may be right.
"metric.tags.map" or "metrics.tag.map" reads better

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -105,6 +105,14 @@
...
+  protected static final String METRICS_TAGS_DOC =
+      "A comma separated list of metrics tag entries of the form <tag_name>:<tag_value>. This can"
```
edenhill: Are we using :-separated key-value somewhere else, otherwise I think =-separated might be more common and intuitive.

mageshn: I have based this format off the Kafka config listener.security.protocol.map

mageshn:
@edenhill ping

PR #82, opened 2018-01-10
Ability to skip authentication for specific paths

mageshn:
This PR adds the ability to configure paths that can be accessed unauthenticated while turning on authentication. We add unauthenticated constraint for each of the path if there is an existing constraint. This is useful when people want to provide open access for all GET requests or health check resources, etc.


mageshn:
@ewencp I certainly like the idea of using something like annotations to achieve this. But unfortunately, the security constraints that we express are part of jetty lifecycle and jersey initialization starts only after jetty starts. So, there is no clean way to do this with annotations. So, we might have to stick with this config based approach.

PR #85, opened 2018-01-26
Make server request log pluggable.

C0urante:
Opening PR because I don't see an easy want to enable `Slf4jRequestLog.setExtended()` right now and there may be a use case for extended logging in the near future; however, there may also be more utility to allowing the request log to be pluggable in other circumstances as well.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -28,6 +28,7 @@ import org.eclipse.jetty.security.SecurityHandler;
 import org.eclipse.jetty.security.authentication.BasicAuthenticator;
 import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.NetworkTrafficServerConnector;
+import org.eclipse.jetty.server.RequestLog;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.handler.DefaultHandler;
 import org.eclipse.jetty.server.handler.HandlerCollection;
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -514,6 +512,29 @@ protected ObjectMapper getJsonMapper() {
     return new ObjectMapper();
   }
 
+  /**
+   * Returns the default {@link Slf4jRequestLog} used. Separate method made available in order to
```
ewencp: not sure i understand the justification for the extra method. a subclass could just use `super` to access the original version of `getRequestLog`, right?

C0urante: Right, but then there would be no guarantee that the call to `getRequestLog()` would return a `Slf4jRequestLog`; instead, it would just be a `RequestLog`. This way, if you want to return something that is a `RequestLog` but isn't a `Slf4jRequestLog` by overriding `getRequestLog()`, you can--and if you want to access the default request log with the guarantee that it's a `Slf4jRequestLog`, you can too.

C0urante:
Use case is in the PR description--just want a way to call `Slf4jRequestLog.setExtended()` to enable extended logging. Making it flexible in case other use cases arise in the future.

ewencp:
if we need to toggle extended format, why not just control it via a configuration? i don't think this has ever come up before, but i'm not sure hard coding it within each Application makes sense.

C0urante:
Responding to offline discussion with @ewencp by modifying changes from creating two new methods to just making the request log a protected class variable. Ready for next round of review.

PR #87, opened 2018-02-22
add support for websockets

norwood:


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -324,8 +328,23 @@ protected void doStop() throws Exception {
     StatisticsHandler statsHandler = new StatisticsHandler();
     statsHandler.setHandler(handlers);
 
-    server.setHandler(statsHandler);
-
+    final ServletContextHandler webSocketServletContext =
+        new ServletContextHandler(ServletContextHandler.SESSIONS);
+    webSocketServletContext.setContextPath("/ws");
```
norwood: should maybe make this configurable?

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -324,8 +328,23 @@ protected void doStop() throws Exception {
...
+        statsHandler
+    });
+
+    server.setHandler(contexts);
+    try {
+      ServerContainer container =
+          WebSocketServerContainerInitializer.configureContext(webSocketServletContext);
+      registerWebSocketEndpoints(container);
+    } catch (ServletException e) {
+      log.error("unable to configure sebsockets", e);
```
ewencp: sebsockets

ewencp: also, if this happens is logging and swallowing the right thing to do or just pass the exception up? letting it propagate seems like it is better unless there are expected cases where we'd try to set this up and but it would be expected to fail.

core/src/main/java/io/confluent/rest/Application.java
ewencp: maybe? is this Path part of the URL path or some other terminology? i've forgotten how jetty actually works. if it is user-facing, then I would say yes, it should be configurable.

core/pom.xml
```diff
@@ -16,6 +16,12 @@
     <name>rest-utils</name>
 
     <dependencies>
+        <dependency>
+            <groupId>org.eclipse.jetty.websocket</groupId>
+            <artifactId>javax-websocket-server-impl</artifactId>
+            <version>9.2.22.v20170606</version>
```
xvrl: can we put this version into dependency management in the parent pom to make sure we keep it in sync with the main jetty version?

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -335,6 +354,10 @@ protected void doStop() throws Exception {
     return server;
   }
 
+  protected void registerWebSocketEndpoints(ServerContainer container) {
```
xvrl: let's add some javadocs to explain when this needs to be overridden

norwood:
only thing that would be testable would really just be testing that jetty does what it's supposed to

core/src/main/java/io/confluent/rest/Application.java
norwood: yeah, its part of the path. this would collide with anyone who used a path prefix of `/ws` previously (nobody afaik)

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -420,6 +424,12 @@ public static ConfigDef baseConfigDef() {
             ENABLE_GZIP_COMPRESSION_DEFAULT,
             Importance.LOW,
             ENABLE_GZIP_COMPRESSION_DOC
+        ).define(
+            WEBSOCKET_PATH_PREFIX_CONFIG,
+            Type.STRING,
+            "/ws",
+            Importance.LOW,
```
xvrl: I think this should not even be documented. this configuration is only meant to be set programmatically. Is there a way to enforce that @ewencp ?

ewencp: AK ConfigDef has a defineInternal that should not get documented but I think that happened way after we split off from it, so probably not without some other changes in common.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -324,8 +328,21 @@ protected void doStop() throws Exception {
...
-    server.setHandler(statsHandler);
-
+    final ServletContextHandler webSocketServletContext =
+        new ServletContextHandler(ServletContextHandler.SESSIONS);
+    webSocketServletContext.setContextPath(
+        config.getString(RestConfig.WEBSOCKET_PATH_PREFIX_CONFIG)
+    );
+    final ContextHandlerCollection contexts = new ContextHandlerCollection();
+    contexts.setHandlers(new Handler[] {
+        webSocketServletContext,
```
xvrl: I wonder if we should just skip adding the websockets endpoint altogether if it isn't defined. This way we don't add it unless explicitly requested by the app?

norwood: the only thing i like more about this is that it will make the default for anything we use more standard. e.g. if some other product wants to use sockets it would by default be under the same location

xvrl: ok, that's a good argument

PR #88, opened 2018-03-06
Bump the version of Jackson

ewencp:


ewencp:
We're not going to get Jenkins feedback on this due to being on an older branch from before we had Jenkinsfiles, but I've run the unit/integration tests locally to validate.

PR #90, opened 2018-03-12
Support getting bound local port from Application and avoid using fixed port in unit tests

ewencp:


ewencp:
https://github.com/confluentinc/schema-registry/pull/758 and https://github.com/confluentinc/kafka-rest/pull/414 are downstream changes that will rely on this being merged first.

core/src/test/java/io/confluent/rest/ApplicationTest.java
```diff
@@ -79,6 +85,48 @@ public void testParseListenersNoPort() {
...
+    );
+    TestApplication testApp = new TestApplication(config);
+    testApp.start();
+    List<Integer> localPorts = testApp.localPorts();
+    assertEquals(1, localPorts.size());
+    // Validate not only that it isn't zero, but also a valid value
+    assertTrue("Should have a valid local port value", localPorts.get(0) > 0);
+    testApp.stop();
+    testApp.join();
+  }
```
edenhill: Mayeb add a test that verifies that localPorts returns the original PORT_CONFIG if != 0.

core/src/test/java/io/confluent/rest/ApplicationTest.java
```diff
@@ -79,6 +85,48 @@ public void testParseListenersNoPort() {
...
+  public void zeroForPortShouldHaveNonZeroLocalPort() throws Exception {
+    TestRestConfig config = new TestRestConfig(
+        Collections.singletonMap(RestConfig.PORT_CONFIG, "0")
+    );
+    TestApplication testApp = new TestApplication(config);
+    testApp.start();
+    List<Integer> localPorts = testApp.localPorts();
+    assertEquals(1, localPorts.size());
+    // Validate not only that it isn't zero, but also a valid value
+    assertTrue("Should have a valid local port value", localPorts.get(0) > 0);
```
edenhill: `&& <= 0xffff`

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -352,6 +380,10 @@ public void stop() throws Exception {
     server.stop();
   }
 
+  public void onStarted() {
+
```
edenhill: `// intentionally left bland`

core/src/test/java/io/confluent/rest/ExceptionHandlingTest.java
```diff
@@ -51,6 +51,7 @@
   public void setUp() throws Exception {
     Properties props = new Properties();
     props.setProperty("debug", "false");
+    props.setProperty(RestConfig.PORT_CONFIG, "0");
```
edenhill: extra points to add a comment why this is necessary

core/src/test/java/io/confluent/rest/ApplicationTest.java
ewencp: Heh, I considered it but this has the exact problem that I was trying to remove by removing the use of `choosePort()` -- you have to somehow choose a fixed value *before* allocating the port, which makes it a flakey test because there is no reliable way of choosing this without the potential for conflicts. I can try to pick something in a range that is unlikely to cause conflicts, in the hope that this will be the only test that will use a fixed port, but it will still be error prone.

I guess something very high in the IANA registered port range and below the ephemeral range could possibly work, at least for a single test...

edenhill: Ha, yeah of course, you are right :)
You could try a range of high ports, skipping to the next if binding fails, but it might be overkill for something like this. :-1:

edenhill:
LGTM

PR #94, opened 2018-05-02
Rely on jackson.version from common pom

ewencp:


ewencp:
This was already setup on earlier release branches, not sure what happened in the merge that it didn't make it into 4.1.x and master...

PR #97, opened 2018-05-22
Remove java.version since it's defined in common

ijuma:


ijuma:
@ewencp good point, I'll update common to have that!

PR #98, opened 2018-05-23
Jetty 9.4 follow-up

xvrl:
* restore proper error dispatch test removed in #71 
* switch from deprecated GzipFilter to GzipHandler
* add integration test for gzip encoding

### DIFF
--- a/core/src/test/java/io/confluent/rest/GzipHandlerIntegrationTest.java
+++ b/core/src/test/java/io/confluent/rest/GzipHandlerIntegrationTest.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2018 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


ijuma:
Are these fixes or improvements? As in, was any behaviour actually broken or it's just a code cleanliness issue (which is good, of course).

xvrl:
@ijuma removing deprecated classes

core/src/test/java/io/confluent/rest/GzipHandlerIntegrationTest.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright 2016 Confluent Inc.
```
ewencp: lol, doesn't really matter but i think your header template is out of date

xvrl: Ha, the Gzip filter was added in 2016 so we can say it’s keeping with the spirit of the code :)

PR #100, opened 2018-05-31
Add idle timeout config

joncourt:
Exposed the jetting idle timeout setting through RestConfig and in the Application class so we can extend the timeout for the connection our mainframe makes to the Kafka Rest Proxy. The mainframe can't deal with the default 30s timeout - needs at least 60s.

New configuration option: idle.timeout.ms, default: 30000 (30s)


joncourt:
update to 5.0.0 - creating new pull request against confluentinc/5.0.0-post

PR #109, opened 2018-08-20
CC-2533: Add support for runtime tags in metrics

wicknicks:
Signed-off-by: Arjun Satish <arjun@confluent.io>

### DIFF
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -55,13 +55,13 @@ import io.confluent.rest.annotations.PerformanceMetric;
  */
 public class MetricsResourceMethodApplicationListener implements ApplicationEventListener {
 
-  public static final String RUNTIME_TAGS_PROP_KEY = "_runtime_tags";
+  public static final String RUNTIME_TAGS_PROP_KEY = "_request_tags";
 
   private final Metrics metrics;
   private final String metricGrpPrefix;
   private Map<String, String> metricTags;
   Time time;
-  private Map<Method, RuntimeMetrics> methodMetrics = new HashMap<>();
+  private Map<Method, RequestScopedMetrics> methodMetrics = new HashMap<>();
 
   public MetricsResourceMethodApplicationListener(Metrics metrics, String metricGrpPrefix,
                                            Map<String,String> metricTags, Time time) {
@@ -77,7 +77,7 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
     if (event.getType() == ApplicationEvent.Type.INITIALIZATION_FINISHED) {
       // Special null key is used for global stats
       MethodMetrics m = new MethodMetrics(null, null, this.metrics, metricGrpPrefix, metricTags);
-      methodMetrics.put(null, new RuntimeMetrics(m, new ConstructionContext(this)));
+      methodMetrics.put(null, new RequestScopedMetrics(m, new ConstructionContext(this)));
 
       for (final Resource resource : event.getResourceModel().getResources()) {
         for (final ResourceMethod method : resource.getAllMethods()) {
@@ -100,7 +100,7 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
 
       MethodMetrics m = new MethodMetrics(method, annotation, metrics, metricGrpPrefix, metricTags);
       ConstructionContext context = new ConstructionContext(method, annotation, this);
-      methodMetrics.put(definitionMethod, new RuntimeMetrics(m, context));
+      methodMetrics.put(definitionMethod, new RequestScopedMetrics(m, context));
     }
   }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -103,6 +109,55 @@ public RequestEventListener onRequest(final RequestEvent event) {
...
+    private MetricsResourceMethodApplicationListener methodAppListener;
+
+    public ConstructionContext(MetricsResourceMethodApplicationListener methodAppListener) {
+      this(null, null, methodAppListener);
+    }
+
+    public ConstructionContext(
+        ResourceMethod method,
+        PerformanceMetric performanceMetric,
+        MetricsResourceMethodApplicationListener methodAppListener
```
mageshn: can we have just the required fields here instead of the whole listener?

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -260,7 +315,23 @@ private MethodMetrics getMethodMetrics(RequestEvent event) {
...
       }
-      return this.metrics.get(method.getInvocable().getDefinitionMethod());
+
+      RuntimeMetrics metrics = this.metrics.get(method.getInvocable().getDefinitionMethod());
+      if (metrics == null) {
+        return null;
+      }
+
+      Map tags = null;
+      if (event.getContainerRequest() != null) {
```
mageshn: I don't think containerRequest would ever be null.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -260,7 +315,23 @@ private MethodMetrics getMethodMetrics(RequestEvent event) {
       if (method == null) {
         return null;
       }
-      return this.metrics.get(method.getInvocable().getDefinitionMethod());
+
+      RuntimeMetrics metrics = this.metrics.get(method.getInvocable().getDefinitionMethod());
+      if (metrics == null) {
```
mageshn: would we ever have this scenario?

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -260,7 +315,23 @@ private MethodMetrics getMethodMetrics(RequestEvent event) {
...
+      if (event.getContainerRequest() != null) {
+        tags = (Map) event.getContainerRequest().getProperty(RUNTIME_TAGS_PROP_KEY);
+      }
+
+      if (tags == null) {
+        return metrics.metrics();
+      }
+
+      // we have additional tags, find the appropriate metrics holder
+      return metrics.metrics(tags);
```
mageshn: can we just use the same metrics method for both scenarios -no additional tags and additional tags

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -51,12 +54,14 @@
  * latency (average, 90th, 99th, etc).
  */
 public class MetricsResourceMethodApplicationListener implements ApplicationEventListener {
+
+  public static final String RUNTIME_TAGS_PROP_KEY = "_runtime_tags";
```
mageshn: nit : perhaps  _request_tags ?

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
wicknicks: Metrics were  immutable in this project (since the underlying sensor objects are also immutable). Once created they hold their tags for the application lifetime.

wicknicks:
@norwood @ewencp added you to this PR since you guys are familiar with this repo. Feel free to add anyone else who might be familiar with it and can help determine if the changes don't break any contracts. thanks!

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
mageshn: oh, I meant instead of two methods metrics() and metrics(tags) can we just use metrics(tags) ?

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -51,12 +54,14 @@
  * latency (average, 90th, 99th, etc).
  */
 public class MetricsResourceMethodApplicationListener implements ApplicationEventListener {
+
+  public static final String RUNTIME_TAGS_PROP_KEY = "_request_tags";
```
ewencp: where does this come from? Since we're getting tags out of the container request, the value here is set by Jersey? If so, is there no existing public constant from Jersey instead?

mageshn: the contant name should perhaps be REQUEST_TAGS_PROP_KEY

mageshn: @ewencp These can be added via the filters as necessary. So, basically allows you to inject tags based on the request. Let's say if someone has an authentication filter and want to track all the metrics by the user, they could do so by using this.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
wicknicks: I think it would be ok to keep two different methods. would it clearer if we renamed `metrics(tags)` to something like `requestMetrics(tags)`?

PR #110, opened 2018-08-24
WIP - enable rest-utils to load authentication plugin via config

big-andy-coates:
VERY MUCH WIP.... but as I'm away on holiday from the 25th until the 4th I thought I'd raise this one to get some feedback while I'm away.

General thinking here is that services such as KSQL and SR will want the ability to configure authentication and authorisation via plugins. 

We shouldn't let internal implementation details, such as our use of Jetty, bleed through into these plugins. So we need to expose new non-jetty interfaces to implement.

I have this code working on a KSQL fork, which functional tests.  I've just hacked this across to rest-utils, (where I believe it should live). I'll tidy up / add tests etc, on my return.


maxzheng:
Removing tools as there isn't any Tools related change in here

big-andy-coates:
[clabot:check]

PR #112, opened 2018-09-05
MMA-2878 - Enable security on websocket end points as well as http.

big-andy-coates:
Fixes: [MMA-2878](https://confluentinc.atlassian.net/browse/MMA-2878)

Noticed that the rest utils does not currently set up security on the websocket end points. This PR adds a puts the same auth on the websocket requests as was already on the http requests, and adds suitable tests.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -115,5 +115,17 @@
             <classifier>test</classifier>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.asynchttpclient</groupId>
+            <artifactId>async-http-client</artifactId>
+            <version>2.2.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
+            <version>${slf4j.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -118,13 +118,12 @@
         <dependency>
             <groupId>org.asynchttpclient</groupId>
             <artifactId>async-http-client</artifactId>
-            <version>2.2.0</version>
+            <version>${asynchttpclient.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-log4j12</artifactId>
-            <version>${slf4j.version}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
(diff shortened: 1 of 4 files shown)


core/pom.xml
```diff
@@ -115,5 +115,17 @@
             <classifier>test</classifier>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.asynchttpclient</groupId>
+            <artifactId>async-http-client</artifactId>
+            <version>2.2.0</version>
```
xvrl: version should be defined in dependency-mgmgt section in the main pom.

core/pom.xml
```diff
@@ -115,5 +115,17 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
+            <version>${slf4j.version}</version>
```
xvrl: version is not necessary, since it's already defined in common pom dependency management

core/src/test/java/io/confluent/rest/SaslTest.java
```diff
@@ -1,4 +1,4 @@
-/**
```
xvrl: nit, existing files have two stars, we should probably put this into checkstyle

core/src/test/java/io/confluent/rest/SaslTest.java
```diff
@@ -46,30 +60,36 @@
...
   private String previousAuthConfig;
   private SaslTestApplication app;
   private CloseableHttpClient httpclient;
-  String httpUri = "http://localhost:8080";
 
   @Before
   public void setUp() throws Exception {
-    jaasFile = File.createTempFile("jaas", ".config");
-    loginPropertiesFile = File.createTempFile("login", ".properties");
+    File jaasFile = File.createTempFile("jaas", ".config");
```
xvrl: we should probably move the tempfiles to use a [junit tempfolder rule](https://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html) to avoid littering tmp with files.

core/src/test/java/io/confluent/rest/SaslTest.java
big-andy-coates: two stars is incorrect. this is not a java doc.

core/src/test/java/io/confluent/rest/SaslTest.java
big-andy-coates: Sure, I'll take a look.

core/pom.xml
big-andy-coates: Yep, will do.

core/pom.xml
big-andy-coates: on it.

PR #113, opened 2018-09-06
Support custom init.

big-andy-coates:
This PR looks to add some very basic hooks to allow the RESTful and WS servlets to be customised using _configuration_, rather than by subclassing.

It's petty basic. To customise, you write a class that `implements BiConsumer<ServletContextHandler, ? extends RestConfig>` and add it to the list of initialisers for either or both servlets. The app will then instantiate your class and pass you the appropriate servlet and app config, (after all other initialisation has been done, but before it is started), for you to do what you want with.

I've used `BiConsumer` as its a standard Java interface, meaning no need for any other deps.

There is obviously a risk that people can do dumb stuff with this, but that's not really the point. People can always do dumb stuff!  This enabled people to tweak things if they really want / need to.

I've made it clear this is an unsupported / internal / likely-to-change / no-promises-made-on-backwards-compatibility feature.

### DIFF
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -228,7 +228,7 @@ public class RestConfig extends AbstractConfig {
       "websocket.servlet.initializor.classes";
   public static final String WEBSOCKET_SERVLET_INITIALIZERS_CLASSES_DOC =
       "Defines one or more initializer for the websocket endpoint's ServletContextHandler. "
-          + "Each initializer must implement <ServletContextHandler>. "
+          + "Each initializer must implement Consumer<ServletContextHandler>. "
           + "It will be called to perform custom initialization of the handler, in order. "
           + "This is an internal feature and subject to change, "
           + "including changes to the Jetty version";
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -219,7 +219,8 @@ public class RestConfig extends AbstractConfig {
       "rest.servlet.initializor.classes";
   public static final String REST_SERVLET_INITIALIZERS_CLASSES_DOC =
       "Defines one or more initializer for the rest endpoint's ServletContextHandler. "
-          + "Each initializer must implement Consumer<ServletContextHandler>. "
+          + "Each initializer must implement "
+          + "BiConsumer<ServletContextHandler, ? extends RestConfig>. "
           + "It will be called to perform initialization of the handler, in order. "
           + "This is an internal feature and subject to change, "
           + "including changes to the Jetty version";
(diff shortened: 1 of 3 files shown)


xvrl:
A user already has to subclass `Application` in order to create a service, so instead of passing the initializer classnames as configs, I believe it would be more flexible to pass the initializer instances as part of instantiation through a method that gets overridden in the user application. This leaves the user control over how to instantiate the `Consumer` classes, which is important if you rely on dependency-injection. Using classnames in configuration properties breaks usability for someone relying on dependency-injection – since instantiation is now outside of their control they can't inject `Consumer` instances. Even without dependency-injection, you may still want to control how the classes are instantiated.

big-andy-coates:
Hi @xvrl, I think our terminology may be a bit mixed here.  When I say a 'user can configure' I'm not talking about someone using rest-utils, but someone how is using an application that is built on top of rest utils, e.g. KSQL.  So a user of KSQL / SR / etc can use these configuration options to customise the application.

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -215,6 +215,25 @@
   protected static final String ENABLE_GZIP_COMPRESSION_DOC = "Enable gzip compression";
   private static final boolean ENABLE_GZIP_COMPRESSION_DEFAULT = true;
 
+  public static final String REST_SERVLET_INITIALIZERS_CLASSES_CONFIG =
+      "rest.servlet.initializor.classes";
+  public static final String REST_SERVLET_INITIALIZERS_CLASSES_DOC =
+      "Defines one or more initializer for the rest endpoint's ServletContextHandler. "
+          + "Each initializer must implement Consumer<ServletContextHandler>. "
```
xvrl: I wonder if we should create an explicit interface that extends `Consumer<ServletContextHandler>`. The benefit would be twofold:
- any type mismatch will be handled with an explicit error message from `RestConfig.getConfiguredInstances(...)` (we currently don't have a way of enforcing the generic type)
- If the interface ever needs to evolve, any plugin that needs to be updated will be caught at compile time rather than runtime.

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -215,6 +215,25 @@
...
+          + "It will be called to perform initialization of the handler, in order. "
+          + "This is an internal feature and subject to change, "
+          + "including changes to the Jetty version";
+
+  public static final String WEBSOCKET_SERVLET_INITIALIZERS_CLASSES_CONFIG =
+      "websocket.servlet.initializor.classes";
+  public static final String WEBSOCKET_SERVLET_INITIALIZERS_CLASSES_DOC =
+      "Defines one or more initializer for the websocket endpoint's ServletContextHandler. "
+          + "Each initializer must implement <ServletContextHandler>. "
```
xvrl: typo. missing class name

core/src/main/java/io/confluent/rest/RestConfig.java
big-andy-coates: Hey @xvrl, as I mentioned in this PR's blurb at the top, I've deliberately _not_ used a custom interface, but a standard Java one, so that there are no rest-util dependencies needed.  

However, if you want, I can create a new interface and stick it in a new module in this repo.

core/src/main/java/io/confluent/rest/RestConfig.java
big-andy-coates: Fixed

PR #114, opened 2018-09-07
Switch checkstyle suppression to comments

big-andy-coates:
Switch checkstyle rule suppression to use comments, rather than suppression file, purely because the former works well with IDE plugins, (specifically the IDEA CheckStyle plugin), where as the latter does not.


norwood:
this looks ugly as shit, but i like the idea.

you can't just do it at the end of the line?

norwood:
also, please either `squash and merge` or fix your commits so we don't end up with merges in our commit history

PR #117, opened 2018-09-17
fix authentication breaking CORS preflight requests

xvrl:
* disable authentication for OPTIONS requests when CORS is enabled
* do not forward CORS preflight requests down the filter chain

Note: regular OPTIONS requests without CORS preflight headers will still
be forwarded down the filter chain, so this would break applications
that require authentication for those requests. I don't think we have
any application explicitly handling OPTIONS requests, so the risk should
be low.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -303,19 +303,20 @@ public abstract class Application<T extends RestConfig> {
     List<String> unsecurePaths = config.getList(RestConfig.AUTHENTICATION_SKIP_PATHS);
     setUnsecurePathConstraints(context, unsecurePaths);
 
-    String allowedOrigins = getConfiguration().getString(
-        RestConfig.ACCESS_CONTROL_ALLOW_ORIGIN_CONFIG
-    );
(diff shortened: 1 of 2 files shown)


xvrl:
@ewencp it only intercepts preflight requests, i.e. OPTIONS requests made by the browser with the sole intent of checking that it is allowed to do cross origin requests on that endpoint, regular requests that only need the right CORS headers added are still passed through.

ewencp:
@xvrl I was referring to this comment

> Note: regular OPTIONS requests without CORS preflight headers will still be forwarded down the filter chain, so this would break applications that require authentication for those requests. I don't think we have any application explicitly handling OPTIONS requests, so the risk should be low.

It just seems like there must be some reasonably easy way to do this without breaking auth'd OPTIONS requests, but maybe i'm overestimating jetty? Just seems kinda bad to break this even if it's not currently being used -- it's a security hole just waiting to happen.

xvrl:
@ewencp in jetty authentication is purely based on path and method and happen before any of the filters kick in. That's the reason I am only disabling auth on OPTIONS methods when CORS is enabled.
As an alternative we could extend the security handler and only allow through CORS preflight requests.

PR #118, opened 2018-09-19
upgrade to jetty 9.4.12

xvrl:
Java 11 + TLS 1.3 here we come


xvrl:
@ijuma we should, we don't have a good place to do this in this project unfortunately, so we'll have to call it out in downstream project docs.

ijuma:
@xvrl sound good. Link for reference (I was on my phone previously): https://github.com/apache/kafka/pull/5600/files

xvrl:
Both KSQL and Schema-Registry appear pass tests with this, so I'm going to merge cc @mageshn @big-andy-coates

PR #119, opened 2018-09-19
Add concept of Resource Extensions

big-andy-coates:


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -186,6 +189,7 @@ public Server createServer() throws ServletException {
     combinedMetricsTags.putAll(configuredTags);
 
     configureBaseApplication(resourceConfig, combinedMetricsTags);
+    configureResourceExtensions(resourceConfig);
     setupResources(resourceConfig, getConfiguration());
```
xvrl: an application already has to implement setupResources to add resources, maybe we should let the application decide whether to configure resource extensions before or after it adds its own resources?

core/src/main/java/io/confluent/rest/extention/ResourceExtension.java
```diff
@@ -0,0 +1,43 @@
...
+public interface ResourceExtension<AppT extends Application<? extends RestConfig>>
+    extends Closeable {
+
+  /**
+   * Called to register the extension's resources when the rest server starts up.
+   *
+   * @param config the configuration used to register resources
+   * @param app the application.
+   */
+  void register(Configurable<?> config, AppT app);
```
xvrl: isn't the config redundant, since it's already available through app.getConfiguration()?

big-andy-coates: `getConfiguration()` returns the `RestConfig` style config.  The `config` is the rest server configuration, which allows the extension to register new resources within the rest server.

core/src/main/java/io/confluent/rest/Application.java
big-andy-coates: As per the last PR this is about allowing users of some product to extend it using config, rather than some user of rest-utils.

PR #122, opened 2018-09-28
Move common auth stuff to util class.

big-andy-coates:


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -442,54 +434,35 @@ protected void registerWebSocketEndpoints(ServerContainer container) {
...
     securityHandler.setRealmName(realm);
+
+    AuthUtil.createUnsecuredConstraints(config)
+        .forEach(securityHandler::addConstraintMapping);
+
     return securityHandler;
   }
 
+  @Deprecated // Use createGlobalAuthConstraint() or AuthUtil.createGlobalAuthConstraint(config)
   protected ConstraintMapping createGlobalAuthConstraint(List<String> roles) {
```
xvrl: We don't need to keep this method around, it's not a public api. We should just update any existing overrides to use `createGlobalAuthConstraint`

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
```diff
@@ -0,0 +1,118 @@
...
+   * <p>To be authorised a user must below to at least one of the supplied {@code roles}.
+   *
+   * <p>If {@code omitOptions} is {@code true} the HTTP OPTIONS requests are excluded from
+   * auth checks, e.g. as required for CORS preflight checks.
+   *
+   * @param roles the list of valid roles
+   * @param omitOptions if {@code true}, OPTIONS requests are not subject to auth checks.
+   * @return the constraint mapping.
+   */
+  public static ConstraintMapping createGlobalAuthConstraint(
```
xvrl: let's remove `Application.createGlobalAuthConstraint(List<String> roles)` so we can simplify things and merge this into a single `createGlobalAuthConstraint(final RestConfig restConfig)`

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  /**
+   * Build constraints for any unsecured paths.
+   *
+   * @param unsecuredPaths the list of paths that should not be secured.
+   * @return the list of constraint mappings.
+   */
+  @SuppressWarnings("WeakerAccess") // API call.
+  public static List<ConstraintMapping> createUnsecuredConstraints(
```
xvrl: this method can be made private

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
```diff
@@ -0,0 +1,118 @@
...
+    return createUnsecuredConstraints(unsecuredPaths);
+  }
+
+  /**
+   * Build constraints for any unsecured paths.
+   *
+   * @param unsecuredPaths the list of paths that should not be secured.
+   * @return the list of constraint mappings.
+   */
+  @SuppressWarnings("WeakerAccess") // API call.
```
xvrl: what is this warning about?

core/src/main/java/io/confluent/rest/Application.java
big-andy-coates: OK

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
big-andy-coates: Currently on the test use this method, so you get a warning saying it can be package private.

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
```diff
@@ -0,0 +1,118 @@
...
+  public static List<ConstraintMapping> createUnsecuredConstraints(
+      final List<String> unsecuredPaths) {
+
+    return unsecuredPaths.stream()
+        .map(AuthUtil::toUnsecuredConstraint)
+        .collect(Collectors.toList());
```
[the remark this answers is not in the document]
big-andy-coates: I would say so, yes.

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
big-andy-coates: Done

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
big-andy-coates: removed.

PR #123, opened 2018-10-02
Add test for default 'anyRole' config.

big-andy-coates:
Add test to check default config adds global constraint for 'anyRole' and not 'anyAuth'.

TBH, it seems to me that the default would be `anyAuth`, i.e. any authenticated user can connect, rather than `anyRole`, i.e. any authenticated user who belongs to _at least one role_. 

If we're happy with the current default, this test at least ensures we don't unintentionally change it.

### DIFF
--- a/core/src/test/java/io/confluent/rest/auth/AuthUtilTest.java
+++ b/core/src/test/java/io/confluent/rest/auth/AuthUtilTest.java
@@ -18,7 +18,7 @@ package io.confluent.rest.auth;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.nullValue;
-import static org.junit.Assert.*;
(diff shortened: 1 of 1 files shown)


xvrl:
@big-andy-coates agree it's confusing, but that's mostly the Servlet Specification's fault, and the fact that the constraint is put on the roles themselves, where `*` means any one role, and `**` means any authenticated user.

big-andy-coates:
> <3 tests

???

big-andy-coates:
@xvrl, yeah I know about `*` vs `**`... I guess what I'm asking / suggesting in a round about way is 'should the default not be `**`, rather than `*`'?

xvrl:
@big-andy-coates I understand your question. I agree the default is a bit confusing. Switching the default would be a breaking change for Schema-Registry and Rest-Proxy, so we would need a migration path to change the default, not sure it's worth the effort.

PR #124, opened 2018-10-10
fix missing newlines

xvrl:


xvrl:
@ewencp I think we should just make it consistent as part of our codestyle, just like any other whitespace. I have a PR out to do so https://github.com/confluentinc/common/pull/166

ewencp:
@xvrl Followed up on that thread. This seems likely to turn extremely bike sheddy....

PR #126, opened 2018-11-28
Expose SslContextFactory

jonchiu:
Allow applications built on `rest.Application` to extract the `SslContextFactory`.

Refactor creation of `SslContextFactory` so that it gets created and configured once instead of potentially multiple times.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -92,6 +92,7 @@ public abstract class Application<T extends RestConfig> {
   protected CountDownLatch shutdownLatch = new CountDownLatch(1);
   protected Metrics metrics;
   protected final Slf4jRequestLog requestLog;
+  protected SslContextFactory sslContextFactory;
 
   private static final Logger log = LoggerFactory.getLogger(Application.class);
 
@@ -109,6 +110,7 @@ public abstract class Application<T extends RestConfig> {
     this.requestLog = new Slf4jRequestLog();
     this.requestLog.setLoggerName(config.getString(RestConfig.REQUEST_LOGGER_NAME_CONFIG));
     this.requestLog.setLogLatency(true);
+    this.sslContextFactory = createSslContextFactory();
   }
 
   /**
@@ -136,6 +138,13 @@ public abstract class Application<T extends RestConfig> {
    */
   protected ResourceCollection getStaticResources() {
     return null;
+  }
+
+  /**
+   * expose SslContextFactory
+   */
+  protected SslContextFactory getSslContextFactory() {
+    return this.sslContextFactory;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -257,6 +265,8 @@ protected void doStop() throws Exception {
           sslContextFactory.setProtocol(config.getString(RestConfig.SSL_PROVIDER_CONFIG));
         }
 
+        sslContextFactory.setRenegotiationAllowed(false);
+        this.sslContextFactory = sslContextFactory;
```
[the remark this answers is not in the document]
jonchiu: to confirm what you mean, create a single `sslContextFactory` to be shared for each `NetworkTrafficServerConnector` if listener is ssl?

jonchiu:
[clabot:check]

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -257,6 +266,7 @@ protected void doStop() throws Exception {
           sslContextFactory.setProtocol(config.getString(RestConfig.SSL_PROVIDER_CONFIG));
         }
 
+        sslContextFactory.setRenegotiationAllowed(false);
```
xvrl: is there a particular reason we are changing this setting?

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -190,13 +198,14 @@ protected void doStop() throws Exception {
...
+    SslContextFactory sslContextFactory = new SslContextFactory();
+    this.sslContextFactory = sslContextFactory;
     for (URI listener : listeners) {
       log.info("Adding listener: " + listener.toString());
       NetworkTrafficServerConnector connector;
       if (listener.getScheme().equals("http")) {
         connector = new NetworkTrafficServerConnector(server);
       } else {
-        SslContextFactory sslContextFactory = new SslContextFactory();
         if (!config.getString(RestConfig.SSL_KEYSTORE_LOCATION_CONFIG).isEmpty()) {
```
xvrl: is there a reason we are re-configuring the factory for each each ssl listener? I think our settings are not specific to any listener, so it probably makes sense to move this into a separate method and call it once rather than inside a loop

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -190,13 +198,14 @@ protected void doStop() throws Exception {
 
     List<URI> listeners = parseListeners(config.getList(RestConfig.LISTENERS_CONFIG),
             config.getInt(RestConfig.PORT_CONFIG), Arrays.asList("http", "https"), "http");
+    SslContextFactory sslContextFactory = new SslContextFactory();
+    this.sslContextFactory = sslContextFactory;
```
norwood: i think its cleaner to either
- do this in the constructor
- do this lazily in the getter

will prevent any races here.

core/src/main/java/io/confluent/rest/Application.java
jonchiu: no, this is an artifact of rebasing against `5.1.x` and moving it to `5.0.x`. 

https://github.com/confluentinc/rest-utils/blob/5.1.x/core/src/main/java/io/confluent/rest/Application.java#L283

i will remove.

PR #129, opened 2019-02-26
Add percentile metrics for latency

mageshn:
Include percentile metrics for latency.

Open Question: I just picked an arbitrary bucket size and sizing scheme. If anyone can recommend a scientific way to get that, it's great.

### DIFF
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -39,6 +39,7 @@ import java.util.Map;
 import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 
 import io.confluent.common.metrics.MetricName;
 import io.confluent.common.metrics.Metrics;
@@ -60,7 +61,7 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
   public static final String REQUEST_TAGS_PROP_KEY = "_request_tags";
 
   private static final int PERCENTILE_NUM_BUCKETS = 100;
-  private static final double PERCENTILE_MAX_LATENCY_IN_MS = 10 * 1000;
+  private static final double PERCENTILE_MAX_LATENCY_IN_MS = TimeUnit.SECONDS.toMillis(5);
 
   private final Metrics metrics;
   private final String metricGrpPrefix;
@@ -226,7 +227,7 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
       Percentiles percs = new Percentiles(Float.SIZE / 8 * PERCENTILE_NUM_BUCKETS,
           0.0,
           PERCENTILE_MAX_LATENCY_IN_MS,
-          Percentiles.BucketSizing.LINEAR,
+          Percentiles.BucketSizing.CONSTANT,
           new Percentile(new MetricName(
               getName(method, annotation, "request-latency-95"), metricGrpName,
               "The 95th percentile request latency in ms", metricTags), 95),
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -16,6 +16,8 @@
 
 package io.confluent.rest.metrics;
 
+import io.confluent.common.metrics.stats.Percentile;
+import io.confluent.common.metrics.stats.Percentiles;
 import org.glassfish.jersey.server.ContainerRequest;
 import org.glassfish.jersey.server.ContainerResponse;
 import org.glassfish.jersey.server.model.Resource;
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -218,6 +220,24 @@ public MethodMetrics(ResourceMethod method, PerformanceMetric annotation, Metric
           "The maximum request latency in ms", metricTags);
       this.requestLatencySensor.add(metricName, new Max());
 
+      Percentiles percs = new Percentiles(4 * 100,
+          0.0,
+          100.0,
```
xvrl: how did we pick 100ms for maximum?

mageshn: as I mentioned I just picked an arbitrary number here, hoping we could finalize them in the PR discussion :)

rhauch:
The 50p doesn't seem that advantageous -- median latency is not terribly useful (to @ewencp's point).

Is there any need to be able to aggregate these metrics across nodes? If not, then percentiles would be sufficient. If that will be a requirement, then would histograms be a better choice or are they overkill?

rhauch:
Ah, we don't capture histograms, and the Histogram class is used for reporting purposes. Sorry for the confusion.

xvrl:
I'm assuming those percentile metrics are copied from Kafka? Do we know what time period the percentiles cover or how the samples get aged out? Understanding that will be key to define alerts and SLAs.
In my experience, only 95th and 99th percentile will ever be looked at, so we can probably expose only those two by default. If we care about the other ones, maybe we can make it configurable.

@rhauch I don't think we have a way to collect or aggregate the raw histogram data today, so it's probably overkill to add that today. At some point however, I'd love to look into other probabilistic sketches that would allow us to provide aggregate visibility into the tail of the distribution.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -218,6 +220,24 @@ public MethodMetrics(ResourceMethod method, PerformanceMetric annotation, Metric
           "The maximum request latency in ms", metricTags);
       this.requestLatencySensor.add(metricName, new Max());
 
+      Percentiles percs = new Percentiles(4 * 100,
```
xvrl: nit, using Float.SIZE / 8 instead of 4 and defining a constant numBuckets = 100 would be more explicit about how this relates to the underlying storage bytes.

mageshn: sure

mageshn:
If everyone agrees, I will just include 95th and 99th percentile in this PR.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
xvrl: The way I would think about this is to pick an absolute maximum upper threshold that we would deem unacceptable for any http endpoint, e.g. 10 seconds, beyond which we don't even care to know how bad things are, while still giving us the granularity we need in the nominal path (e.g. 100 buckets would give us 100ms granularity in this case). For low latency services, maybe 5s max, 50ms granularity is better.
The key thing to keep in mind is that once we set a max, e.g. 5s, then a 5s 99th percentile is not that meaningful anymore, it just tells us that 99th percent of the request are 5s *or more*, the real 99th percentile could be 30s for all we know.

mageshn: I think 10s should cover pretty much all CP components at the moment. The only question would be if we 100 ms or 50 ms granularity.

mageshn:
@xvrl @rhauch do you want to take another pass at this?

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -57,6 +59,9 @@
 
   public static final String REQUEST_TAGS_PROP_KEY = "_request_tags";
 
+  private static final int PERCENTILE_NUM_BUCKETS = 100;
+  private static final double PERCENTILE_MAX_LATENCY_IN_MS = 10 * 1000;
```
rhauch: How about:

```
  private static final double PERCENTILE_MAX_LATENCY_IN_MS = TimeUnit.SECONDS(10).toMillis();
```

10 seconds still seems high to me, and I might prefer a max of 5 seconds with a 50ms granularity. After all, anything in the 99p (5 seconds or more) is still bad performance.

mageshn:
Thanks, @rhauch and @xvrl. Per @rhauch suggestion, I will switch to use 5s as the max. I also noticed that I was using Liner bin vs Constant bin. Based on our discussions, I assume that we all tend to believe that this should be a constant bin. Is that right?

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
ewencp: rest proxy has a small number of request paths where high latencies are desirable in some cases. in particular, consumers blocking when there is no data is a good thing, which results in high request latencies.

worth noting since it might make certain percentiles hard to use in that case, e.g. if you have sufficient traffic that consumes low-flow topics.

not sure that specific case should keep us from providing sane defaults for others since i'm not sure how to sanely take advantage of aggregate latency metrics in that case, but just wanted to point out that there is at least one case where the settings mentioned above might not match expectations.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
ewencp: re: latencies, we should make sure there aren't any cross DC use cases that could have larger latencies than we tend to think of with single DC cases. e.g., it is possible to run schema registry in single global cluster mode cross DC, in which case a write request in one DC might need to be forwarded across DCs, wait for Kafka write, and come back across DCs. 5s still wouldn't be great, but for that setup it might be acceptable for some operations.

mageshn:
retest this please

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
mageshn: In that case, would it be acceptable to have a max of 10s with 200 bins then? We would still have a 50 ms granularity. Is that a better compromise?

ewencp: tbh, i don't have strong opinions here. i was just calling out an example where such high latency might not be considered crazy since the thread seemed to presume that less than ideal out-of-DC possible latencies were unreasonable.

it's reasonable to consider defaults/common case as well and default for those. most use cases are in-DC. if we want to default for those and figure out how to customize/extend later if needed, i'm fine with that.

PR #133, opened 2019-03-07
move rest-utils to ak configdefs

norwood:


norwood:
tagged ya'll since thats what github told me to do, and it will affect all of us.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -191,7 +191,16 @@ public Server createServer() throws ServletException {
...
     ResourceConfig resourceConfig = new ResourceConfig();
 
-    Map<String, String> configuredTags = getConfiguration().getMap(RestConfig.METRICS_TAGS_CONFIG);
+    Map<String, String> configuredTags = new HashMap<>();
+    // This is copied from the old MAP implementation from cp ConfigDef.Type.MAP
+    List<String> metricsTags = getConfiguration().getList(RestConfig.METRICS_TAGS_CONFIG);
+    for (String entry : getConfiguration().getList(RestConfig.METRICS_TAGS_CONFIG)) {
+      String[] keyValue = entry.split("\\s*:\\s*", -1);
+      if (keyValue.length != 2) {
+        throw new ConfigException("Map entry should be of form <key>:<value");
```
woshifudayun:
```suggestion
        throw new ConfigException("Map entry should be of form <key>:<value>");
```

norwood:
@ewencp if people are 👍 here i will make prs for all downstream repos and then we can merge all at once 😰

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -191,7 +191,16 @@ public Server createServer() throws ServletException {
     // The configuration for the JAX-RS REST service
     ResourceConfig resourceConfig = new ResourceConfig();
 
-    Map<String, String> configuredTags = getConfiguration().getMap(RestConfig.METRICS_TAGS_CONFIG);
+    Map<String, String> configuredTags = new HashMap<>();
+    // This is copied from the old MAP implementation from cp ConfigDef.Type.MAP
+    List<String> metricsTags = getConfiguration().getList(RestConfig.METRICS_TAGS_CONFIG);
```
big-andy-coates: nit: not used.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -191,7 +191,16 @@ public Server createServer() throws ServletException {
...
+    for (String entry : getConfiguration().getList(RestConfig.METRICS_TAGS_CONFIG)) {
```
big-andy-coates: Probably worth putting this into a utility method.

norwood: also added a test

big-andy-coates:
@norwood any thoughts about just adding a MAP type on Apache ConfigDef?

norwood:
yeah, i'd be down to move it. but i'd rather not gate this pr on moving something to AK 👀

PR #134, opened 2019-04-24
adding helper methods to AuthUtil for creating constraints

brianbushree:
### what
adding two methods:
- `AuthUtil.createSecuredConstraint(config, path)`
- `AuthUtil.createUnsecuredConstraint(config, path)`

& generalizing some `Constraint` creation code

### why
these methods would be useful for adding custom auth constraints

### DIFF
--- a/core/src/main/java/io/confluent/rest/auth/AuthUtil.java
+++ b/core/src/main/java/io/confluent/rest/auth/AuthUtil.java
@@ -27,6 +27,12 @@ public final class AuthUtil {
   private AuthUtil() {
   }
 
+  /**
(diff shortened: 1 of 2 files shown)


brianbushree:
anyone opposed to some extra helper methods here? 😁

brianbushree:
the use case here is when you don't want to use `createGlobalAuthConstraint()` because you want an arbitrary undefined URL to forward to a static resource without getting a 401

instead, you can use `createSecuredConstraint(...)` selectively on endpoints you want secure, leaving the everything else unsecured by default. (https://github.com/confluentinc/blueway/pull/1829/files#diff-132a1215b45674b1f310666a1a675204R344)

also, added tests!

PR #135, opened 2019-04-24
Add RestSecurityContext as @Context annotation on REST calls

spena:
This PR allows REST applications to inject a `RestSecurityContext` on each of the REST calls like this:
```
@POST
  public Response handleKsqlStatements(
      final @Context RestSecurityContext securityContext,
      final KsqlRequest request
  ) {}
```

This new context includes extra security information, such as Bearer tokens, so that we can use it for impersonation. 

The new context is registered by `RestSecurityContextBinder`, which is called by the `Application` class.  This registers the `RestSecurityContextFactory` which is called on every REST request to generate a new `RestSecurityContext`.

The `RestSecurityContextFactory` searches for the Bearer token and includes it into the `RestSecurityContext`.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -136,5 +136,10 @@
             <version>${guava.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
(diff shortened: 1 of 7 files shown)


spena:
This PR is not needed anymore. The application can get the Bearer token by using the `@HeaderParam("Authorization")`, and just parse the line to get the token. This PR was too much for a simple way to get it.

PR #136, opened 2019-04-25
add oauthbearer auth mechanism

brianbushree:
### what
Enable support for OAuthBearer but require user to implement `createAuthenticator()` & `createLoginService()`  

### why
having support for OAUTHBEARER would be nice (even without default implementations for an authenticator or login-service)

### how
`authentication.method=OAUTHBEARER`
then implement  `createAuthenticator()` & `createLoginService()`

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -19,6 +19,8 @@ package io.confluent.rest;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.jaxrs.base.JsonParseExceptionMapper;
 
+import io.confluent.common.security.jetty.JwtLoginService;
+import io.confluent.common.security.jetty.OAuthBearerAuthenticator;
 import io.confluent.rest.auth.AuthUtil;
 
 import org.apache.kafka.common.config.ConfigException;
@@ -378,6 +380,8 @@ public abstract class Application<T extends RestConfig> {
     String authMethod = config.getString(RestConfig.AUTHENTICATION_METHOD_CONFIG);
     if (enableBasicAuth(authMethod)) {
       context.setSecurityHandler(createBasicSecurityHandler());
+    } else if (enableOAuthBearerAuth(authMethod)) {
+      context.setSecurityHandler(createOAuthBearerSecurityHandler());
     }
   }
 
@@ -470,6 +474,10 @@ public abstract class Application<T extends RestConfig> {
     return RestConfig.AUTHENTICATION_METHOD_BASIC.equals(authMethod);
   }
 
+  static boolean enableOAuthBearerAuth(String authMethod) {
+    return RestConfig.AUTHENTICATION_METHOD_OAUTHBEARER.equals(authMethod);
+  }
+
   protected ConstraintSecurityHandler createBasicSecurityHandler() {
     final String realm = getConfiguration().getString(RestConfig.AUTHENTICATION_REALM_CONFIG);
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -209,6 +215,23 @@
...
+  public static final String AUTHENTICATION_OAUTHBEARER_PUBLIC_KEY_PATH =
+      "authentication.oauthbearer.public.key.path";
+  public static final String AUTHENTICATION_OAUTHBEARER_PUBLIC_KEY_PATH_DOC =
+      "Path to public key for authenticating JWT tokens.";
+
+  public static final String AUTHENTICATION_OAUTHBEARER_ROLES_CLAIM =
+      "authentication.oauthbearer.roles.claim";
+  public static final String AUTHENTICATION_OAUTHBEARER_ROLES_CLAIM_DOC =
+      "JWT roles claim.";
+  public static final String AUTHENTICATION_OAUTHBEARER_ROLES_CLAIM_DEFAULT = "clusters";
```
brianbushree: @rnpridgeon does this JWT roles claim default seem correct?

brianbushree:
closing since we don't need this at the moment.

brianbushree:
reopening since actually we do... 

either we need https://github.com/confluentinc/rest-utils/pull/139 OR this.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -470,14 +475,74 @@ static boolean enableBasicAuth(String authMethod) {
...
+      return new BasicAuthenticator();
+    } else if (enableOAuthBearerAuth(getConfiguration()
+                                     .getString(RestConfig.AUTHENTICATION_METHOD_CONFIG))) {
+      throw new UnsupportedOperationException(
+          "Must implement Application.createAuthenticator() when using '"
+          + RestConfig.AUTHENTICATION_METHOD_CONFIG + "="
+          + RestConfig.AUTHENTICATION_METHOD_OAUTHBEARER + "'."
+      );
+    } else if (realm != null && !realm.isEmpty()) {
+      // default case
```
norwood: feels like this should break instead of doing some default

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -470,14 +475,74 @@ static boolean enableBasicAuth(String authMethod) {
...
+      return new JAASLoginService(realm);
+    } else if (enableOAuthBearerAuth(getConfiguration()
+                                     .getString(RestConfig.AUTHENTICATION_METHOD_CONFIG))) {
+      throw new UnsupportedOperationException(
+          "Must implement Application.createLoginService() when using '"
+              + RestConfig.AUTHENTICATION_METHOD_CONFIG + "="
+              + RestConfig.AUTHENTICATION_METHOD_OAUTHBEARER + "'."
+      );
+    } else if (realm != null && !realm.isEmpty()) {
+      // default case
```
norwood: same here

omkreddy:
@brianbushree I am not sure, how to utilize these changes in open source rest projects like Kafka-rest, ksql etc. We can not add proprietary implementations to open source projects.
https://github.com/confluentinc/kafka-rest/blob/master/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestApplication.java#L47

brianbushree:
@omkreddy You're right. This is mainly for use with C3 and any other proprietary projects using rest-utils.

xvrl:
@brianbushree if we're not going to provide an implementation for this particular method, how about we allow arbitrary authentication method string? i.e. instead of hardcoding this particular string without an implementation, let an implementation decide whether a given auth method string is valid and provide hooks to construct the necessary auth bits for a given string.

brianbushree:
spoke offline with @xvrl . As suggested, I think a good follow up to this PR would be to let users add arbitrary auth mechanisms to work with pluggable authenticators/login-services

PR #139, opened 2019-04-30
adding OAUTHBEARER config

brianbushree:
### what
allow config `authentication.method=OAUTHBEARER`

### why
since `RestConfig` actually validates `authentication.method`, we can't just pass in `OAUTHBEARER` and expect it to work...

also, since we don't want to fully-support `OAUTHBEARER` in rest-utils, this is a hack...

fallback for this PR: https://github.com/confluentinc/rest-utils/pull/136


brianbushree:
this is a terrible approach. see https://github.com/confluentinc/rest-utils/pull/136

PR #141, opened 2019-05-09
CC-3633: Enable SSL client authentication to be requested but not required

C0urante:
This is part of our effort to enable anonymous access for Schema Registry.

The changes here add a configuration, `ssl.client.auth.requested`, which corresponds to a boolean value that's passed to the `setWantClientAuth(...)` method of the underlying `SslContextFactory` if and only if the `ssl.client.auth` configuration is not set to `true`. If both `ssl.client.auth` and `ssl.client.auth.requested` are set to `true`, the `ssl.client.auth` configuration is given precedence and a warning message is logged.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -468,7 +468,7 @@ public abstract class Application<T extends RestConfig> {
       } else {
         log.warn(
             "The configuration {} is deprecated and should be replaced with {}",
-            RestConfig.SSL_CLIENT_AUTH_DEFAULT,
+            RestConfig.SSL_CLIENT_AUTH_CONFIG,
             RestConfig.SSL_CLIENT_AUTHENTICATION_CONFIG
         );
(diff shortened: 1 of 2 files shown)


mageshn:
@C0urante Would it better to introduce a new config that takes the 2 values rather than the boolean and deprecate the existing config?

C0urante:
@mageshn I considered a new config. Not opposed to adding one; having two configs this similar in name might be confusing and the difference between them is non-obvious. Plus, it's not great to have four possible states with these two boolean configs when only three really make sense. I'll give that a shot, LMK what you think.

C0urante:
@mageshn alright, I've taken a stab at deprecating the current config in favor of a new one. LMK what you think.

C0urante:
@mageshn I believe I've already added the deprecation notice with https://github.com/confluentinc/rest-utils/pull/141/files#diff-81175690ead830577d721ce1cdee181cR187; is there somewhere else we should add that as well?

C0urante:
@mageshn I believe I've addressed your comment and will merge; if I missed something please let me know and I'll file a follow-up PR ASAP.

PR #142, opened 2019-05-09
SEC-191 - Upgrade Jetty in rest-utils

milosimpson:
- 9.4.14.v20181114 -> 9.4.18.v20190429

PR #143, opened 2019-05-10
Set default value for ssl.endpoint.identification.algorithm to https

omkreddy:
In newer [version](https://github.com/confluentinc/rest-utils/pull/142) of Jetty,  default value for ssl.endpoint.identification.algorithm was changed to https. This PR sets default value for ssl.endpoint.identification.algorithm to https. Also allow users to set empty string to restore the previous behaviour.

### DIFF
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -199,8 +199,8 @@ public class RestConfig extends AbstractConfig {
       "ssl.endpoint.identification.algorithm";
   protected static final String SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_DOC =
       "The endpoint identification algorithm to validate the server hostname using the "
-      + "server certificate. Leave blank to use Jetty's default.";
-  protected static final String SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_DEFAULT = "";
+      + "server certificate. Default algorithm is https.";
+  protected static final String SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_DEFAULT = "https";
 
   public static final String AUTHENTICATION_METHOD_CONFIG = "authentication.method";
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -199,8 +199,8 @@
       "ssl.endpoint.identification.algorithm";
   protected static final String SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_DOC =
       "The endpoint identification algorithm to validate the server hostname using the "
-      + "server certificate. Leave blank to use Jetty's default.";
-  protected static final String SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_DEFAULT = "";
+      + "server certificate. Default algorithm is https.";
+  protected static final String SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_DEFAULT = "https";
```
omkreddy: Another option is to support previous behavior i.e disable the host verification and allow users to override.

ijuma:
Thanks for the PR. What was the issue that triggered this PR?

omkreddy:
Some of the Rest proxy integration tests were failing after jetty upgrade and also current code does not allow to disable the host verification.

ijuma:
The only options are "https" and "", right?

omkreddy:
As per https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html, there are two standard algorithms:  "HTTPS",  "LDAPS"

core/src/main/java/io/confluent/rest/RestConfig.java
rajinisivaram: I think it is better to enable hostname verification as in the current PR since we already did this for Kafka clients and disabling it is insecure. We need to make sure we include this change in the upgrade notes though.

omkreddy:
@rajinisivaram Thanks for the review. Raised doc issue here: https://confluentinc.atlassian.net/browse/DOCS-1711

PR #145, opened 2019-06-20
SEC-276: Rename authentication method OAUTHBEARER to BEARER

rnpridgeon:


core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -205,7 +205,7 @@
   public static final String AUTHENTICATION_METHOD_CONFIG = "authentication.method";
   public static final String AUTHENTICATION_METHOD_NONE = "NONE";
   public static final String AUTHENTICATION_METHOD_BASIC = "BASIC";
-  public static final String AUTHENTICATION_METHOD_OAUTHBEARER = "OAUTHBEARER";
+  public static final String AUTHENTICATION_METHOD_BEARER = "BEARER";
   public static final String AUTHENTICATION_METHOD_DOC =
       "Method of authentication. Must be BASIC or OAUTHBEARER to enable authentication. "
```
jonchiu: do we need to update this to `BEARER`

rnpridgeon: Technically speaking OAUTHBEARER is a sasl mechanism not an http authorization method which makes this a bit inaccurate. It's a bit odd though since the BEARER auth scheme was registered with the adoption of the oauth framework bearer token usage. 

OAUTHBEARER mechanism
https://tools.ietf.org/html/rfc7628

Oauth framework: token usage 
https://tools.ietf.org/html/rfc6750

auth scheme registry: 
https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml

ewencp: @rnpridgeon I think the question was whether the docs are now just incorrect. the `_DOC` string seems to suggest you can still set to `OAUTHBEARER`, but the config only accepts the above methods, where you just replaced `OAUTHBEARER` with `BEARER`.

rnpridgeon: I see that now, 🤦‍♂ opening a PR now

PR #147, opened 2019-07-16
CC-5576: Use SslContextFactory.Server to disable server-side hostname verification

ijuma:
Jetty recently enabled hostname verification by default, but this
only makes sense for clients. They introduced a new subclass that
should be used in server contexts.


ijuma:
@rayokota I pushed an additional commit that partially reverts 4b0bcc441e8

rayokota:
LGTM for the partial revert

PR #148, opened 2019-07-22
Remove jackson databind version override

arvindth:
- This should get inherited from the common pom
  - Related PR: confluentinc/common#212

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,11 +126,6 @@
                 <version>${jetty.version}</version>
             </dependency>
             <dependency>
-                <groupId>com.fasterxml.jackson.core</groupId>
-                <artifactId>jackson-databind</artifactId>
-                <version>${jackson.version}</version>
-            </dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -125,11 +125,6 @@
                 <artifactId>jetty-jaas</artifactId>
                 <version>${jetty.version}</version>
             </dependency>
-            <dependency>
```
ewencp: presumably we still want the dependency, just not override the version?

arvindth: Actually we do want to remove the entire dependency here in the dependencymanagement section of the repo pom. Otherwise the dependency specified in the common pom gets overridden. The core submodule's pom still references the jackson-databind dependency, and inherits correctly from the common pom.

ewencp: ah, i missed that this was in dependencyManagement, I thought it was in dependencies.

PR #149, opened 2019-08-21
Include separate metrics for each HTTP error code

cyrusv:
This will allow us to distinguish between server errors and
user/input errors, which will let us specify alerting better.

Specifically, for projects like Schema Registry we may want to alert only on 5xx errors and not 4xx errors.

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0"?>
+
+<!DOCTYPE suppressions PUBLIC
+        "-//Puppy Crawl//DTD Suppressions 1.1//EN"
+        "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
+
+<suppressions>
+
+    <suppress checks="ClassDataAbstractionCoupling"
+              files="MetricsResourceMethodApplicationListener.java"/>
+</suppressions>
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -18,6 +18,7 @@ package io.confluent.rest.metrics;
 
 import io.confluent.common.metrics.stats.Percentile;
 import io.confluent.common.metrics.stats.Percentiles;
+import io.confluent.rest.exceptions.RestException;
 import org.glassfish.jersey.server.ContainerRequest;
 import org.glassfish.jersey.server.ContainerResponse;
 import org.glassfish.jersey.server.model.Resource;
@@ -60,6 +61,7 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
 
   public static final String REQUEST_TAGS_PROP_KEY = "_request_tags";
 
+  private static final String ERROR_CODE_TAG_KEY = "error_code";
   private static final int PERCENTILE_NUM_BUCKETS = 200;
   private static final double PERCENTILE_MAX_LATENCY_IN_MS = TimeUnit.SECONDS.toMillis(10);
 
@@ -173,6 +175,7 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
     private Sensor responseSizeSensor;
     private Sensor requestLatencySensor;
     private Sensor errorSensor;
+    private Sensor[] errorSensorByStatus;
 
     public MethodMetrics(ResourceMethod method, PerformanceMetric annotation, Metrics metrics,
                          String metricGrpPrefix, Map<String, String> metricTags) {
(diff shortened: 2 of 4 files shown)


core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -236,9 +244,64 @@ public MethodMetrics(ResourceMethod method, PerformanceMetric annotation, Metric
...
+      tags.put(ERROR_CODE_TAG_KEY, "100");
+      metricName = new MetricName(getName(method, annotation, "request-error-rate"),
+          metricGrpName,
+          "The average number of requests per second that resulted in HTTP error responses",
+          tags);
+      this.errorSensor100s.add(metricName, new Rate());
+
+      this.errorSensor200s = metrics.sensor(getName(method, annotation, "errors"));
+      tags = new HashMap<>(metricTags);
+      tags.put(ERROR_CODE_TAG_KEY, "200");
```
mageshn: may have the tags as 1xx, 2xx, 3xx, etc.

checkstyle/suppressions.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+
+<!DOCTYPE suppressions PUBLIC
+        "-//Puppy Crawl//DTD Suppressions 1.1//EN"
+        "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
+
+<suppressions>
+
+    <suppress checks="JavaNCSS"
+              files="MetricsResourceMethodApplicationListener.java"/>
```
ewencp: It looks like rather than suppressing this, we should refactor the code that's repeated 6 or 7 times into a method that we can reuse with only one or two parameters.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -256,7 +319,24 @@ public void finished(long requestSize, long responseSize, long latencyMs) {
     /**
      * Indicate that a request has failed with an exception.
      */
-    public void exception() {
+    public void exception(final RequestEvent event) {
+      int errorCode = -1;
+      if (event.getException() instanceof RestException) {
```
mageshn: why not read the status of the response in the event? In which case, we would never have an unknown scenario.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -236,9 +244,64 @@ public MethodMetrics(ResourceMethod method, PerformanceMetric annotation, Metric
...
+          "The average number of requests per second that resulted in HTTP error responses",
+          tags);
+      this.errorSensor300s.add(metricName, new Rate());
+
+      this.errorSensor400s = metrics.sensor(getName(method, annotation, "errors"));
+      new HashMap<>(metricTags);
+      tags.put(ERROR_CODE_TAG_KEY, "400");
+      metricName = new MetricName(getName(method, annotation, "request-error-rate"),
+          metricGrpName,
+          "The average number of requests per second that resulted in HTTP error responses",
```
mageshn: I'm not sure how the description is used but I would prefer to update the description for each status code if it's safer to do so.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -256,7 +319,24 @@ public void finished(long requestSize, long responseSize, long latencyMs) {
...
+        errorSensorUnknown.record();
+      } else if (errorCode < 200) {
+        errorSensor100s.record();
+      } else if (errorCode < 300) {
+        errorSensor200s.record();
+      } else if (errorCode < 400) {
+        errorSensor300s.record();
+      } else if (errorCode < 500) {
+        errorSensor400s.record();
+      } else {
```
ewencp: i can't imagine it'd work anyway, but `>=600` should be `errorSensorUnknown` as well (same as `<0`, which I also can't imagine wouldn't work).

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -173,6 +175,12 @@ public ConstructionContext(
     private Sensor responseSizeSensor;
     private Sensor requestLatencySensor;
     private Sensor errorSensor;
+    private Sensor errorSensorUnknown;
+    private Sensor errorSensor100s;
```
ewencp: you could refactor these into a simple array, which makes the code for deciding which sensor it should go an integer division and fewer conditionals. we should take care about performance in this code since it has overhead per-request (at least, in this case, for error cases).

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
ewencp: If you're looking for classifying http status codes, `RestException` and `getErrorCode()` aren't quite right. `RestException` inherits from the more general `WebApplicationException`. The error code is an extension that `rest-utils` puts in place to allow more granularity and standardize the output format for the body of errors to include the more detailed version. e.g. if schema registry doesn't find a subject, the status is 404, but the error code is 40401 so you can classify the issue more specifically (in an automated way, without relying on the text of the error message).

I don't know the details of the motivation here, but you probably want to use `WebApplicationException` and `getStatus()`. If not, you have to deal with more of range of errors since most exceptions thrown by actual rest-utils applications use error codes with 5 digits (3 for http status, 2 additional to classify with more granularity), so only things like framework-provided errors would be covered by this (which works fine for everything other than 4-500 ranges, but presumably those are the ones you're interested in granular metrics for generally).

mageshn: Why not look at the Response status which has the https status rather than evaluating the Exception?

ewencp: Sure, that works too. Key point being that we want status, not error code.

cyrusv: I think the best strategy here is to use `WebApplicationException..getResponse().getStatus()`. It is as general as we can get and still pretty clean. Sound good?

checkstyle/suppressions.xml
cyrusv: Even with the refactor, I'm instantiating too many classes and getting "MetricsResourceMethodApplicationListener". This is probably fine, right? I'm not really fundamentally changing the structure of the code, just putting some of the sensors into an array.

cyrusv:
Thanks @mageshn they are addressed

PR #150, opened 2019-09-25
Ability to customize the thread pool and size of request queue.

jeffhuang26:
This is first version for preview, do NOT merge it. I will add unit testing later.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -50,6 +50,9 @@ import org.eclipse.jetty.servlet.ServletHolder;
 import org.eclipse.jetty.servlets.CrossOriginFilter;
 import org.eclipse.jetty.util.resource.ResourceCollection;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.eclipse.jetty.util.thread.QueuedThreadPool;
+import org.eclipse.jetty.util.BlockingArrayQueue;
+import org.eclipse.jetty.util.thread.ThreadPool;
 import org.eclipse.jetty.websocket.jsr356.server.ServerContainer;
 import org.eclipse.jetty.websocket.jsr356.server.deploy.WebSocketServerContainerInitializer;
 import org.glassfish.jersey.server.ResourceConfig;
@@ -73,6 +76,7 @@ import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
+import java.util.concurrent.BlockingQueue;
 
 import javax.servlet.DispatcherType;
 import javax.servlet.ServletException;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -266,7 +266,28 @@
   public static final String IDLE_TIMEOUT_MS_CONFIG = "idle.timeout.ms";
   public static final String IDLE_TIMEOUT_MS_DOC =
           "The number of milliseconds to hold an idle session open for.";
-  public static final long IDLE_TIMEOUT_MS_DEFAULT = 30_000;
+  public static final long IDLE_TIMEOUT_MS_DEFAULT = 60_000;
```
milosimpson: why is the idle timeout default changing?

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -783,5 +790,27 @@ private void doShutdown() {
    */
   public void onShutdown() {
   }
+
+  /**
+   * Create the thread pool with request queue.
+   *
+   * @return thread pool used by the server
+   */
+  private ThreadPool createThreadPool() {
```
milosimpson: I think maybe REQUEST_QUEUE_CAPACITY_INITIAL_DEFAULT should be like -1, so that it continues to use an UnboudedQueue, instead of a bounded queue with bound of INT_MAX.

core/src/main/java/io/confluent/rest/RestConfig.java
jeffhuang26: The default value for idle timeout is 60_000 in Jetty. I do not know why it was 30_000 in rest-utils. So I changed it back to 60_000.

core/src/main/java/io/confluent/rest/Application.java
jeffhuang26: I think you mean REQUEST_QUEUE_CAPACITY_GROWBY_DEFAULT which controls how much capacity will be increased in need. The default value growby is Math.max(_minThreads, 8) in Jetty if user create Jetty server like new Server(). I put this default value to REQUEST_QUEUE_CAPACITY_GROWBY_DEFAULT = 64. I think we should be fine. 

The unbound queue will be controlled by both maximum capacity and growby. Current default maximum capacity is REQUEST_QUEUE_CAPACITY_DEFAULT = Integer.MAX_VALUE is same as Jetty one. So the default queue is unbounded.

core/src/main/java/io/confluent/rest/RestConfig.java
jeffhuang26: Changed back to 30_000 due I think this might break other REST application.

jeffhuang26:
In HTTP client side will see:
The requested URL can't be reached
The service might be temporarily down or it may have moved permanently to a new web address
connect ECONNREFUSED 127.0.0.1:8090

jeffhuang26:
Milo,

Please review and approve this PR when get time. This PR has been there for a while. I think we cover all testing cases we had discussed.

jeffhuang26:
PR 172 replaced this PR.

PR #151, opened 2019-10-21
Update Jetty version to 9.4.20.v20190813 and Jersey version to 2.28

omkreddy:


core/src/test/java/io/confluent/rest/SaslTest.java
```diff
@@ -163,14 +163,14 @@ public void testNoAuthAttemptOnWs() throws Exception {
   @Test
   public void testBadLoginAttempt() throws Exception {
     try (CloseableHttpResponse response =
-        makeGetRequest("/test", "this shouldnt work")) {
+        makeGetRequest("/test", "dGVzdA==")) {
```
milosimpson: what is dGVzdA== ?

omkreddy: Its string “test” encoded in base64 format.  In newer version, jetty basic auth validates base63 string format. current string is not a valid base64 string,

omkreddy:
Thanks for the reviews. Merging to master.

PR #153, opened 2019-10-23
Revert Jersey to v2.27

spena:
REST applications (especially in KSQL) were failing to start due to the previous Jersy v2.28. Reverting version to v2.27 fixes the incompatibility issue.


spena:
Closing this PR as the issue will be fixed in another project.

PR #154, opened 2019-10-23
Add error log with thrown exception

Prathibha-m:


core/src/main/java/io/confluent/rest/exceptions/DebuggableExceptionMapper.java
```diff
@@ -56,6 +59,7 @@ public DebuggableExceptionMapper(RestConfig restConfig) {
    */
   public Response.ResponseBuilder createResponse(Throwable exc, int errorCode,
                                                  Response.Status status, String msg) {
+    log.error("Debuggable Exception Response " + exc);
```
mageshn: We don't typically concatenate the exception object 
```suggestion
    log.error("Request Failed with exception " ,  exc);
```

PR #155, opened 2019-11-01
Enable Rest Utils to provide multiple servlet contexts

rnpridgeon:


rnpridgeon:
sorry @bashburn , accidentally tagged you on the original

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2014 Confluent Inc.
```
jnaaman: nit: do we change copyright year when touching older files?

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -197,20 +200,18 @@ public Server createServer() throws ServletException {
     // The configuration for the JAX-RS REST service
     ResourceConfig resourceConfig = new ResourceConfig();
 
+    Map<String, String> combinedMetricsTags = new HashMap<>();
+    for (ApplicationContext ctx : appCtx) {
+      combinedMetricsTags.putAll(ctx.getMetricsTags());
```
jnaaman: Is it possible two different context may try to add metrics with the same key? If so, do we want to error out or at least log some warning?

core/src/main/java/io/confluent/rest/Application.java
rnpridgeon: I actually have no idea @ewen?

core/src/main/java/io/confluent/rest/Application.java
rnpridgeon: It will actually fail if you have duplicate tags on registration.

rnpridgeon: Ill add a test though good thinking

rnpridgeon: Actually you are right since I'm adding these tags before registration. Good catch fixed

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -256,52 +256,47 @@ protected void doStop() throws Exception {
...
+    for (ApplicationContext ctx : appCtx) {
 
-    if (isCorsEnabled()) {
-      String allowedOrigins = config.getString(RestConfig.ACCESS_CONTROL_ALLOW_ORIGIN_CONFIG);
-      FilterHolder filterHolder = new FilterHolder(CrossOriginFilter.class);
-      filterHolder.setName("cross-origin");
-      filterHolder.setInitParameter(
-          CrossOriginFilter.ALLOWED_ORIGINS_PARAM, allowedOrigins
+      // Isolate contexts by making a defensive copy of the application config
+      ResourceConfig ctxConfig = ctx.ctxConfig;
```
rigelbm: 1. I can't see the copy being made.
2. What is the difference between config and ctxConfig? Things are getting pretty confusing for me here when I read it.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -783,5 +774,41 @@ private void doShutdown() {
    */
   public void onShutdown() {
   }
+
+  /**
+   * By default the RestUtils will expose one context the DefaultContext
+   */
+  private class DefaultContext extends ApplicationContext<T> {
```
rigelbm: final

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -256,52 +256,47 @@ protected void doStop() throws Exception {
...
+      FilterHolder servletHolder = new FilterHolder(servletContainer);
+      ctx.addFilter(servletHolder, "/*", null);
+
+      ctx.configurePostResourceHandling();
+      ServletHolder defaultHolder = new ServletHolder("default", DefaultServlet.class);
+      defaultHolder.setInitParameter("dirAllowed", "false");
+      ctx.addServlet(defaultHolder, "/*");
+
+      applyCustomConfiguration(ctx, REST_SERVLET_INITIALIZERS_CLASSES_CONFIG);
+      handlers.addHandler(ctx);
```
rigelbm: Could all this block be in ApplicationContext, and do handlers.addHandler(ctx.configureHandler()) or something similar?

core/src/main/java/io/confluent/rest/Application.java
rnpridgeon: That actually got backed out after I decided to give each context its own ResourceConfig. 

Yes `config` is the application config (RestConfig) ctxConfig is the ResourceConfig (Jersey)

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -187,30 +183,34 @@ protected void configureWebSocketPostResourceHandling(ServletContextHandler cont
...
 
-    // The configuration for the JAX-RS REST service
-    ResourceConfig resourceConfig = new ResourceConfig();
+    Map<String, String> combinedMetricsTags = new HashMap<>();
+    for (ApplicationContext ctx : appCtx) {
+      Map<String, String> metrics = ctx.getMetricsTags();
+      metrics.forEach((key, value) -> {
+        String found = combinedMetricsTags.put(key, value);
+        if (found != null) {
+          log.warn("Found duplicate metric tag {}, ignoring", found);
```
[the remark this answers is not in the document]
rnpridgeon: I actually decided to throw the same exception as the Metric class here as it would on registration.

core/src/main/java/io/confluent/rest/ApplicationGroup.java
```diff
@@ -0,0 +1,86 @@
...
+
+  public void addApplication(Application application) {
+    this.applications.add(application);
+  }
+
+  public List<Application> getApplications() {
+    return applications;
+  }
+
+  public RestConfig getConfiguration() {
```
rigelbm: I'm confused to why an group has a configuration. I was thinking that only `Application` and `ApplicationServer` have configurations, for app specific settings and global settings respectively. `ApplicationGroup` itself is not doing anything with the configuration, so I guess it does not belong here.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -191,74 +189,26 @@ protected void configureWebSocketPostResourceHandling(ServletContextHandler cont
...
+  public Server createServer() throws Exception {
     // CHECKSTYLE_RULES.ON: MethodLength|CyclomaticComplexity|JavaNCSS|NPathComplexity
+    if (server == null) {
+      new ApplicationServer(new ApplicationGroup(config,this));
+    }
+    return server;
+  }
 
-    // The configuration for the JAX-RS REST service
+  Handler configureHandler() {
```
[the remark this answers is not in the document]
rnpridgeon: Out of curiosity what is the advantage of marking a package-private method final. My understanding was that with the current access level subclasses won't be able to overwrite this method anyway.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -387,139 +323,34 @@ private void applyCustomConfiguration(
         });
   }
 
-  protected void configureSecurityHandler(ServletContextHandler context) {
+  static void configureSecurityHandler(ServletContextHandler context, RestConfig config) {
```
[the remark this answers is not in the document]
rnpridgeon: This is actually leftover from when ApplicationContext was an inner class to Application. In the end I decided to keep these static variants to offer some flexibility in code reuse. My thinking was that eventually we can decouple the server logic from the application logic entirely allowing Applications to use a generic of type `AbstractConfig` rather than the more restrictive RestConfig type. I don't have strong feelings either way though so we can revert it.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -387,139 +323,34 @@ private void applyCustomConfiguration(
...
-            "Unexpected value for {} configuration: {}",
-            RestConfig.SSL_CLIENT_AUTHENTICATION_CONFIG,
-            clientAuthentication
-        );
-    }
+  protected void configureSecurityHandler(ServletContextHandler context) {
+    configureSecurityHandler(context, this.config);
   }
 
-  public Handler wrapWithGzipHandler(Handler handler) {
```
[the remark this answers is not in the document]
rnpridgeon: you are right, wasn't thinking there

core/src/main/java/io/confluent/rest/ApplicationGroup.java
```diff
@@ -0,0 +1,86 @@
...
+
+public class ApplicationGroup<T extends RestConfig> {
+  private final T config;
+  private Server server = null;
+  private final List<Application> applications = new ArrayList<>();
+
+  ApplicationGroup(T config, Application ...applications) {
+    this.config = config;
+    this.applications.addAll(Arrays.asList(applications));
+    System.out.println(applications.toString());
```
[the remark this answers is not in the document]
rnpridgeon: 🤦‍♂

core/src/main/java/io/confluent/rest/ApplicationGroup.java
rnpridgeon: There is nothing preventing App1 and App2 from having differing configurations with regard to how the shared resource, the ApplicationServer itself, is configured. Rather than picking adding some logic to pick a winner I make the developer explicitly decide which configuration will be used to setup the ApplicaitonServer.

core/src/main/java/io/confluent/rest/ApplicationGroup.java
```diff
@@ -0,0 +1,86 @@
...
+  /* Visible for testing */
+  Server getServer() {
+    return this.server;
+  }
+
+  /**
+   * Start the server (creating it if necessary).
+   * @throws Exception If the application fails to start
+   */
+  public void start() throws Exception {
```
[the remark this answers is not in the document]
rnpridgeon: Yeah it just felt weird to leave them off because the individual applications expose life cycle methods. You are right though we should remove it from the group so we can position ourselves to remove them from the group and deprecate the application lifecycle methods to move rest-utils closer to where we want it to be in the future. .

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -764,7 +557,7 @@ public void stop() throws Exception {
     server.stop();
   }
 
-  private void doShutdown() {
+  void doShutdown() {
```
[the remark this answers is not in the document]
rnpridgeon: As long as we have resource extensionsI think we have to contend with at least some lifecycle management. Package-protected in conjunction with final seems like a decent compromise for v1. In the long run I question if resource extensions are still needed since we can add multiple applications to a server now.

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -0,0 +1,368 @@
...
+      connector.setHost(listener.getHost());
+      connector.setIdleTimeout(config.getLong(RestConfig.IDLE_TIMEOUT_MS_CONFIG));
+
+      connectors.add(connector);
+      super.addConnector(connector);
+
+    }
+  }
+
+  public  List<URL>  getListeners() {
```
rigelbm: nit: there are some extra white spaces here.

PR #156, opened 2019-11-05
Enable auto ssl keystore reload

yanglei99:
https://confluentinc.atlassian.net/browse/MCM-462

Here is the one pager: https://confluentinc.atlassian.net/wiki/spaces/MCM/pages/938970260/One+Pager+Auto+Reload+SSL+Certificate+for+Http+Endpoints#OnePager:AutoReloadSSLCertificateforHttpEndpoints-CPComponentsusingrest-util 

Here is the related logic in kafka ( using WatchService)

https://github.com/confluentinc/ce-kafka/blob/cc608b2d0856e9eee0b0f0faef280a3572f8e20d/ce-broker-plugins/src/main/java/io/confluent/kafka/multitenant/PhysicalClusterMetadata.java#L508

Here is the end-end test for kafka-api:

https://github.com/confluentinc/cc-spec-kafka/pull/54


core/src/main/java/io/confluent/rest/FileWatcher.java
```diff
@@ -0,0 +1,110 @@
...
+    try {
+      for (;;) {
+        log.info("Watching file change: " + file);
+        // wait for key to be signalled
+        WatchKey key = watchService.take();
+        if (this.key != key) {
+          log.info("WatchKey not recognized");
+          continue;
+        }
+        log.info("Watch Key notified");
```
norwood: may want many of these `.info` to be `.debug`

core/src/main/java/io/confluent/rest/FileWatcher.java
```diff
@@ -0,0 +1,110 @@
...
+              log.info("Watch found file: " + file);
+              try {
+                callback.run();
+              } catch (Exception e) {
+                log.error("Error while reloading cert", e);
+              }          
+              break;
+            }
+          } catch (java.io.IOException e) {
+            log.info("Hit error process the change event", e);
```
norwood: and this should probably be warn instead of info.

core/src/main/java/io/confluent/rest/FileWatcher.java
```diff
@@ -0,0 +1,110 @@
...
+          }
+          WatchEvent<Path> ev = (WatchEvent<Path>)event;
+          Path changed = this.file.getParent().resolve(ev.context());
+          try {
+            if (Files.isSameFile(changed, this.file)) {
+              log.info("Watch found file: " + file);
+              try {
+                callback.run();
+              } catch (Exception e) {
+                log.error("Error while reloading cert", e);
```
norwood: feels weird for this to specifically mention cert reloading, but the class is written to be generic

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -120,6 +120,10 @@
       + "details, etc.";
   protected static final String METRICS_TAGS_DEFAULT = "";
 
+  public static final String SSL_KEYSTORE_RELOAD_CONFIG = "ssl.keystore.reload";
+  protected static final String SSL_KEYSTORE_RELOAD_DOC =
+      "Enable auto reload of ssl keystore";
+  protected static final boolean SSL_KEYSTORE_RELOAD_DEFAULT = false;
```
norwood: this seems like a great addition. why do you want this default false?

core/src/main/java/io/confluent/rest/FileWatcher.java
```diff
@@ -0,0 +1,110 @@
...
+        // wait for key to be signalled
+        WatchKey key = watchService.take();
+        if (this.key != key) {
+          log.info("WatchKey not recognized");
+          continue;
+        }
+        log.info("Watch Key notified");
+        for (WatchEvent<?> event : key.pollEvents()) {
+          WatchEvent.Kind<?> kind = event.kind();
+          if (kind != StandardWatchEventKinds.ENTRY_MODIFY) {
```
norwood: would this miss a delete followed by a create?

since we already know `this.file`, is there anything to be gained by doing all these checks anyhow? can we not just jump straight to `callback.run()` and lfg?

core/src/main/java/io/confluent/rest/FileWatcher.java
```diff
@@ -0,0 +1,110 @@
...
+              }          
+              break;
+            }
+          } catch (java.io.IOException e) {
+            log.info("Hit error process the change event", e);
+          }
+        }
+        // reset key
+        if (!key.reset()) {
+          break;
```
norwood: this seems to be killing the watcher, so probably a good place for a `.info("Ending my watch")` too.

core/src/main/java/io/confluent/rest/FileWatcher.java
yanglei99: As we only listen to MODIFY, the only thing I need to check is `OVERFLOW – Indicates that events might have been lost or discarded. You do not have to register for the OVERFLOW event to receive it.`. I can make it specific.

core/src/main/java/io/confluent/rest/FileWatcher.java
```diff
@@ -0,0 +1,111 @@
...
+import java.nio.file.WatchService;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchEvent;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+// reference https://gist.github.com/danielflower/f54c2fe42d32356301c68860a4ab21ed
+public class FileWatcher implements Runnable {
```
xvrl: what do we use in ce-kafka to auto-reload ssl certs, can we reuse some of the code there?

core/src/main/java/io/confluent/rest/FileWatcher.java
```diff
@@ -0,0 +1,111 @@
...
+            log.warn("Hit error process the change event", e);
+          }
+        }
+        // reset key
+        if (!key.reset()) {
+          log.warn("Ending watch due to key reset error");
+          break;
+        }
+      }
+    } catch (InterruptedException e) {
```
xvrl: what if any other kind of exception happens, wouldn't that kill the watcher thread and prevent future reloads?

core/src/main/java/io/confluent/rest/FileWatcher.java
```diff
@@ -0,0 +1,111 @@
...
+
+  public FileWatcher(Path file, Callback callback) throws IOException {
+    this.file = file;
+    this.watchService = FileSystems.getDefault().newWatchService();
+    this.key = file.getParent().register(watchService, StandardWatchEventKinds.ENTRY_MODIFY);
+    this.callback = callback;
+  }
+
+  /**
+    * Starts watching a file and the given path and calls the callback when it is changed.
```
xvrl: what do we mean by "file and the given path", do we mean a file or a directory?

core/src/test/java/io/confluent/rest/SslTest.java
```diff
@@ -69,6 +71,7 @@
 
   public static final String SSL_PASSWORD = "test1234";
   public static final String EXPECTED_200_MSG = "Response status must be 200.";
+  public static final int CERT_RELOAD_WAIT_TIME = 20000;
```
xvrl: mind adding a comment to explain the purpose of this timeout?

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -416,6 +418,16 @@ private SslContextFactory createSslContextFactory() {
         sslContextFactory.setKeyManagerFactoryAlgorithm(
             config.getString(RestConfig.SSL_KEYMANAGER_ALGORITHM_CONFIG));
       }
+      if (config.getBoolean(RestConfig.SSL_KEYSTORE_RELOAD_CONFIG)) {
+        Path keystorePath = Paths.get(URI.create(sslContextFactory.getKeyStorePath()));
+        try {
+          FileWatcher.onFileChange(keystorePath, () -> 
+              sslContextFactory.reload(scf -> log.info("Reloaded SSL cert")));
+          log.info("Enabled SSL cert auto reload: " + keystorePath);
+        } catch (java.io.IOException e) {
```
xvrl: do we know when this might fail?

yanglei99: As it is IOException, it means the file can not be loaded or something like that.

core/src/main/java/io/confluent/rest/FileWatcher.java
yanglei99: I checked with @aashishkohli , she mentioned Kafka is using library from open source kafka repo... I will let her add more input.

core/src/main/java/io/confluent/rest/FileWatcher.java
yanglei99: Changed most of the info to debug. But want to keep this notified...

core/src/main/java/io/confluent/rest/FileWatcher.java
yanglei99: Changed.

core/src/main/java/io/confluent/rest/FileWatcher.java
yanglei99: All the methods we call now do not have exception than InterruptedException, which happens a thread is interrupted

core/src/main/java/io/confluent/rest/RestConfig.java
norwood: aren't your changes backwards compatible? 

i mean, i guess not if people are relying on the ability to change the keystore file and have nothing happen. but that seems like a crazy thing to do in the first place.

yanglei99:
[clabot:check]

core/src/main/java/io/confluent/rest/RestConfig.java
yanglei99: revised, thanks!

core/src/main/java/io/confluent/rest/FileWatcher.java
yanglei99: revised the comment

core/src/main/java/io/confluent/rest/FileWatcher.java
xvrl: we should still have a catch for any unexpected runtime exceptions, otherwise the thread will stop

core/src/main/java/io/confluent/rest/FileWatcher.java
```diff
@@ -0,0 +1,114 @@
...
+        WatchKey key = watchService.take();
+        log.info("Watch Key notified");
+        for (WatchEvent<?> event : key.pollEvents()) {
+          WatchEvent.Kind<?> kind = event.kind();
+          if (kind == StandardWatchEventKinds.OVERFLOW) {
+            log.debug("Watch event is OVERFLOW");
+            continue;
+          }
+          WatchEvent<Path> ev = (WatchEvent<Path>)event;
+          Path changed = this.file.getParent().resolve(ev.context());
```
xvrl: nit, we don't seem to be consistent about `file` vs. `this.file`. Typically the distinction is only needed if there is ambiguity between a local variable and a member variable

core/src/main/java/io/confluent/rest/FileWatcher.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /**
+    * Starts watching a file calls the callback when it is changed.
+    * A shutdown hook is registered to stop watching.
+  */
+  public static void onFileChange(Path file, Callback callback) throws IOException {
+    log.info("Configure watch file change: " + file);
+    FileWatcher fileWatcher = new FileWatcher(file, callback);
+    ExecutorService executor = Executors.newSingleThreadExecutor();
```
xvrl: let's make sure this is a daemon thread, so it doesn't prevent the service from shutting down. Then you can also remove the shutdownhook

core/src/main/java/io/confluent/rest/FileWatcher.java
```diff
@@ -0,0 +1,114 @@
...
+    Future<?> future = executor.submit(fileWatcher);
+    Runtime.getRuntime().addShutdownHook(new Thread(executor::shutdownNow));
```
xvrl: shutdown hook is not necessary if you make this a daemon thread.

yanglei99: FileWatcher is a Runnable... it is using newSingleThreadExecutor to run it... which uses w worker thread... https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html#newSingleThreadExecutor()

Issue #158, opened 2019-11-12
NPE in MetricsRequestEventListener when @PreMatching ContainerRequestFilter throws exception.


fabiohecht:
I registered a @PreMatching ContainerRequestFilter via kafka.rest.resource.extension.class and sometimes need to throw exceptions that are mapped to responses. This works fine (client gets the correct response), but the MetricsRequestEventListener logs a NPE.

Root cause: The event RequestEvent.Type.MATCHING_START is never invoked in this case (exception is thrown in PreMatching filter), thus never initializing the wrappedRequestStream field. Later, handling the FINISHED event assumes wrappedRequestStream is set, throwin the NPE.

I believe MetricsRequestEventListener must be patched to initialize the `wrappedRequestStream` and `started` without depending on MATCHING_START event, because in this case it is never triggered.

I'm thinking about the following alternative to fix the issue: Replace the condition `if (event.getType() == RequestEvent.Type.MATCHING_START)` by `if (started == 0L)` or `if (wrappedRequestStream == null)` (at https://github.com/confluentinc/rest-utils/blob/5.3.1-post/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java#L301) to initialize variables the first time any event is triggered.

I'll open a PR.

Cheers,

Fabio

junzebao:
Same issue happened here. Anyone to review the PR?

PR #159, opened 2019-11-12
Fix NPE in MetricsRequestEventListener when @PreMatching ContainerRequestFilter throws exception

fabiohecht:
I registered a @PreMatching ContainerRequestFilter via kafka.rest.resource.extension.class and sometimes need to throw exceptions that are mapped to responses. This works fine (client gets the correct response), but the MetricsRequestEventListener logs a NPE.

Root cause: The event RequestEvent.Type.MATCHING_START is never invoked in this case (exception is thrown in PreMatching filter), thus never initializing the wrappedRequestStream field. Later, handling the FINISHED event assumes wrappedRequestStream is set, throwin the NPE.

I believe MetricsRequestEventListener must be patched to initialize the wrappedRequestStream and started without depending on MATCHING_START event, because in this case it is never triggered.

This PR simply replaces the condition if (event.getType() == RequestEvent.Type.MATCHING_START) by if (started == 0L) or if (wrappedRequestStream == null) (at https://github.com/confluentinc/rest-utils/blob/5.3.1-post/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java#L301) to initialize variables the first time any event is triggered.

### DIFF
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -325,12 +325,13 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
 
     @Override
     public void onEvent(RequestEvent event) {
-      if (event.getType() == RequestEvent.Type.MATCHING_START) {
+      if (started == 0L) {
         started = time.milliseconds();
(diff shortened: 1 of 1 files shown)


fabiohecht:
[clabot:check]

fabiohecht:
Fix issue #158

PR #167, opened 2019-12-06
updated registerApplication to work after jetty starts

atxalex8:
This PR is to allow the addition of `Application` to a running Jetty `ApplicationServer`. 
The use case is to allow Kafka to start up a Jetty ApplicationServer (MetadataServer) in `ce-kafka` and manage it's lifecycle while other MDS modules (RBAC, config, etc) can register themselves to this jetty server. 
1. KafkaServer starts Jetty HTTP
2. RbacProvider loads rbac and builds rbac application, then registers to Jetty
3. ConfigProvider loads config settings and builds application, then registers to Jetty
4. Other modules do the same, obviously with different context paths.

### DIFF
--- a/core/src/test/java/io/confluent/rest/ApplicationGroupTest.java
+++ b/core/src/test/java/io/confluent/rest/ApplicationGroupTest.java
@@ -116,6 +116,20 @@ public class ApplicationGroupTest {
     assertThat(makeGetRequest("/app2/index.html"), is(Code.OK));
   }
 
+  /* Test Dynamic App Loading */
+  @Test
+  public void testDynamicAppLoading() throws Exception {
+    TestApp app1 = new TestApp("/app1");
+    TestApp app2 = new TestApp("/app2");
+
+    server.registerApplication(app1);
+    server.start();
+    server.registerApplication(app2);
+
+    assertThat(makeGetRequest("/app1/resource"), is(Code.OK));
+    assertThat(makeGetRequest("/app2/resource"), is(Code.OK));
+  }
+
   @SuppressWarnings("SameParameterValue")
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/confluent/rest/ApplicationGroupTest.java
```diff
@@ -116,6 +116,20 @@ protected ResourceCollection getStaticResources() {
...
+  public void testDynamicAppLoading() throws Exception {
+    TestApp app1 = new TestApp("/app1");
+    TestApp app2 = new TestApp("/app2");
+
+    server.registerApplication(app1);
+    server.start();
+    server.registerApplication(app2);
+
+    assertThat(makeGetRequest("/app1/resource"), is(Code.OK));
+    assertThat(makeGetRequest("/app2/resource"), is(Code.OK));
```
milosimpson: just to be really pedantic I would also want to verify that things aren't started early.

```
server.registerApplication(app1);
// call and get no response from app1

 server.start();
    assertThat(makeGetRequest("/app1/resource"), is(Code.OK));
// verify app2 does not start

    server.registerApplication(app2);

    // app2 is not good
    assertThat(makeGetRequest("/app2/resource"), is(Code.OK));
    // app 2 still works
    assertThat(makeGetRequest("/app1/resource"), is(Code.OK));
```

rigelbm:
Can you please write a description to this PR explaining why you doing it?

atxalex8:
[clabot:check]

rigelbm:
Hi @atxalex8. Could you please expand your PR description. It is still not clear to me why this PR is necessary. Why would it be nice to add applications after Jetty startup? What is the specific use case or problem you are trying to solve? Thanks!

PR #169, opened 2020-01-23
Add OIDC authentication

rnpridgeon:


rnpridgeon:
http://ec2-54-193-21-243.us-west-1.compute.amazonaws.com:8080 < auth0

pom.xml
```diff
@@ -51,7 +51,7 @@
         <apache.httpclient.version>4.5.3</apache.httpclient.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
         <jersey.version>2.28</jersey.version>
-        <jetty.version>9.4.20.v20190813</jetty.version>
+        <jetty.version>9.4.23.v20191118</jetty.version>
```
brianbushree: is there a reason we're bumping?

rnpridgeon: so we don't have to write our own OIDC authenticator

rnpridgeon: This was the earliest version which includes it so I'm hopeful it has a lesser chance of breaking things unexpectedly.

examples/src/main/java/io/confluent/rest/examples/hellooidc/HelloOpenID.java
```diff
@@ -0,0 +1,95 @@
...
+    config.register(new HelloWorldResource(appConfig));
+    config.property(ServletProperties.FILTER_STATIC_CONTENT_REGEX, "/(static/.*|.*\\.html|)");
+  }
+
+  @Override
+  protected ResourceCollection getStaticResources() {
+    return new ResourceCollection(Resource.newClassPathResource("static"));
+  }
+
+  public static void main(String[] args) {
```
milosimpson: So looks like this was manually tested?

I.e. you can use this example  call the main method and pass params.

Can you comment in this file or in the commit message
* An example of how you ran this, (against Okta)?
* What the expected behavior is if no creds are are passed, if the creds are wrong, if the creds are right.

pom.xml
milosimpson: Curious if the version Bump of Jetty was just a version bump for freshness sake, or did this bump unlock OpenID support.

Would be good to comment that in the commit message, so that if the version bump breaks other things downstream, they can know why it was bumped.

rnpridgeon: It's the first version that supports OIDC. Will add it to the commit message good idea

examples/src/main/java/io/confluent/rest/examples/hellooidc/HelloOpenID.java
rnpridgeon: Yea I wasn't super sure how to write a test for this. I'll see if there is a mock OIDC provider in the jetty code base somewhere. If memory serves they use google as a provider in their tests which isn't super awesome. I'll try to think of something 

1. OIDC (the protocol) isn't vendor specific, but their admin apis are. We could link their docs I suppose. In the long run(not the initial PR) I'd like to support dynamic client registration in rest utils but thats for another day. 

2. Ack will add some comments around this.

rnpridgeon: found a mock OpenIdProvider I think 👍

jeffhuang26:
Hi @rnpridgeon , Is there one page for this PR?

PR #171, opened 2020-02-10
Upgrade Jetty and Jersey to latest versions (fixes #170)

### LINKED ISSUE #170: Java 11 dependency resolution failure on artifact "javafx.base"
Hi. I recently upgraded to Confluent Platform 5.4.0 and pull in `kafka-avro-serializer` (which has a dependency on `rest-utils`), and I noticed that our Java 11 builds were failing with a dependency resolution failure:

```
[error] (tests / update) sbt.librarymanagement.ResolveException: Error downloading org.openjfx:javafx.base:11.0.0-SNAPSHOT
```

We were able to track this down to a dependency `rest-utils` declares here

https://github.com/confluentinc/rest-utils/blob/f4b871c578777cacdb6487445a66a745f0077308/pom.xml#L53

The artifact `org.glassfish.jersey.ext:jersey-bean-validation:jar:2.28` brings in a version of `hibernate-validator` that references the non-published `javafx.base` artifact.

The issue is described in more detail in this upstream [hibernate ticket](https://hibernate.atlassian.net/browse/HV-1644), and why it's only applicable in Java 11 builds.  There's a [StackOverflow post](https://stackoverflow.com/questions/53835014/what-makes-maven-want-openjfx-with-zulu-java-11-and-not-zulu-java-8-for-a-spring/59872126#59872126) that discusses the problem as well.  I think upgrading your jersey version may address the issue, but in the meantime I've added an exclusion rule for `javafx.base`.

ewencp:


ewencp:
This will affect a few downstream repos, see https://github.com/confluentinc/schema-registry/pull/1323 and https://github.com/confluentinc/ksql/pull/4508, which should all be merged together since the SR change doesn't have a way to migrate cleanly and the KSQL change is required since that build fails on deprecation warnings.

ewencp:
bump

PR #174, opened 2020-02-28
Catch invalid format exception: #501

MarkC0x:
Simple PR to better handle the `InvalidFormatException` which gets thrown if you try to commit offsets with a String rather than a Long with Rest Proxy.  This is for https://github.com/confluentinc/kafka-rest/issues/501 and now returns the 400 error code and suppresses the long stack trace being output into the log.


stanislavkozlovski:
cc @rigelbm

PR #175, opened 2020-03-03
CPKAFKA-4948: Make ssl.client.auth for REST consistent with Kafka

rajinisivaram:
`ssl.client.auth` is a String in Kafka that accepts lower-case values `required/requested/none`. In REST utils, we have two configs, a boolean `ssl.client.auth=true/false` that is deprecated and another config `ssl.client.authentication` that takes upper-case values `REQUIRED/REQUESTED/NONE`. `ssl.client.authentication`  has higher precedence than `ssl.client.auth`.

To make REST utils consistent with Kafka without breaking existing configuration, this PR:
  1) Undeprecates `ssl.client.auth` and deprecates `ssl.client.authentication`
  2) Makes both configs case insenstive
  3) Makes `ssl.client.auth` a String config that accepts `true/false/required/requested/none`
  4) Retains the existing precedence where`ssl.client.authentication` overrides `ssl.client.auth`

This PR was originally written because Kafka brokers failed to start with `ssl.client.auth` configured. This turned out to be another bug which affects multiple configs and is being fixed separately in MDS. Submitting this PR for review anyway to decide whether we want to make this change for consistency.

### DIFF
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -18,6 +18,7 @@ package io.confluent.rest;
 
 import io.confluent.common.metrics.Metrics;
 
+import java.util.Locale;
 import org.apache.kafka.common.config.ConfigException;
 import org.eclipse.jetty.jmx.MBeanContainer;
 import org.eclipse.jetty.server.Handler;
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -210,25 +211,31 @@ private void configureClientAuth(SslContextFactory sslContextFactory, RestConfig
...
         );
-        clientAuthentication = config.getBoolean(RestConfig.SSL_CLIENT_AUTH_CONFIG)
-                ? RestConfig.SSL_CLIENT_AUTHENTICATION_REQUIRED
-                : RestConfig.SSL_CLIENT_AUTHENTICATION_NONE;
+        clientAuthentication = config.getString(RestConfig.SSL_CLIENT_AUTH_CONFIG);
       }
     }
 
-    switch (clientAuthentication) {
+    switch (clientAuthentication.toLowerCase(Locale.ROOT)) {
```
rigelbm: Given all acceptable values are in the ASCII range, do we really need to use locale?

rigelbm: Nevermind. This very correct. My mistake. :?

rajinisivaram:
@rigelbm @arvindth Thanks for the review, merging to 5.5.x.

rajinisivaram:
Reverting this because it breaks REST proxy's ConsumerManagerTest that uses Kafka 1.1.0 without the case insensitive config validator class. We can do this in 6.0 when we remove this dependency.

PR #180, opened 2020-05-19
KIP-606 MetricsContext

rnpridgeon:
built on top of  https://github.com/apache/kafka/pull/8691. Will attempt to be diligent about keeping this PR in sync with the progress of 8691. A preliminary review would be nice though.


rnpridgeon:
POM changes need to be backed out prior to merger.

core/src/main/java/io/confluent/rest/metrics/RestMetricsContext.java
```diff
@@ -0,0 +1,97 @@
...
+            config.getString(RestConfig.METRICS_JMX_PREFIX_CONFIG));
+
+    /* Never overwrite preexisting resource labels */
+    this.putResourceLabel(RESOURCE_LABEL_TYPE,
+            config.getString(RestConfig.METRICS_JMX_PREFIX_CONFIG));
+  }
+
+  public RestMetricsContext(Map<String, Object> config) {
+    this.metadata = new HashMap<>();
+    config.forEach((key, value) -> metadata.put(key, (String) value));
```
srpanwar-confluent, line 56: value.toString()? This will break for non-string types

core/src/main/java/io/confluent/rest/metrics/RestMetricsContext.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * Sets {@link MetricsContext} namespace label.
+   */
+  protected void putNamespaceLabel(String labelKey, String labelValue) {
+    /* Remove resource label if present */
+    this.metadata.put(labelKey.replace(RESOURCE_LABEL_PREFIX, ""),
```
srpanwar-confluent, line 64: Do we want to pass labelKey here? I mean we know namespace key has to be MetricsContext.NAMESPACE ('_namespace').

rnpridgeon: NamespaceKey is probably a bad name. Basically I meant any label bound to the namespace itself in case a downstream component wants to add some additional labels. Perhaps just "Label" would be better.

rnpridgeon: 🤦 good catch

core/src/main/java/io/confluent/rest/metrics/RestMetricsContext.java
rnpridgeon: 🤦 good catch

core/src/main/java/io/confluent/rest/metrics/RestMetricsContext.java
rnpridgeon: Went with putLabel, open to suggestions though

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -134,6 +138,14 @@ public final String getPath() {
     return path;
   }
 
+
+  /**
+   * Set MetricsContext labels.
+   */
+  protected void setupMetricsContext(T appConfig) {
```
srpanwar-confluent, line 145: Thinking out loud here - since this is protected, a developer can assume that calling this after initialization would update the MetricsContext in Metrics object but that is not the case. Maybe we can remove this altogether or make it private

srpanwar-confluent: Ah, I am wrong

rnpridgeon: We could defer Metrics initialization to `configureHandler` so it's more similar to the pattern used by `setupResources`, `configureSecurityHandler` etc. I'm not sure there is any good reason you start the metric reporters before you can handle any traffic. That assumption may be incorrect fo something like C3 though.

srpanwar-confluent: That would be ideal but we do need Metrics object before we start the Application in C3. Dependency is bit convoluted in C3. For now, let keep this the way it is until we can fully understand the impact on downstream projects.

core/pom.xml
```diff
@@ -80,6 +80,7 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
+            <version>2.6.0-SNAPSHOT</version>
```
rigelbm, line 83: Should not this come from `<dependencyManagement>` in parent pom?

core/src/main/java/io/confluent/rest/metrics/RestMetricsContext.java
```diff
@@ -0,0 +1,97 @@
...
+
+package io.confluent.rest.metrics;
+
+import io.confluent.rest.RestConfig;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.kafka.common.metrics.MetricsContext;
+
+public class RestMetricsContext<T> implements MetricsContext {
```
rigelbm, line 25: What is `T` used for?

core/src/main/java/io/confluent/rest/Application.java
rigelbm, line 145: Do we want to allow people to override this method?

core/pom.xml
rnpridgeon: this is to be removed, it's built on top of a PR not an actual branch. I wasn't thinking about how this only works for me because I populated my local maven cache with it.

core/src/main/java/io/confluent/rest/Application.java
rnpridgeon: Yes to allow application to add additional labels to the MetadataContext. I suppose we could also just append the labels we want to the Configdef but it felt akward since they aren't really meant to be configurable.

rnpridgeon: I'll add a docstring for now

rnpridgeon:
https://github.com/apache/kafka/pull/8691 has been merged and this PR updated with the changes. I also removed the resource label constraints from the base RestMetricsContext as it felt a bit weird to leak a specific metrics reporters details into rest-utils. We will leak them into the Application implementations instead :p

core/src/main/java/io/confluent/rest/Application.java
rnpridgeon: I just went ahead and moved everything into the config. It feels a bit weird as MetricContext labels don't really feel configurable like MetricLabels but they are propagated as configs to downstream components so it is what it is.

core/pom.xml
rnpridgeon: removed, note build will not succeed until AK changes are propagated to ccs

core/src/main/java/io/confluent/rest/Application.java
rnpridgeon: removed

core/src/main/java/io/confluent/rest/metrics/RestMetricsContext.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package io.confluent.rest.metrics;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.kafka.common.metrics.MetricsContext;
+
+public class RestMetricsContext implements MetricsContext {
```
rigelbm, line 24: It's not clear to me how this class is allowed to be subclassed. It's usually an anti-pattern to have a non-final non-abstract class. What are subclasses allowed to override? Everything? This class keeps state around (`contextLabels`). Are there invariants that subclasses need to maintain on this state? Is this class supposed to be thread-safe? Are implementations also supposed to be? If we are allowing subclassing just for code reuse, there's almost no code in here. I'm not sure it's worth it keeping this class open (and having to think about all the above), instead of just closing it and requiring people to just reimplement it if they need different behaviour.

core/src/test/java/io/confluent/rest/metrics/TestRestMetricsContext.java
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+package io.confluent.rest.metrics;
+
+import org.apache.kafka.common.metrics.MetricsContext;
+
+import java.util.Map;
+
+public class TestRestMetricsContext extends RestMetricsContext {
```
rigelbm, line 23: This is an example where delegation would actually be much better than inheritance.

core/src/main/java/io/confluent/rest/metrics/RestMetricsContext.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.kafka.common.metrics.MetricsContext;
+
+public final class RestMetricsContext implements MetricsContext {
+  /**
+   * Client or Service's metadata map.
+   */
+  protected final Map<String, String> contextLabels;
```
rigelbm, line 28: Here and everywhere else: Can be `private`.

core/src/test/java/io/confluent/rest/metrics/TestRestMetricsContext.java
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+
+package io.confluent.rest.metrics;
+
+import java.util.Map;
+import org.apache.kafka.common.metrics.MetricsContext;
+
+
+public class TestRestMetricsContext {
```
rigelbm, line 23: final

PR #181, opened 2020-05-19
Fix pom to inherit from common-parent, instead of common.

rigelbm:
By inheriting from common, it adds a dependency (common-utils) to all the poms that inherit from this one, which is not the intended behaviour. Child poms that rely on this dependency should depend on common-utils directly.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -81,6 +81,10 @@
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.confluent</groupId>
+            <artifactId>common-utils</artifactId>
+        </dependency>
         <!--test-->
         <dependency>
             <groupId>junit</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
 
     <parent>
         <groupId>io.confluent</groupId>
-        <artifactId>common</artifactId>
+        <artifactId>common-parent</artifactId>
         <version>6.0.0-SNAPSHOT</version>
     </parent>
 


rigelbm:
> Can you run a test build against muckrake? There's a chance downstream repos depend on the dependency from common.

Is not the downstream-validation stage of the Jenkins PR builder enough? I fixed the only repository that failed already.

ewencp:
Something seems off with the PR link, for some reason it's pointing to an internal build instead of the community PR builder, so I can't tell what happened with this build.

However, for downstream builds we're trying to allow for a balance between overhead/cost and risk. By default, which is what this repo uses, we only build repos that declare an explicit dependency on a rest-utils module (see https://github.com/confluentinc/jenkins-common/blob/master/vars/downStreamValidation.groovy#L7 but the full details are also tied into the details of how the dependency resolution works). Because of this, a muckrake build will still (potentially) be more complete since it will also build things with transitive dependencies. That test has always been a bit of a hack that just happens to encompass pretty much everything we care about (especially when there was only one product sku). This could be addressed by enforcing explicit dependencies for imports, but since we're not doing that today, muckrake build gives better coverage than just the downstream check.

We could make this more flexible to support more extensive tests in low traffic repos with larger potential blast area, e.g. by changing https://github.com/confluentinc/jenkins-common/blob/master/vars/common.groovy#L83 to support some flag from the build config, but that isn't there today.

either muckrake directly or ducker would be fine, normally it only takes ~30min to build serially, but it's just a nice sanity check so we don't have to scramble post-merge to unblock another team.

ewencp:
@rigelbm Is this still relevant? If so, can we rebase, re-run jenkins, and get validation of downstreams?

PR #182, opened 2020-05-22
Add getter for Application's Metrics object

srpanwar-confluent:
**Why**
Allows the currently configured metrics object to be reused


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -134,6 +134,13 @@ public final String getPath() {
     return path;
   }
 
+  /**
+   * Returns {@link Metrics} object
+   */
+  public Metrics getMetrics() {
```
rigelbm, line 140: Make it `final`. :)

PR #186, opened 2020-06-30
Extensible Application Metrics instance configuration

rnpridgeon:
We leak some metrics reporter configs into kafka-rest. This is intended facilitate breaking some of that coupling.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -19,7 +19,6 @@ package io.confluent.rest;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.jaxrs.base.JsonParseExceptionMapper;
 
-import io.confluent.rest.metrics.RestMetricsContext;
 import org.apache.kafka.common.config.ConfigException;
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -137,6 +125,37 @@ public final String getPath() {
     return path;
   }
 
+  /**
+   * Configure Application MetricReport instances.
+   */
+  private List<MetricsReporter> configureMetricsReporters(T appConfig) {
```
rigelbm, line 131: Not specific to this method, as I see many methods follow this pattern, but why do you need to receive `appConfig` here? Why can't you just use `config`? And why does it need to be a `T`, instead of a `RestConfig`?

rnpridgeon:
Jenkins exploded and I needed to retargeting to 6.0.x which is the intended target anyway.

core/src/main/java/io/confluent/rest/Application.java
rnpridgeon: Mostly following convention here. Since there isn't really anything application specific in this method I opted to pass the config in and mark it private rather than accessing Application.config. Leaving as is for now.

PR #187, opened 2020-07-01
Lazily load Application Metrics instance

rnpridgeon:
Also provides common prefix for extracting MetricsReporter configs


rnpridgeon:
in practice I feel its better to rely on getConfiguration over referencing  the Application.config reference directly as derived classes will not need to import the Metrics dependency directly. They can opt to override getConfiguration instead to influence what tags and/or  reporters they want to use.

PR #189, opened 2020-07-08
Removed OPTION omission for CORS

maverick64:
This had to be done for https://confluentinc.atlassian.net/browse/ESCALATION-3168.
When Customer has RBAC/CORS enabled together, anonymous requests run into issue because of jetty/OAuthBearerAuthenticator. 
We came to conclusion to get it fixed in rest-utils. 
This change will require all HTTP methods (including OPTION) to require authentication if it was enabled.


maverick64:
> this will likely break browser CORS preflight requests, which require OPTIONS methods to not authenticate.

Added this flag through which we can disable preflight option check for the servers. The components can just add this flag to true and it will behave as before.
access.control.skip.options

PR #192, opened 2020-07-10
All original configs must be passed to MetricReporters

rnpridgeon:


srpanwar-confluent:
You may want to target this fix to 6.0.x branch.

PR #193, opened 2020-07-10
Removed OPTION omission for CORS

maverick64:
This had to be done for https://confluentinc.atlassian.net/browse/ESCALATION-3168.
When Customer has RBAC/CORS enabled together, anonymous requests run into issue because of jetty/OAuthBearerAuthenticator.
We came to conclusion to get it fixed in rest-utils.
This change will require all HTTP methods (including OPTION) to require authentication if it was enabled.

### DIFF
--- a/core/src/main/java/io/confluent/rest/auth/AuthUtil.java
+++ b/core/src/main/java/io/confluent/rest/auth/AuthUtil.java
@@ -128,9 +128,6 @@ public final class AuthUtil {
     final ConstraintMapping mapping = new ConstraintMapping();
     mapping.setConstraint(constraint);
     mapping.setMethod("*");
-    if (authenticate && AuthUtil.isCorsEnabled(restConfig)) {
-      mapping.setMethodOmissions(new String[]{"OPTIONS"});
-    }
     mapping.setPathSpec(pathSpec);
     return mapping;
   }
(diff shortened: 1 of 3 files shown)


brianbushree:
Did we ever get to the bottom of why this extra `OPTIONS` check was here to begin with? Presumably it was added for a good reason...

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
```diff
@@ -128,9 +128,6 @@ private static ConstraintMapping createConstraint(
     final ConstraintMapping mapping = new ConstraintMapping();
     mapping.setConstraint(constraint);
     mapping.setMethod("*");
-    if (authenticate && AuthUtil.isCorsEnabled(restConfig)) {
```
norwood, line 131: @maverick64 did you verify this works with basic auth?

that i can make an OPTIONS call from the browser and the 401 does not prevent the return, or at least prompts the user for a login

maverick64:
> Did we ever get to the bottom of why this extra `OPTIONS` check was here to begin with? Presumably it was added for a good reason...

Yes, we did. The OPTION check results in jetty to set "mandatory" flag for a security context for anonymous request.  We either need a WA in OAuthOrBasicAuthenticator::validateRequest or in this rest-util call. Rest-util change will just require OPTION to be authenticated which is a preferred solution as per @jeffhuang26

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
maverick64: I didn't even know this code path gets used by C3. Browser could be sending unauthenticated OPTION request. @jeffhuang26 can you comment on this as well?

maverick64:
Here is the slack communication for the history, https://confluent.slack.com/archives/CM6PEBEGM/p1593558160076500

PR #204, opened 2020-08-22
ST-3461: Nano versioning

elismaga:
Updated the regex used to get the rpm version so it supports the format of nano versioned releases.

These changes will be merged during phase two of the nano versioning roll out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ test:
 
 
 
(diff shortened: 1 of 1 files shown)


gracechensd (commented):
should project.version in pom.xml be changed as well? (https://github.com/confluentinc/rest-utils/blob/master/pom.xml#L71-L78)

elismaga:
> should parent [version](https://github.com/confluentinc/rest-utils/blob/master/pom.xml#L11) and [project.version](https://github.com/confluentinc/rest-utils/blob/master/pom.xml#L71-L78) in pom.xml be changed as well, or not needed?

@gracechensd This PR is for the rpm branch, and those files are on the master branch. There is another PR for the rest-utils master branch that covers those.

gracechensd (approved):

brianbushree (approved):

PR #206, opened 2020-10-13
SEC-1473 Allow no Truststore password

WanderingStar:
If the Keystore is loaded without a password (with the password argument `null`), the integrity check is skipped. (see: https://docs.oracle.com/javase/8/docs/api/java/security/KeyStore.html#load-java.io.InputStream-char:A-). Kafka handles this by giving the corresponding parameter a `null` default: https://github.com/confluentinc/ce-kafka/blob/master/clients/src/main/java/org/apache/kafka/common/config/SslConfigs.java#L163 But rest-utils was using `""` for the default, which doesn't skip the integrity check. 
 
This also fixes what seems like a long-existing error in the tests where the wrong number of arguments was supplied (see first commit).


core/src/test/java/io/confluent/rest/ApiHeadersTest.java
```diff
@@ -130,7 +130,7 @@ private static void createKeystoreWithCert(File file, String alias, Map<String,
     final X509Certificate cert = new CertificateBuilder(30, "SHA1withRSA")
         .sanDnsNames("localhost").generate("CN=mymachine.local, O=A client", keypair);
 
-    TestSslUtils.createKeyStore(file.getPath(), new Password(SSL_PASSWORD), alias,
+    TestSslUtils.createKeyStore(file.getPath(), new Password(SSL_PASSWORD), new Password(SSL_PASSWORD), alias,
```
WanderingStar, line 133: This confuses me. It seems like `createKeyStore`'s signature has been stable since 2015, but there have been changes to these tests more recently than that. How could the tests have passed?

milosimpson:
Given that this issue was identified in an MDS deployment, there should be a test in MDS repo for `confluent.metadata.server.ssl.truststore` with no key yes?

WanderingStar:
We should verify that this fix actually fixes the observed problem, but is it reasonable to test your dependencies on an ongoing basis?

milosimpson:
> We should verify that this fix actually fixes the observed problem, but is it reasonable to test your dependencies on an ongoing basis?

Yeah, I was just pointing out there "royal we" should also verify/add a test for this in MDS repo before ticket is closed?

core/src/test/java/io/confluent/rest/ApiHeadersTest.java
arvindth: FYI, although this method has been stable, the previously called method now no longer exists, hence the compile failure. This just bit MDS as well.

PR #210, opened 2020-12-02
ST-4542: Upgrade jetty to address CVE-2020-27216

niteshmor:
Note that Jersey needs to be upgraded as well to ensure compatibility. 
See this word of caution that was included in AK before this upgrade was applied https://github.com/apache/kafka/blob/dbdd1b1bb72ab0e012f7a746cefb6e593da44683/gradle/dependencies.gradle#L71

Note that the CI build fails because of a few downstream repositories using Jersey's private Base64 for some tests. Those fixes will come later.


niteshmor:
Note the related, but subtly different, pull requests #211, #212, #213, #214 against more recent branches.

xli1996 (approved):
lgtm with a green build.

niteshmor:
> lgtm with a green build.

Thanks. There are two downstream repositories that have a failed build. PRs are already out to fix those downstreams. I will merge those first, and then trigger another build here to vet the change.

PR #215, opened 2020-12-04
ff-3928: bump version of Guava in parent of schema-registry.

xjin-Confluent:
https://github.com/confluentinc/schema-registry takes Guava version from this parent. 
There is a CVE related to Guava: https://confluentinc.atlassian.net/jira/software/c/projects/FF/issues/FF-3928

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
         <jersey.version>2.31</jersey.version>
         <jetty.version>9.4.33.v20201020</jetty.version>
         <asynchttpclient.version>2.2.0</asynchttpclient.version>
-        <guava.version>24.0-jre</guava.version>
+        <guava.version>24.1.1-jre</guava.version>
         <checkstyle.suppressions.location>checkstyle/suppressions.xml</checkstyle.suppressions.location>
     </properties>
 


xjin-Confluent:
[clabot:check]

ewencp (commented):
Seems fine if we pass downstream builds and tests -- this is the type of change that's likely worth running some system tests against.

Is this intentionally only going back to 5.5.x?

xjin-Confluent:
> Seems fine if we pass downstream builds and tests -- this is the type of change that's likely worth running some system tests against.
> 
> Is this intentionally only going back to 5.5.x?

Yes. This is the version of release customers look for.  I am relatively new here ( almost 2 months). Could you please let me know how I can verify this before checking in the change? Thanks.

ewencp:
You get some validation via the "downstream validation" check in Jenkins (if you look at the Pipeline Steps link on the Jenkins check you should see that parent step and a bunch of steps under it that build individual projects).

For dependency changes that can affect a lot of repos, I also like to do a system test run. You can modify muckrake's default-version.sh to point to your fork + branch for this change, then do a test run via Jenkins that builds the platform and runs some end to end tests to get better confidence that the change doesn't break anything accidentally.

@niteshmor might have more thoughts on testing since he's been doing a bunch of CVE updates recently. Ideally we would get a mostly standardized process documented, though it will differ a bit for common shared libraries like this (used across probably a dozen repos) vs if a CVE is only in a particular product (e.g. maybe only one connector uses a specific library that has a CVE).

ewencp:
Forgot to mention too, I asked about 5.5.x because if there's something worth fixing in 5.5.x, it probably means we should be fixing it back to our oldest supported release as well unless there's some reason it can't be. You can see timeframes for support here https://docs.confluent.io/platform/current/installation/versions-interoperability.html, so 4.1.x would be our current oldest supported release. When you merge into an older branch, we have merge checks (and a tool `pint merge`) that ensures the change makes it onto all newer branches up to master.

xjin-Confluent:
> Forgot to mention too, I asked about 5.5.x because if there's something worth fixing in 5.5.x, it probably means we should be fixing it back to our oldest supported release as well unless there's some reason it can't be. You can see timeframes for support here https://docs.confluent.io/platform/current/installation/versions-interoperability.html, so 4.1.x would be our current oldest supported release. When you merge into an older branch, we have merge checks (and a tool `pint merge`) that ensures the change makes it onto all newer branches up to master.

Yes. It definitely worth fixing for all repos for this specific issue: the outdated Guava that contains a medium CVE. We intentionally do 5.5.x now because it is currently blocking a deal and we are to release 5.5.3 soon.  There is another effort to sync the version of Guava company wide which I think takes more time and changes.

xjin-Confluent:
It seems rest-utils is the parent of some repos and so I would like to do this for 5.5.x only this time since we have 2 weeks to figure out the release and unblock a customer deal.  I worried that if we go all the way back to 5.1.x, this has a wide impact on all dependencies and we might not have time to address all.  WDYT?

xjin-Confluent:
This is the system test https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/4780

xjin-Confluent:
@ ewencp I see seveal failures in the above tests. Searching for ": failure", I found 7 and they are mostly related to "Master election failed" and "IO".  My guess is it is due to "zookeper can not be started".  There are also "mkdir: cannot create directory ‘/mnt/kafka’: No space left on device".   This also happened to another test.

xjin-Confluent:
Here is the report: http://confluent-platform-5.5.x-system-test-results.s3-us-west-2.amazonaws.com/2020-12-08--002.1607528856--xjin-Confluent--HEAD--47bcd856/report.html

xjin-Confluent:
Ran several system tests, but ends up with errors likely not-related with Guava. Examples are "disk space full" and "zookeeper can not start". Tried different tests with only control center tests with more nodes, but failed with different reasons. See my worklog here: https://confluentinc.atlassian.net/wiki/spaces/~913794610/pages/1809418291/Javassist+and+Guava+Workpage#Rest-util

Given that we want to verify that this is effective in fixing Guava in 5.5.x for customer deal blocker, I will push the change now and monitor for system tests. This will only be pushed to 5.5.x branch. Also, this gives us chances to verify that this is doable for 5.5.x. Worst case, we revert the changes and this proves there is no fixes for 5.5.x in terms of Guava. @ ewencp : please DM me if you find anything suspicious.

PR #216, opened 2020-12-17
MMA-7520: add CsrfTokenProtectionFilter

srpanwar-confluent:
Jira: https://confluentinc.atlassian.net/browse/FF-3535

**What**
Simple server-side request filter that implements CSRF protection as per the Guidelines for implementation of REST by NSA (section IV.F), section 4.3 of this paper[http://seclab.stanford.edu/websec/csrf/csrf.pdf] and
 https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#synchronizer-token-pattern.
 
As part of this PR, few new flags have been added
- To enable CSRF protection, the following property must be set to true. 
     Default: `false`
    **csrf.prevention.enable=**
    
- The endpoint that client would call to get the CSRF token
     Default: `/csrf`
    **csrf.prevention.token.endpoint=**
    
- The token expiration period (in minutes)
     Default: `30`
    **csrf.prevention.token.expiration.minutes=**
    
- The maimum number of session entries that the server should cache
     Default: `10000`
    **csrf.prevention.token.max.entries=**

**Request flow** 
1. Client calls `/csrf` endpoint with  a `session-id`. Session-Id is set in the header - `X-Requested-By`
3. Server responds with the new token value. The token is set in the header - `X-CONFLUENT-CSRF-TOKEN`. Subsequent calls from the client should include this value in the header `X-Requested-With`.
4. Per the specification, `GET, OPTIONS, HEAD` requests are no guarded by CSRF token
5. For other requests, the client has to pass both `X-Requested-By` & `X-Requested-With`. If the values stored in the server do match the values sent by the client, the request is aborted.


core/src/main/java/io/confluent/rest/filters/CsrfTokenProtectionFilter.java
```diff
@@ -0,0 +1,182 @@
...
+    if (!METHODS_TO_IGNORE.contains(httpServletRequest.getMethod())) {
+      // For guarded methods 'requestedBy' is required
+      if (Strings.isNullOrEmpty(requestedBy)) {
+        log.error("(Cross site request forgery): Missing user session identifier");
+        throw new BadRequestException(INVALID_REQUESTER_MESSAGE);
+      }
+
+      // Check if the token in request header indeed belongs to requested user
+      if (!tokenSupplier
+          .getUnchecked(requestedBy)
```
cryoshida, line 138: should we change here to getIfPresent() s.t. if it's an invalid token, we don't load it in the cache?

cryoshida (approved):

core/src/main/java/io/confluent/rest/filters/CsrfTokenProtectionFilter.java
srpanwar-confluent: Great catch

PR #217, opened 2020-12-18
MMA-9580: Fix test failure

srpanwar-confluent:
**What**
1. Don't rely on Thread.sleep. It can be flaky.
2. Shutdown the app even if the test has failed.
3. Minor refactoring


milosimpson:
@srpanwar-confluent 
This still a thing?

yanglei99 (approved):
lgtm

PR #218, opened 2021-01-21
DG:1183 Address http status 422 which is not supported in Response.status in  rest-util's ExceptionMapper

maverick64:
Jira ticket https://confluentinc.atlassian.net/browse/DG-1183

### DIFF
--- a/core/src/main/java/io/confluent/rest/exceptions/DebuggableExceptionMapper.java
+++ b/core/src/main/java/io/confluent/rest/exceptions/DebuggableExceptionMapper.java
@@ -51,6 +51,13 @@ public abstract class DebuggableExceptionMapper<E extends Throwable> implements 
     private final Response.Status.Family family;
 
     public HttpStatus(int statusCode, String reasonPhrase) {
+
+      // Sanity check to make sure status code is within valid range
+      if (statusCode < 0 || statusCode > 599) {
+        throw new IllegalArgumentException("Http status code is not within valid range:"
+                + "statusCode: " + Integer.toString(statusCode));
+      }
+
       this.code = statusCode;
       this.reason = reasonPhrase;
       this.family = Response.Status.Family.familyOf(statusCode);
--- a/core/src/main/java/io/confluent/rest/exceptions/DebuggableExceptionMapper.java
+++ b/core/src/main/java/io/confluent/rest/exceptions/DebuggableExceptionMapper.java
@@ -58,7 +58,7 @@ public abstract class DebuggableExceptionMapper<E extends Throwable> implements 
    * @param status HTTP response status
    */
   public Response.ResponseBuilder createResponse(Throwable exc, int errorCode,
-                                                 Response.Status status, String msg) {
+                                                 Response.StatusType status, String msg) {
     log.error("Request Failed with exception " ,  exc);
     String readableMessage = msg;
     if (restConfig != null && restConfig.getBoolean(RestConfig.DEBUG_CONFIG)) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/confluent/rest/exceptions/WebApplicationExceptionMapper.java
```diff
@@ -32,6 +32,45 @@
   @Context
   HttpHeaders headers;
 
+  public static enum Status implements Response.StatusType {
```
rayokota, line 35: In order to account for other status codes besides 422, maybe just make this a class that can hold an arbitrary `statusCode`?

rayokota (commented):
Thanks @maverick64 , left a comment

core/src/main/java/io/confluent/rest/exceptions/DebuggableExceptionMapper.java
```diff
@@ -58,7 +99,7 @@ public DebuggableExceptionMapper(RestConfig restConfig) {
    * @param status HTTP response status
    */
   public Response.ResponseBuilder createResponse(Throwable exc, int errorCode,
-                                                 Response.StatusType status, String msg) {
+                                                 HttpStatus status, String msg) {
```
rayokota, line 102: This can stay as `Response.StatusType` right?

core/src/main/java/io/confluent/rest/exceptions/GenericExceptionMapper.java
```diff
@@ -40,7 +40,7 @@ public Response toResponse(Throwable exc) {
     // There's no more specific information about the exception that can be passed back to the user,
     // so we can only use the generic message. Debug mode will append the exception info.
     return createResponse(exc, Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(),
-                          Response.Status.INTERNAL_SERVER_ERROR,
+                          new HttpStatus(Response.Status.INTERNAL_SERVER_ERROR),
```
rayokota, line 43: Then you don't need to wrap Response.Status here

core/src/main/java/io/confluent/rest/exceptions/WebApplicationExceptionMapper.java
```diff
@@ -83,19 +45,22 @@ public Response toResponse(WebApplicationException exc) {
...
     // status to be null above for the cases of some valid 4xx status i.e 422.
-    // Below we are trying to replace the null with a valid status listed in
-    // WebApplicationExceptionMapper.Status
+    // Below we are trying to replace the null with the status passed in the exception
 
-    if (status == null) {
-      status = Status.fromStatusCode(exc.getResponse().getStatus());
+    HttpStatus internalStatus;
+    if (status != null) {
+      internalStatus = new HttpStatus(status);
```
rayokota, line 52: And you don't need the wrapper here.

rayokota (commented):
Left some more comments

core/src/main/java/io/confluent/rest/exceptions/WebApplicationExceptionMapper.java
maverick64: I was not inclined to change the behavior  for other non-supported status codes, but as you think it can be changed for all, I have updated it.

core/src/main/java/io/confluent/rest/exceptions/WebApplicationExceptionMapper.java
```diff
@@ -32,6 +32,7 @@
   @Context
   HttpHeaders headers;
 
+
```
rayokota, line 35: nit: extra line

rayokota (approved):
LGTM, just left a minor comment

norwood (approved):

PR #219, opened 2021-02-02
FF-4241: Upgrade jetty version for 5.3.x and earlier

xjin-Confluent:
https://confluentinc.atlassian.net/browse/FF-4241

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,8 +50,8 @@
         <activation.version>1.1.1</activation.version>
         <apache.httpclient.version>4.5.13</apache.httpclient.version>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
-        <jersey.version>2.27</jersey.version>
-        <jetty.version>9.4.11.v20180605</jetty.version>
+        <jersey.version>2.31</jersey.version>
+        <jetty.version>9.4.35.v20201120</jetty.version>
     </properties>
 
     <repositories>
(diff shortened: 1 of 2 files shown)


omkreddy:
Note that we may need to fix compilation/test failures due to the change. I think we need to run platform system tests with this change + kafka changes.

pom.xml
```diff
@@ -55,7 +55,7 @@
         <!-- can't upgrade jersey past 2.25 until we move to java 8 -->
         <jersey.version>2.25</jersey.version>
```
omkreddy, line 56: maybe we can also update jersey to 2.31

pom.xml
```diff
@@ -55,7 +55,7 @@
         <!-- can't upgrade jersey past 2.25 until we move to java 8 -->
```
omkreddy, line 55: its better to merge this PR from 5.0 branch on-wards.

omkreddy:
its better to merge this PR from 5.0 branch on-wards.

pom.xml
xjin-Confluent: noticed here: https://confluentinc.atlassian.net/browse/CPKAFKA-6127. Will add that change.

xjin-Confluent:
> its better to merge this PR from 5.0 branch on-wards.

 Yep, it's better skipping 4.0.x. I will change my base branch.

omkreddy:
looks like SSL tests are failing. maybe we need to backport some changes from 5.4 branches.

this may be risky update. maybe we should consider updating jetty from 5.4.x and above.

omkreddy:
Also looks like we have dependency on jersey 2.27 and below for some modules like ksql/control center. So we can not update jersey without code changes to those modules :(.   I think, we can not update jersey for 5.3.x and earlier.

PR #223, opened 2021-02-09
UIF-302: Upgrade guava for 5.4.x and older.

xjin-Confluent:
We did it for 5.5.x and later: https://github.com/confluentinc/rest-utils/pull/215/files, this PR wants to fix back to 5.1.x when Guava first appears.


ewencp (approved):
lgtm

PR #227, opened 2021-02-18
MMA-9930: Use Guava version defined in Common.

xjin-Confluent:
See the last comment in https://confluentinc.atlassian.net/browse/MMA-9930.
Common has been updated with Guava defined in all branches: https://github.com/confluentinc/common/pull/314
Removing Rest-Utils Guava property will inherit definition from Common.
Also Rest-utils 6.0.x and onwards do not have Guava definition.


xjin-Confluent:
This is too big a change for downstream consumers. Since we have rest-utils 5.1.x - 5.5.x using 24.1.1-jre version, it is less risker to add this to 5.0.x compared to upgrading everyone to 28.1 as defined in Common.

PR #228, opened 2021-02-19
MMA-9930: Use compatible version of Guava

xjin-Confluent:
5.1.x ---- 5.5.x all use 24.1.1-jre version of Guava. Although Common uses 28.1, for the purpose of fixing Guava CVE, we do want to enforce all consumers of Rest-Utils to use 24.1.1-jre for now. It can be future work to sync Guava version.

This will only be in 5.0.x.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,7 @@
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
         <jersey.version>2.27</jersey.version>
         <jetty.version>9.4.11.v20180605</jetty.version>
+        <guava.version>24.1.1-jre</guava.version>
     </properties>
 
     <repositories>


srpanwar-confluent:
>5.1.x ---- 5.5.x all use 24.1.1-jre version of Guava. Although Common uses 28.1, for the purpose of fixing Guava CVE, we do want to enforce all consumers of Rest-Utils to use 24.1.1-jre for now. It can be future work to sync Guava version.
This will only be in 5.0.x.

@xjin-Confluent from the description it sounds like common is already using a higher version of the lib?

xjin-Confluent:
> > 5.1.x ---- 5.5.x all use 24.1.1-jre version of Guava. Although Common uses 28.1, for the purpose of fixing Guava CVE, we do want to enforce all consumers of Rest-Utils to use 24.1.1-jre for now. It can be future work to sync Guava version.
> > This will only be in 5.0.x.
> 
> @xjin-Confluent from the description it sounds like common is already using a higher version of the lib?

Yes. Common 5.0.x--5.5.x uses 28.1 Guava. However, Blueway's parent is confluent-security-plugin (5.4.x---master) and Rest-Utils (5.3.x and older). So the effective Guava version for Blueway's 4 and 5 releases is 24.1.1-jre or 20.0 (defined in Blueway).  I tried to bump it to 28.1 using another PR (by removing Guava from Rest-Utils, and thus Blueway will pick up Guava 28.1) but there are build errors. Despite that we can fix it, I think it is not safe to do it in a security fix. So this fix is much smaller: it fixes 5.0.x since 5.1.x and later are fixed. 

We have another PR to remove the explicit Guava definition in Blueway so that it picks up versions on Rest-Utils. https://github.com/confluentinc/blueway/pull/2204

srpanwar-confluent:
> > > 5.1.x ---- 5.5.x all use 24.1.1-jre version of Guava. Although Common uses 28.1, for the purpose of fixing Guava CVE, we do want to enforce all consumers of Rest-Utils to use 24.1.1-jre for now. It can be future work to sync Guava version.
> > > This will only be in 5.0.x.
> > 
> > 
> > @xjin-Confluent from the description it sounds like common is already using a higher version of the lib?
> 
> Yes. Common 5.0.x--5.5.x uses 28.1 Guava. However, Blueway's parent is confluent-security-plugin (5.4.x---master) and Rest-Utils (5.3.x and older). So the effective Guava version for Blueway's 4 and 5 releases is 24.1.1-jre or 20.0 (defined in Blueway). I tried to bump it to 28.1 using another PR (by removing Guava from Rest-Utils, and thus Blueway will pick up Guava 28.1) but there are build errors. Despite that we can fix it, I think it is not safe to do it in a security fix. So this fix is much smaller: it fixes 5.0.x since 5.1.x and later are fixed.
> 
> We have another PR to remove the explicit Guava definition in Blueway so that it picks up versions on Rest-Utils. [confluentinc/blueway#2204](https://github.com/confluentinc/blueway/pull/2204)

@xjin-Confluent thanks for the explanation. Appreciate it.

srpanwar-confluent (approved):

PR #230, opened 2021-02-23
MMA-10032: Make configure handler methods public

srpanwar-confluent:
Jira: https://confluentinc.atlassian.net/browse/MMA-10032

**What**
Change the visibility of `configure handler` methods

**Why**
To allow hosting app to dynamically add/remove handlers at runtime from `Application` object


thomaskwscott (approved):
LGTM

PR #232, opened 2021-03-01
CIAM-526 Count error codes after ExceptionMappers and use actual status code for all exception classes

cyrusv:
Why/How
-----
MDS in cloud is occasionally registering unknown exceptions. These were marked as "status unknown", and upon reading the kibana logs, they looked like 400s, not 500s. So I looked closer at how we emit this metric, and it looks like this accounting happens before any mappers are applied. Since a mapper could take any exception and set any status code, it's clear that this error accounting should happen after mappers (in the "Finished" event). Added bonus is now we can directly ask for the actual status code returned, rather than the weird type-casting.

Also added error logs in the event that we can't tell how to tag this exception, so we can study those cases in detail more easily.

### DIFF
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -41,8 +41,6 @@ import java.util.concurrent.TimeUnit;
 
 import io.confluent.rest.annotations.PerformanceMetric;
 
-import javax.ws.rs.WebApplicationException;
-
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.metrics.Metrics;
 import org.apache.kafka.common.metrics.stats.Avg;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -278,11 +280,14 @@ public void finished(long requestSize, long responseSize, long latencyMs) {
      * Indicate that a request has failed with an exception.
      */
     public void exception(final RequestEvent event) {
-      final int statusCode = event.getException() instanceof WebApplicationException
-          ? ((WebApplicationException) event.getException()).getResponse().getStatus() : 0;
-      int idx = statusCode / 100;
+      int idx = event.getContainerResponse() != null
```
milosimpson, line 283: // map the http status codes down to their classes (2xx, 4xx, 5xx)
// use the containerResponse status as it has the http status after ExceptionMappers are applied

milosimpson (approved):
LGTM 
Would like the comment added/updated

Also saw that you wrote this code in 2019 ;)

rnpridgeon:
LGTM thanks!

rnpridgeon (approved):

PR #237, opened 2021-04-02
Define Netty version to be shared to downstreams.

xjin-Confluent:


andrewegel (approved):
Approved upon PR build pass.

PR #239, opened 2021-04-08
KREST-860 Fix deprecated usages of org.apache.kafka.common.Metric.value()

dimitarndimitrov:
This addresses the deprecated API usages in `6.0.x` that will break after the first wave of Kafka 3.0.0 changes are merged in (the relevant discussion is available in Confluent's Slack).
- AFAICS, there's another set of usages that live just in `master`, related to the `5xx` metrics - I'll address these with a separate PR.

Tested by running all `rest-utils` tests locally.


rigelbm (approved):
LGTM

PR #244, opened 2021-04-21
Upgrade Jetty to 9.4.39.v20210325

xjin-Confluent:
Catch up with 5.5.x on Jetty version https://github.com/confluentinc/rest-utils/commit/2af4448a290fdca8ffc159796260c5ea3a9adc78


niteshmor (approved):
Tests pass, so looks good. You'll need to do a similar update in kafka as well.

PR #247, opened 2021-04-30
ST-5293: Cleanup of netty version property

elismaga:
This netty version property is now defined in common so no need to have it defined here as well.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,6 @@
         <asynchttpclient.version>2.2.0</asynchttpclient.version>
         <checkstyle.suppressions.location>checkstyle/suppressions.xml</checkstyle.suppressions.location>
         <io.confluent.rest-utils.version>6.2.0-0</io.confluent.rest-utils.version>
-        <netty.version>4.1.63.Final</netty.version>
     </properties>
 
     <repositories>


andrewegel (approved):
We should figure out the build issues and get the team to address them - FWIW I believe these tests are rather flaky.

ewencp:
Doesn't look like flakiness:

> java.lang.NoSuchMethodError: io.netty.buffer.ByteBuf.isContiguous()Z

elismaga:
> Doesn't look like flakiness:
> 
> > java.lang.NoSuchMethodError: io.netty.buffer.ByteBuf.isContiguous()Z

@ewencp i saw the same test failures on the 6.2.x CI job so I didn't think they were related to my change here. Also, I didn't see anywhere in rest-utils where this property was even being used. I merged my change before seeing your comment though. I will look further into this.

elismaga:
@ewencp I think this may be caused by a previous change I made to common where I set the netty version and changed the dependency there. The first build of rest-utils that started failing with these errors was triggered by the build of common with my change. I'm not sure what to do here. Probably will need some help from the rest utils folks because I didn't even see where netty was being used here.
https://jenkins.confluent.io/job/confluentinc/job/common/job/6.2.x/148/
https://jenkins.confluent.io/job/confluentinc/job/rest-utils/job/6.2.x/174/

PR #252, opened 2021-05-16
KREST-1030 Upgrade Jetty and Jersey to address a number of CVEs

dimitarndimitrov:
These two changes address multiple CVEs related to Jetty and Jersey (including an exclusion plus direct dependency for `hibernate-validator`).
They have already been introduced in `5.4.x` and above, but need to land in `5.2.x` and `5.3.x` too.
- See https://github.com/confluentinc/rest-utils/pull/224, https://github.com/confluentinc/rest-utils/pull/246 and https://github.com/confluentinc/rest-utils/pull/249 for the bumps to the latest versions on the higher branches.
    - For the Jetty change, you also might want to check out #147, #151, and #155.
- There will be [a separate PR for `5.3.x`](https://github.com/confluentinc/rest-utils/pull/253) as there are conflicts when merging this there anyway.
- We do bump to the latest versions, very similar to the ones we're currently using in Apache Kafka.

The changes might break things downstream, so we should carefully monitor CI for downstream dependencies.

PR #249, opened 2021-05-03
DGS-1621 Upgrade Jetty to 9.4.40

rayokota:
https://confluentinc.atlassian.net/browse/DGS-1621

PR #246, opened 2021-04-27
fix: CVE-2021-28168

xjin-Confluent:
https://nvd.nist.gov/vuln/detail/CVE-2021-28168

PR #253, opened 2021-05-16
KREST-1031 Upgrade Jetty and Jersey to address a number of CVEs

dimitarndimitrov:
See #252 - this merges the changes from there to `5.3.x`, fixing the conflicts along the way. It should then be merged up with `-s ours`, as #246 and #249 have already introduced these changes in `5.4.x` and up.
- If #252 is merged with "Rebase and merge", we should be able to push the changes from here without further rebasing.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -50,6 +50,17 @@
         <dependency>
             <groupId>org.glassfish.jersey.ext</groupId>
             <artifactId>jersey-bean-validation</artifactId>
+            <exclusions>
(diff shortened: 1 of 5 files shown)


dimitarndimitrov:
The CI run is failing because of a downstream dependency failure - blueway's `control-center`.

See https://github.com/confluentinc/rest-utils/pull/252#issuecomment-842114666.

rigelbm (approved):
LGTM

PR #258, opened 2021-05-24
KREST-661: Add support for HTTP/2 on Java 11 and later

AndrewJSchofield:
Added support for HTTP/2 on Java 11 and later. Because HTTP/2 and ALPN are newer than Java 8, adding support for Java 8 as well is not practical. So, the behaviour is:

* If java 11 compatible
  * if (http)
    * configure HTTP and HTTP/2 Clear text connection factories
  * else (https)
    * configure SSL, ALPN, HTTP/2 and HTTP connection factories
* else as before

This means that when the code is running on a Java 11 JVM or later, connections will default to HTTP/2 not HTTP/1.1 if the client can support it. This change in behaviour is appropriate for the major version bump to CP 7.0 but not CP 6.x.

PR is marked as WIP because investigating the best way to introduce the dependency on the `jetty-alpn-java-server` package into the repos that depend on rest-utils.

### DIFF
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -72,7 +72,8 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
 
   private static final Logger log = LoggerFactory.getLogger(ApplicationServer.class);
 
-  private static boolean isJava11Compatible() {
+  // Package-visible for tests
+  static boolean isJava11Compatible() {
     final String versionString = System.getProperty("java.specification.version");
   
     final StringTokenizer st = new StringTokenizer(versionString, ".");
@@ -405,7 +406,8 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
 
   private void addConnectorForListener(HttpConfiguration httpConfiguration,
                                        HttpConnectionFactory httpConnectionFactory,
-                                       URI listener, boolean http2Enabled) {
+                                       URI listener,
+                                       boolean http2Enabled) {
     NetworkTrafficServerConnector connector;
 
     // Default to supporting HTTP/2 for Java 11 and later
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -393,7 +393,8 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
     final HttpConnectionFactory httpConnectionFactory =
             new HttpConnectionFactory(httpConfiguration);
 
-    final boolean http2Enabled = config.getBoolean(RestConfig.HTTP2_ENABLED_CONFIG);
+    // Default to supporting HTTP/2 for Java 11 and later
+    final boolean http2Enabled = isJava11Compatible() && config.getBoolean(RestConfig.HTTP2_ENABLED_CONFIG);
 
     @SuppressWarnings("deprecation")
     List<URI> listeners = parseListeners(config.getList(RestConfig.LISTENERS_CONFIG),
@@ -410,8 +411,7 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
                                        boolean http2Enabled) {
     NetworkTrafficServerConnector connector;
 
-    // Default to supporting HTTP/2 for Java 11 and later
-    if (http2Enabled && isJava11Compatible()) {
+    if (http2Enabled) {
       log.info("Adding listener with HTTP/2: " + listener.toString());
       if (listener.getScheme().equals("http")) {
         // HTTP2C is HTTP/2 Clear text
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -21,6 +21,10 @@ import org.apache.kafka.common.metrics.Gauge;
 import org.apache.kafka.common.metrics.Metrics;
 
 import org.apache.kafka.common.config.ConfigException;
+import org.eclipse.jetty.alpn.server.ALPNServerConnectionFactory;
+import org.eclipse.jetty.http.HttpVersion;
+import org.eclipse.jetty.http2.server.HTTP2ServerConnectionFactory;
+import org.eclipse.jetty.http2.server.HTTP2CServerConnectionFactory;
 import org.eclipse.jetty.jmx.MBeanContainer;
 import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.HttpConfiguration;
@@ -28,6 +32,7 @@ import org.eclipse.jetty.server.HttpConnectionFactory;
 import org.eclipse.jetty.server.NetworkTrafficServerConnector;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
+import org.eclipse.jetty.server.SslConnectionFactory;
 import org.eclipse.jetty.server.handler.ContextHandlerCollection;
 import org.eclipse.jetty.server.handler.DefaultHandler;
 import org.eclipse.jetty.server.handler.HandlerCollection;
@@ -52,6 +57,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.StringTokenizer;
 import java.util.stream.Collectors;
 import java.util.concurrent.BlockingQueue;
 
@@ -65,6 +71,47 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
   private List<NetworkTrafficServerConnector> connectors = new ArrayList<>();
 
   private static final Logger log = LoggerFactory.getLogger(ApplicationServer.class);
+
+  private static final Version VERSION =
+      parseVersion(System.getProperty("java.specification.version"));
+
+  // Package private for testing
+  static Version parseVersion(String versionString) {
+    final StringTokenizer st = new StringTokenizer(versionString, ".");
+    int majorVersion = Integer.parseInt(st.nextToken());
+    int minorVersion;
+    if (st.hasMoreTokens()) {
+      minorVersion = Integer.parseInt(st.nextToken());
+    } else {
+      minorVersion = 0;
+    }
+    return new Version(majorVersion, minorVersion);
+  }
+
+  // Having these as static final provides the best opportunity for compiler optimization
+  public static final boolean IS_JAVA11_COMPATIBLE = VERSION.isJava11Compatible();
+
+  // Package private for testing
+  static class Version {
+    public final int majorVersion;
+    public final int minorVersion;
+
+    private Version(int majorVersion, int minorVersion) {
+      this.majorVersion = majorVersion;
+      this.minorVersion = minorVersion;
+    }
+
+    @Override
+    public String toString() {
+      return "Version(majorVersion=" + majorVersion
+              + ", minorVersion=" + minorVersion + ")";
+    }
+
+    // Package private for testing
+    boolean isJava11Compatible() {
+      return majorVersion >= 11;
+    }
+  }
 
   public ApplicationServer(T config) {
     this(config, createThreadPool(config));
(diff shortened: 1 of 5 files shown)


AndrewJSchofield:
[clabot:check]

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -384,11 +431,38 @@ private void configureConnectors(SslContextFactory sslContextFactory) {
     for (URI listener : listeners) {
       log.info("Adding listener: " + listener.toString());
       NetworkTrafficServerConnector connector;
-      if (listener.getScheme().equals("http")) {
-        connector = new NetworkTrafficServerConnector(this, httpConnectionFactory);
+
+      // Default to supporting HTTP/2 for Java 11 and later
+      if (IS_JAVA11_COMPATIBLE) {
```
dimitarndimitrov, line 436: I think we should have a way to opt out of this new default if we end up hitting some problems with it in some circumstances. I am not sure what would be the most idiomatic way to do this, but having in mind that we don't expect to be using it extensively, I think a custom `RestConfig` property or maybe even a system property can do the job.

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -66,6 +72,47 @@
...
+    public String toString() {
+      return "Version(majorVersion=" + majorVersion
+              + ", minorVersion=" + minorVersion + ")";
+    }
+
+    // Package private for testing
+    boolean isJava11Compatible() {
+      return majorVersion >= 11;
+    }
+  }
```
dimitarndimitrov, line 114: I'd prefer replacing this with a single `static` method that gets the system property, extracts the major version (post Java 8 style), and returns a `boolean`. It will be the same in terms of actual efficiency (because we need to call it just once), but it will be much simpler to read and understand.

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -384,11 +431,38 @@ private void configureConnectors(SslContextFactory sslContextFactory) {
...
+        }
       } else {
-        connector = new NetworkTrafficServerConnector(this, httpConnectionFactory,
-                sslContextFactory);
+        if (listener.getScheme().equals("http")) {
+          connector = new NetworkTrafficServerConnector(this, httpConnectionFactory);
+        } else {
+          connector = new NetworkTrafficServerConnector(this, httpConnectionFactory,
+                  sslContextFactory);
+        }
```
dimitarndimitrov, line 465: nit: What do you think about extracting this in its own method?

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -66,6 +72,47 @@
...
+    } else {
+      minorVersion = 0;
+    }
+    return new Version(majorVersion, minorVersion);
+  }
+
+  // Having these as static final provides the best opportunity for compiler optimization
+  public static final boolean IS_JAVA11_COMPATIBLE = VERSION.isJava11Compatible();
+
+  // Package private for testing
```
dimitarndimitrov, line 94: nit: Are there test changes that were supposed to be part of the PR? I don't think I'm seeing any. Anyway, see my next comment below.

dimitarndimitrov (commented):
Thank you for the PR, the approach looks good to me overall.
Some comments below.

P.S. I suspect that if we wanted to add full support for differences between Java 8 and Java 11 builds, we would have to do some complex things with a dedicated Maven profile that activates depending on the Java version (including having it add dependencies conditionally, having it trigger the compiler plugin on a separate source set, etc.), and we would have to use dependency injection in the places where we want the conditional behaviour. Not something that is needed here though.

core/src/main/java/io/confluent/rest/ApplicationServer.java
AndrewJSchofield: @dimitarndimitrov I think that the `configureConnectors` method doesn't do a lot more than this anyway. Also, the configuration of the connectors isn't used anywhere else. Personally, I don't think there's a heap of value in extracting this into its own method at this point.

core/src/main/java/io/confluent/rest/ApplicationServer.java
AndrewJSchofield: Yes, I agree. Having discuss with @rigelbm, I'll add a test.

core/src/main/java/io/confluent/rest/ApplicationServer.java
rigelbm: +1 to flag protect. I don't think we will have any problem with having it on my default , but it costs us nothing to flag protect.

core/src/main/java/io/confluent/rest/ApplicationServer.java
rigelbm: +1

core/src/main/java/io/confluent/rest/ApplicationServer.java
rigelbm: +1 to @dimitarndimitrov suggestion. Having such a long body inside a for-loop decreases readability a little bit. I would extract the whole body of the for-loop out. I would also in addition possibly extract the body of each branch of the conditional in its own function.

core/src/main/java/io/confluent/rest/ApplicationServer.java
rigelbm: +1, waiting on the test that confirms HTTP/2 works, since I believe the HTTP client we currently use in our tests does not support HTTP/2 at all.

AndrewJSchofield: Tests added.

core/src/main/java/io/confluent/rest/ApplicationServer.java
AndrewJSchofield: for-loop body refactored into own function.

core/src/main/java/io/confluent/rest/ApplicationServer.java
AndrewJSchofield: Done

core/src/main/java/io/confluent/rest/ApplicationServer.java
AndrewJSchofield: Flag added to config. Defaults to true.

core/src/test/java/io/confluent/rest/Http2Test.java
```diff
@@ -0,0 +1,390 @@
...
+  // returns the http response status code.
+  private int makeGetRequestHttp(String url) throws Exception {
+    log.debug("Making GET using HTTP " + url);
+    HttpClient httpClient = new HttpClient();
+    httpClient.start();
+
+    int statusCode = httpClient.GET(url).getStatus();
+    httpClient.stop();
+    return statusCode;
+  }
```
dimitarndimitrov, line 284: I think we should compact the 4 `makeXxxRequestYyy` methods, extracting the common parts into separate methods. All of them do:
```
httpClient.start();

int statusCode = httpClient.GET(url).getStatus();
httpClient.stop();
return statusCode;
```
They only differ in the way they set up the used client - it needs to be an HTTP/2 client in the case of HTTP/2, and it needs to be set up with an `SslContextFactory` using some pre-defined keystore config in the case of HTTPS. I'd say let's extract the `SslContextFactory` creation in one method, and then the client creation depending on whether HTTPS is being used and on whether HTTP/2 is being used.

dimitarndimitrov (commented):
The change is in a pretty good shape now - practically all my comments are related to compacting the code even more, so that it's shorter and simpler.

dimitarndimitrov:
The CI run is currently failing because of the `blueway` downstream dependency, however the problem there seems to be caused by some Kafka changes that broke its build, unrelated to your change here.

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -66,6 +72,21 @@
...
+    final String versionString = System.getProperty("java.specification.version");
+  
+    final StringTokenizer st = new StringTokenizer(versionString, ".");
+    int majorVersion = Integer.parseInt(st.nextToken());
+    int minorVersion;
+    if (st.hasMoreTokens()) {
+      minorVersion = Integer.parseInt(st.nextToken());
+    } else {
+      minorVersion = 0;
+    }
```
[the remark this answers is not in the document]
AndrewJSchofield: Already done :) There was a change since you reviewed because Jenkins picked up the redundant variable.

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -377,28 +398,63 @@ private void configureConnectors(SslContextFactory sslContextFactory) {
     final HttpConnectionFactory httpConnectionFactory =
             new HttpConnectionFactory(httpConfiguration);
 
+    final boolean http2Enabled = config.getBoolean(RestConfig.HTTP2_ENABLED_CONFIG);
```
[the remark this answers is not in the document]
AndrewJSchofield: I did it like that to avoid re-evaluating the `config.getBoolean` for each listener.

core/src/test/java/io/confluent/rest/Http2Test.java
```diff
@@ -0,0 +1,390 @@
...
+
+  private void configServerTruststore(Properties props) {
+    props.put(RestConfig.SSL_TRUSTSTORE_LOCATION_CONFIG, trustStore.getAbsolutePath());
+    props.put(RestConfig.SSL_TRUSTSTORE_PASSWORD_CONFIG, SSL_PASSWORD);
+  }
+
+  private void enableSslClientAuth(Properties props) {
+    props.put(RestConfig.SSL_CLIENT_AUTHENTICATION_CONFIG, RestConfig.SSL_CLIENT_AUTHENTICATION_REQUIRED);
+  }
```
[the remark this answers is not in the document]
AndrewJSchofield: Yup. No longer used as I evolved the test. Thanks.

core/src/main/java/io/confluent/rest/ApplicationServer.java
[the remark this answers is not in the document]
dimitarndimitrov: What I'm suggesting is, let's do the same for `isJava11Compatible()`:
```
final boolean http2Enabled = isJava11Compatible() && config.getBoolean(RestConfig.HTTP2_ENABLED_CONFIG);
```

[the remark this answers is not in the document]
AndrewJSchofield: Ah, got it :)

core/src/test/java/io/confluent/rest/Http2Test.java
```diff
@@ -0,0 +1,340 @@
...
+
+  private void configServerKeystore(Properties props) {
+    props.put(RestConfig.SSL_KEYSTORE_LOCATION_CONFIG, serverKeystore.getAbsolutePath());
+    props.put(RestConfig.SSL_KEYSTORE_PASSWORD_CONFIG, SSL_PASSWORD);
+    props.put(RestConfig.SSL_KEY_PASSWORD_CONFIG, SSL_PASSWORD);
+  }
+
+  private TestRestConfig buildTestConfig(boolean httpListener,
+                                         boolean httpsListener,
+                                        boolean disableHttp2) {
```
dimitarndimitrov, line 121: nit: Indent looks wrong here. Also I would prefer having this flag be `enableHttp2` (the important part being it being `true` if HTTP/2 is enabled, not the actual name) , as this is how we define the flag in `Application.java`.

core/src/test/java/io/confluent/rest/Http2Test.java
```diff
@@ -0,0 +1,340 @@
...
+      if (httpsListener) {
+        props.put(RestConfig.LISTENERS_CONFIG, HTTP_URI + "," + HTTPS_URI);
+      } else {
+        props.put(RestConfig.LISTENERS_CONFIG, HTTP_URI);
+      }
+    }
+    else {
+      assertTrue(httpsListener);
+      props.put(RestConfig.LISTENERS_CONFIG, HTTPS_URI);
+    }
```
dimitarndimitrov, line 133: Personally, I would prefer to always add both the `HTTP_URI` and the `HTTPS_URI`. This will remove the need for this bulky conditional logic, and will significantly simplify the `buildTestConfig` method (its current signature taking 3 `boolean` arguments is a bit error-prone and requires looking back to the definition to check what exactly was their ordering).

dimitarndimitrov (approved):
+1.

The change looks great to me now!
Thanks for addressing a ton of comments (I have added a bit more if you have the appetite for that, but don't consider them blockers).

PR #259, opened 2021-05-27
Fix build problems

AndrewJSchofield:
Fixing a couple of minor build problems that slipped through.

### DIFF
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -394,7 +394,8 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
             new HttpConnectionFactory(httpConfiguration);
 
     // Default to supporting HTTP/2 for Java 11 and later
-    final boolean http2Enabled = isJava11Compatible() && config.getBoolean(RestConfig.HTTP2_ENABLED_CONFIG);
+    final boolean http2Enabled = isJava11Compatible()
+                              && config.getBoolean(RestConfig.HTTP2_ENABLED_CONFIG);
 
     @SuppressWarnings("deprecation")
     List<URI> listeners = parseListeners(config.getList(RestConfig.LISTENERS_CONFIG),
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -394,7 +394,8 @@ private void configureConnectors(SslContextFactory sslContextFactory) {
             new HttpConnectionFactory(httpConfiguration);
 
     // Default to supporting HTTP/2 for Java 11 and later
-    final boolean http2Enabled = isJava11Compatible() && config.getBoolean(RestConfig.HTTP2_ENABLED_CONFIG);
+    final boolean http2Enabled = isJava11Compatible()
+                              && config.getBoolean(RestConfig.HTTP2_ENABLED_CONFIG);
```
dimitarndimitrov, line 398: nit: I think the indent here should be 2x the regular indent, or 4 spaces.

core/src/test/java/io/confluent/rest/Http2Test.java
```diff
@@ -257,7 +259,9 @@ private int makeGetRequestHttp(String url, String clientKeystoreLocation, String
                                 String clientKeyPassword)
       throws Exception {
     log.debug("Making GET using HTTPS " + url);
-    HttpClient httpClient = new HttpClient(buildSslContextFactory());
+    HttpClient httpClient = new HttpClient(buildSslContextFactory(clientKeystoreLocation,
+                                                                  clientKeystorePassword,
+                                                                  clientKeyPassword));
```
dimitarndimitrov, line 264: Ha, interesting - did this (and the similar code below) cause a test failure?

swist (approved):
Looks good to me! Thanks for taking care of it

dimitarndimitrov (approved):
+1.

Thanks for catching and addressing this really quick!

core/src/test/java/io/confluent/rest/Http2Test.java
AndrewJSchofield: The maven target I was using to check the tests didn't actually rebuild then, which was a surprise.

PR #262, opened 2021-06-15
KREST-1346: Add option to apply rate-limit globally across all connections.

rigelbm:
Also removes dos.filter.track.session option, since Kafka REST does not use sessions, so that option has no effect.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -73,6 +73,7 @@ import java.util.function.Consumer;
 
 import javax.servlet.DispatcherType;
 import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
 import javax.ws.rs.core.Configurable;
 
 import io.confluent.rest.auth.AuthUtil;
@@ -608,7 +609,13 @@ public abstract class Application<T extends RestConfig> {
     if (!config.isDosFilterEnabled()) {
       return;
     }
-    FilterHolder filterHolder = new FilterHolder(new DoSFilter());
+    DoSFilter dosFilter;
+    if (!config.getDosFilterRemotePort() && config.getDosFilterTrackGlobal()) {
+      dosFilter = new GlobalDosFilter();
+    } else {
+      dosFilter = new DoSFilter();
+    }
+    FilterHolder filterHolder = new FilterHolder(dosFilter);
     filterHolder.setInitParameter(
         "maxRequestsPerSec", String.valueOf(config.getDosFilterMaxRequestsPerSec()));
     filterHolder.setInitParameter(
@@ -625,8 +632,7 @@ public abstract class Application<T extends RestConfig> {
         "maxIdleTrackerMs", String.valueOf(config.getDosFilterMaxIdleTrackerMs().toMillis()));
     filterHolder.setInitParameter(
         "insertHeaders", String.valueOf(config.getDosFilterInsertHeaders()));
-    filterHolder.setInitParameter(
-        "trackSessions", String.valueOf(config.getDosFilterTrackSessions()));
+    filterHolder.setInitParameter("trackSessions", "false");
     filterHolder.setInitParameter(
         "remotePort", String.valueOf(config.getDosFilterRemotePort()));
     filterHolder.setInitParameter(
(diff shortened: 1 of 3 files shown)


dimitarndimitrov:
@rigelbm [`f1921b2`](https://github.com/confluentinc/rest-utils/pull/262/commits/f1921b22ecfda1ef9d91e4cb7417459784b794a8) belongs to #261, and should not be in this PR, right?

rigelbm:
> @rigelbm [`f1921b2`](https://github.com/confluentinc/rest-utils/pull/262/commits/f1921b22ecfda1ef9d91e4cb7417459784b794a8) belongs to #261, and should not be in this PR, right?

That's just GitHub inability to set diff-bases. I will merge #261 and update this PR soon.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -625,8 +632,7 @@ private void configureDosFilter(ServletContextHandler context) {
         "maxIdleTrackerMs", String.valueOf(config.getDosFilterMaxIdleTrackerMs().toMillis()));
     filterHolder.setInitParameter(
         "insertHeaders", String.valueOf(config.getDosFilterInsertHeaders()));
-    filterHolder.setInitParameter(
-        "trackSessions", String.valueOf(config.getDosFilterTrackSessions()));
+    filterHolder.setInitParameter("trackSessions", "false");
```
dimitarndimitrov, line 635: This actually changes the value we've been using for this parameter which also changes the codepath we'll be using most of the time, even though we don't have a configured session manager. I don't think it will have an effect, but it's probably a good idea for you to check the `DoSFilter` code too for that (if you haven't already).

core/src/test/java/io/confluent/rest/DosFilterTest.java
```diff
@@ -48,23 +55,142 @@ public void dosFilterEnabled_throttlesRequests() throws Exception {
...
 
     // Following requests should all be throttled.
     for (int i = 0; i < 100; i++) {
-      Response response2 = client.target(server.getURI()).path("/foo").request().get();
-      assertEquals(Status.TOO_MANY_REQUESTS.getStatusCode(), response2.getStatus());
+      CloseableHttpResponse response2 = ephemeralClient.execute(request);
+      assertEquals(
+          Status.TOO_MANY_REQUESTS.getStatusCode(), response2.getStatusLine().getStatusCode());
+      response2.close();
+    }
```
dimitarndimitrov, line 78: I'm a bit concerned about my previous suggestion to bump up the number of requests, but for now we haven't seen flakiness here...

core/src/test/java/io/confluent/rest/DosFilterTest.java
```diff
@@ -48,23 +55,142 @@ public void dosFilterEnabled_throttlesRequests() throws Exception {
...
 
-    Client client = ClientBuilder.newClient();
+    HttpGet request =
+        new HttpGet(UriBuilder.fromUri(server.getURI()).path("/foo").build());
+
+    CloseableHttpClient ephemeralClient =
+        HttpClients.custom()
+            .setConnectionManager(new BasicHttpClientConnectionManager())
+            .setKeepAliveStrategy((httpResponse, httpContext) -> -1)
+            .build();
```
dimitarndimitrov, line 65: nit: Let's extract this and the persistent client creation in separate static factory methods.

dimitarndimitrov (approved):
+1.

Awesome, I'll sleep much better now! I really don't love `DoSFilter`, and therefore I don't love that we support all of its configuration directly, but it is what it is.

Once this goes in, we'll need some `cc-spec-kafka` changes, right?

core/src/main/java/io/confluent/rest/Application.java
rigelbm: As I said in the PR description, `trackSessions` has no effect, since we don't use sessions. Might as well disable it explicitly.

core/src/test/java/io/confluent/rest/DosFilterTest.java
rigelbm: Done.

core/src/test/java/io/confluent/rest/DosFilterTest.java
rigelbm: Acknowledged.

rigelbm:
> Once this goes in, we'll need some `cc-spec-kafka` changes, right?
That's correct.

PR #263, opened 2021-06-21
KREST-1403: Set RFC7230 mode for connections

AndrewJSchofield:
There is a difference in URI handled between HTTP/1.1 and HTTP/2 in Jetty 9.4. Essentially, this is related to security and preventing malicious HTTP clients from using clever URIs containing / and .. from discovering information they have no right to obtain. So, Jetty 9.4 changed its behaviour to prevent use of these characters in URIs which then caused compatibility problems.

As a result, Jetty changed again so that HTTP/1.1 connections by default use a legacy HTTP compliance mode, implemented as a parameter to a constructor. HTTP/2 connections are not constructed in the same way and do not benefit from the same mechanism.

Nowadays rest-utils supports both HTTP/1.1 and HTTP/2. Any client that can support HTTP/2 gets that in preference to HTTP/1.1. Schema Registry users sometimes use URI-encoded / characters in subject names, and this breaks with HTTP/2 so far.

The resolution to the issue is to use the same HTTP compliance mode for HTTP/2 as is already used for HTTP/1.1.

This problem was originally reported as https://confluentinc.atlassian.net/browse/ESCALATION-5329.

### DIFF
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -22,6 +22,7 @@ import org.apache.kafka.common.metrics.Metrics;
 
 import org.apache.kafka.common.config.ConfigException;
 import org.eclipse.jetty.alpn.server.ALPNServerConnectionFactory;
+import org.eclipse.jetty.http.HttpCompliance;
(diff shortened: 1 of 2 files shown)


rigelbm (approved):
LGTM

Thanks @AndrewJSchofield !!!

PR #266, opened 2021-07-09
Named listeners

mhowlett:
Support for named listeners analogous to similar capability in Kafka.


mhowlett:
@rigelbm - i'm having second thoughts about putting the config related changes in rest-utils - please hold off reviewing until i've sorted that through.

PR #276, opened 2021-08-05
Allow disabling OPTIONS on REST proxy

ehumber:
Using the sample application

REJECT_OPTIONS_REQUEST set to default (false)
% curl http://localhost:8080/hello -X OPTIONS
HEAD, GET, OPTIONS%                                                                                                                                                                           

REJECT_OPTIONS_REQUEST set to true
% curl http://localhost:8080/hello -X OPTIONS
% 

Tested with both no auth and basic auth enabled, as the code path is slightly different between the two.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -452,6 +452,13 @@ public abstract class Application<T extends RestConfig> {
       context.setSecurityHandler(createBasicSecurityHandler());
     } else if (enableBearerAuth(authMethod)) {
       context.setSecurityHandler(createBearerSecurityHandler());
+    } else {
+      AuthUtil.createDisableOptionsConstraint(config)
+          .ifPresent(optionsConstraint -> {
+            ConstraintSecurityHandler securityHandler = new ConstraintSecurityHandler();
+            securityHandler.addConstraintMapping(optionsConstraint);
+            context.setSecurityHandler(securityHandler);
+          });
     }
   }
 
@@ -550,7 +557,9 @@ public abstract class Application<T extends RestConfig> {
     securityHandler.setIdentityService(createIdentityService());
     securityHandler.setRealmName(realm);
     AuthUtil.createUnsecuredConstraints(config)
-            .forEach(securityHandler::addConstraintMapping);
+        .forEach(securityHandler::addConstraintMapping);
+    AuthUtil.createDisableOptionsConstraint(config)
+        .ifPresent(securityHandler::addConstraintMapping);
 
     return securityHandler;
   }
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/confluent/rest/auth/AuthUtil.java
```diff
@@ -140,10 +152,35 @@ private static ConstraintMapping createConstraint(
...
+      forbidConstraint.setAuthenticate(true);
+
+      ConstraintMapping forbidMapping = new ConstraintMapping();
+      forbidMapping.setMethod("OPTIONS");
+      forbidMapping.setPathSpec("/*");
+      forbidMapping.setConstraint(forbidConstraint);
+      return Optional.of(forbidMapping);
+
+    }
+    return Optional.ofNullable(null);
```
rigelbm, line 184: Optional.empty()

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -452,6 +452,13 @@ protected void configureSecurityHandler(ServletContextHandler context) {
       context.setSecurityHandler(createBasicSecurityHandler());
     } else if (enableBearerAuth(authMethod)) {
       context.setSecurityHandler(createBearerSecurityHandler());
+    } else {
```
rigelbm, line 455: I don't understand. What happens if you are using Basic Auth or Bearer Auth? The REJECT_OPTIONS_REQUEST config is ignored? Whoever uses this config is going to very security-aware, so they are probably going to be using authentication as well.

rigelbm: Ow, you said you tested with Basic Auth and Bearer Auth. Sorry, couldn't find that piece of code. Let me look again, and please ignore. :)

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -550,7 +557,9 @@ protected ConstraintSecurityHandler createSecurityHandler() {
     securityHandler.setIdentityService(createIdentityService());
     securityHandler.setRealmName(realm);
     AuthUtil.createUnsecuredConstraints(config)
-            .forEach(securityHandler::addConstraintMapping);
+        .forEach(securityHandler::addConstraintMapping);
+    AuthUtil.createDisableOptionsConstraint(config)
```
rigelbm, line 561: This is the one for Basic Auth and Bearer Auth. :)

core/src/test/java/io/confluent/rest/ApplicationTest.java
```diff
@@ -66,10 +66,12 @@
 import org.eclipse.jetty.security.LoginService;
 import org.eclipse.jetty.security.authentication.BasicAuthenticator;
 import org.eclipse.jetty.server.ServerConnector;
+import org.eclipse.jetty.servlet.ServletContextHandler;
 import org.eclipse.jetty.util.security.Constraint;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
+import scala.App;
```
rigelbm, line 74: nit: unused

rigelbm (approved):
LGTM

PR #277, opened 2021-08-13
Fix request logger to print request time in milliseconds

milosimpson:
manually verified
request log changes from `GETsT` to a number that looks like request time

```
from
GET /test/endpoint HTTP/1.1" 200 21 "-" "curl/7.64.1" GETsT (io.confluent.rest-utils.requests:62)
to
GET /test/endpoint HTTP/1.1" 200 21 "-" "curl/7.64.1" 2 (io.confluent.rest-utils.requests:62)
```

I think `%{ms}T` is probably what was meant.   Request time in milliseconds.
* https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/server/CustomRequestLog.html

```
%{UNIT}T	The time taken to serve the request, in a time unit given by UNIT. Valid units are ms for milliseconds, us for microseconds, and s for seconds. Using s gives the same result as %T without any format; using us gives the same result as %D.
```

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -131,7 +131,9 @@ public abstract class Application<T extends RestConfig> {
 
     Slf4jRequestLogWriter logWriter = new Slf4jRequestLogWriter();
     logWriter.setLoggerName(config.getString(RestConfig.REQUEST_LOGGER_NAME_CONFIG));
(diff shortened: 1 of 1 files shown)


jeffhuang26 (approved):

rigelbm (approved):
LGTM

ijuma:
Can we please fix apache/kafka too:
https://github.com/apache/kafka/search?q=%25msT

:)

milosimpson:
Made a ticket for the AK fix.

PR #278, opened 2021-08-18
Add support for PROXY protocol

bob-barrett:


core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -424,14 +429,17 @@ private void configureConnectors(SslContextFactory sslContextFactory) {
         SUPPORTED_URI_SCHEMES, "http");
 
     for (NamedURI listener : listeners) {
-      addConnectorForListener(httpConfiguration, httpConnectionFactory, listener, http2Enabled);
+      addConnectorForListener(httpConfiguration, httpConnectionFactory, listener,
+              http2Enabled, proxyProtocolEnabled);
```
dimitarndimitrov, line 433: nit: Personally I prefer formatting where either all method parameters are on the same line or each method parameter is on its own line, but unlike `kafka-rest`, `rest-utils` doesn't have a well established formatting rule about that, so you don't necessarily need to change this.

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -454,11 +466,19 @@ private void addConnectorForListener(HttpConfiguration httpConfiguration,
...
+
+        connectionFactories.add(sslConnectionFactory);
+        connectionFactories.add(alpnConnectionFactory);
+        connectionFactories.add(h2ConnectionFactory);
+        connectionFactories.add(httpConnectionFactory);
       }
 
+      connector = new NetworkTrafficServerConnector(this, null, null, null, 0, 0,
+              connectionFactories.toArray(new ConnectionFactory[0]));
+
```
dimitarndimitrov, line 481: nit: I think this and the similar expression at the end of the non-HTTP2 path can be combined by being moved out of the HTTP2 if-else.

dimitarndimitrov (approved):
+1.

Looks good to me overall, although I think we should try to test this on the 4 different paths - HTTPS vs HTTP x HTTP/2 vs HTTP/1.1.

core/src/main/java/io/confluent/rest/ApplicationServer.java
bob-barrett: The existing practice in this file at least seems to be mostly multiple parameters per line, so I'm going to stick with that

core/src/main/java/io/confluent/rest/ApplicationServer.java
bob-barrett: I originally didn't do this because there's one additional thing that has to happen in the HTTP2 case: https://github.com/confluentinc/rest-utils/blob/master/core/src/main/java/io/confluent/rest/ApplicationServer.java#L470. I've made this change now, and I think it's fine, but I could go either way.

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -424,14 +429,17 @@ private void configureConnectors(SslContextFactory sslContextFactory) {
         SUPPORTED_URI_SCHEMES, "http");
 
     for (NamedURI listener : listeners) {
-      addConnectorForListener(httpConfiguration, httpConnectionFactory, listener, http2Enabled);
+      addConnectorForListener(httpConfiguration, httpConnectionFactory, listener,
+              http2Enabled, proxyProtocolEnabled);
     }
   }
 
   private void addConnectorForListener(HttpConfiguration httpConfiguration,
```
dimitarndimitrov, line 437: Not really a comment to your change, but `addConnectorForListener` has become really hard to read. We should probably have an early return method that does the connection factories juggling and returns the set of relevant factories, and another one which creates a connector out of these.

dimitarndimitrov (approved):
+1.

Thanks for the change, and especially for the test!
Some nits below.

PR #280, opened 2021-08-23
KREST-578: Add configurable DosFilter support. (#238)

ehumber:
Cherry pick of commit 51e387e to backport the fix to 5.2.x and later.

This PR exposes per-application DosFilter support. See https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/servlets/DoSFilter.html for more information about the options available.

Config names are kept equal to the ones in Jetty's DosFilter. That unfortunately includes a usage of "whitelist", which otherwise would not meet Confluent's bar.

This PR also changes several tests to use a random port, to avoid race conditions with port binding.


rigelbm (approved):
LGTM

PR #284, opened 2021-09-01
KREST-1671 Add counter metrics

ehumber:
At present, rest-utils only collects rate based metrics.  Absolute metrics will also be useful.

This commit adds metrics to count, both cumulatively and also on a windowed basis:
 - requests that finish successfully
 - error status responses (for any error response)
 - error status response, per grouping of errors.  The groupings are unknown, 1xx, 2xx, 3xx, 4xx and 5xx

### DIFF
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -16,6 +16,7 @@
 
 package io.confluent.rest.metrics;
 
+import org.apache.kafka.common.metrics.stats.CumulativeCount;
 import org.glassfish.jersey.server.ContainerRequest;
 import org.glassfish.jersey.server.ContainerResponse;
 import org.glassfish.jersey.server.model.Resource;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -189,6 +190,14 @@ public MethodMetrics(ResourceMethod method, PerformanceMetric annotation, Metric
 
       this.requestSizeSensor = metrics.sensor(getName(method, annotation, "request-size"));
       MetricName metricName = new MetricName(
+          getName(method, annotation, "request-count-windowed"), metricGrpName,
+          "The request count using a windowed counter", metricTags);
+      this.requestSizeSensor.add(metricName, new WindowedCount());
+      metricName = new MetricName(
+          getName(method, annotation, "request-count-cumulative"), metricGrpName,
+          "The request count using a cumulative counter", metricTags);
+      this.requestSizeSensor.add(metricName, new WindowedCount());
```
rigelbm, line 199: Should this be CumulativeCount?

rigelbm (approved):
LGTM

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
ehumber: Oh, good spot!

ehumber: I have updated the unit tests to spot this too.

PR #285, opened 2021-09-09
KREST-1671 remove cumulative metric counts as we aren't using them

ehumber:
When I first added the metrics I wasn't sure whether I needed the windowed count or the cumulative, so added both.

I can't think of an occasion where we want cumulative counts, and as suggested by Rigel, I'm removing them

### DIFF
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -16,7 +16,6 @@
 
 package io.confluent.rest.metrics;
 
-import org.apache.kafka.common.metrics.stats.CumulativeCount;
 import org.glassfish.jersey.server.ContainerRequest;
 import org.glassfish.jersey.server.ContainerResponse;
 import org.glassfish.jersey.server.model.Resource;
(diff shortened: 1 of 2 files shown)


rigelbm (approved):
LGTM

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -193,10 +192,6 @@ public MethodMetrics(ResourceMethod method, PerformanceMetric annotation, Metric
           getName(method, annotation, "request-count-windowed"), metricGrpName,
```
dimitarndimitrov, line 192: If we have settled on an approach, I think we should remove the qualifier differentiating from the alternative. I.e. I think we should use `X` instead of `X-windowed` everywhere.
I do realize this will also affect the `cc-dd-agent` and Datadog changes, but I think we should do it, in which case now is the best time to do it.

ehumber: @dimitarndimitrov 

I did have a think about that before I PR-ed and my conclusions was to leave in the "windowed".

A count metric can be windowed or cumulative, and without the qualifier it's not clear which type this is.

But, I don't have strong feelings, and you're right, now is the time to remove it if we do want to.  

Let me know what you think

dimitarndimitrov: I think the big question here is - can the additional bit of information coming from the qualifier be missed somehow when we're processing the metric values?
I would say no, because with cumulative metrics, the stream of values is guaranteed to be monotonically increasing, and with windowed metrics, at least for count, the stream of values is guaranteed to not be monotonically increasing - I think we will immediately, even subconsciously infer this when working with the metric.

And some weaker arguments that nevertheless did add to the above for me personally:
1. "windowed" provides only partial context about the actual implementation of the metric - one other thing that is not provided there is what is the size of the aggregating window, and another one is what's the aggregation function used within that window.
2. Sampled metrics are a big part of the Kafka Metrics library, but I don't think I've seen any other metric having a qualifier indicating if it's sampled / windowed.

Having said that, this is a big improvement in my eyes, and I will not be too concerned if you do keep the suffix :)

ehumber: I agree with what you're saying.  Removing :)

dimitarndimitrov (approved):
+1.

Thank you very much for addressing these!

PR #287, opened 2021-09-24
SEC-2720: Add SecureRequestCustomizer for correct resolution of http vs https requests.

niteshmor:
This pull request adds a `SecureRequestCustomizer()` to the configuration in case SSL is set up. 
Without the `SecureRequestCustomizer()`, if a server-side application calls `.getScheme()` on incoming requests, it gets `http` even for HTTPS requests. 

Added a test to validate the behavior. 

Notes to reviewers: 
- This bug doesn't affect 6.2.x, which I have verified manually. The test passes when backported to 6.2.x without the need for an explicit customizer.
- The first commit in this PR is only the test but not the fix, so you can inspect that this is indeed an issue. The second commit adds the fix, so build no. 2 should pass.
- The exact reason that caused this change of behavior from `6.2.x` => `7.0.x` is still to be investigated.

### DIFF
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -392,9 +392,6 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
     final HttpConfiguration httpConfiguration = new HttpConfiguration();
     httpConfiguration.setSendServerVersion(false);
 
-    // we need to set below only if listener is https
-    httpConfiguration.addCustomizer(new SecureRequestCustomizer());
-
     final HttpConnectionFactory httpConnectionFactory =
             new HttpConnectionFactory(httpConfiguration);
 
@@ -407,6 +404,11 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
             config.getInt(RestConfig.PORT_CONFIG), Arrays.asList("http", "https"), "http");
 
     for (URI listener : listeners) {
+      if (listener.getScheme().equals("https")) {
+        if (httpConfiguration.getCustomizer(SecureRequestCustomizer.class) == null) {
+          httpConfiguration.addCustomizer(new SecureRequestCustomizer());
+        }
+      }
       addConnectorForListener(httpConfiguration, httpConnectionFactory, listener, http2Enabled);
     }
   }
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -391,6 +392,9 @@ private void configureConnectors(SslContextFactory sslContextFactory) {
     final HttpConfiguration httpConfiguration = new HttpConfiguration();
     httpConfiguration.setSendServerVersion(false);
 
+    // we need to set below only if listener is https
+    httpConfiguration.addCustomizer(new SecureRequestCustomizer());
```
omkreddy, line 396: I think we may want to set if `https` listener in configured
```suggestion
    //moved line 405-406 to here
    @SuppressWarnings("deprecation")
    List<URI> listeners = parseListeners(config.getList(RestConfig.LISTENERS_CONFIG),
            config.getInt(RestConfig.PORT_CONFIG), Arrays.asList("http", "https"), "http");
            
    for (URI listener : listeners) {
      if (listener.getScheme().equals("https")) {
        if (httpConfiguration.getCustomizer(SecureRequestCustomizer.class) == null)
          httpConfiguration.addCustomizer(new SecureRequestCustomizer());
      }
    }

omkreddy:
looks like issue is introduced in https://github.com/confluentinc/rest-utils/pull/258

We started using new [NetworkTrafficServerConnector constructor]( https://github.com/eclipse/jetty.project/blob/jetty-9.4.43.v20210629/jetty-server/src/main/java/org/eclipse/jetty/server/NetworkTrafficServerConnector.java#L59) at   [ApplicationServer.java#L436](https://github.com/confluentinc/rest-utils/blob/7.0.x/core/src/main/java/io/confluent/rest/ApplicationServer.java#L436)
 which do not call [AbstractConnectionFactory.java#L139]( https://github.com/eclipse/jetty.project/blob/jetty-9.4.43.v20210629/jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnectionFactory.java#L139)


Previously we used  [NetworkTrafficServerConnector constructor]( https://github.com/eclipse/jetty.project/blob/jetty-9.4.43.v20210629/jetty-server/src/main/java/org/eclipse/jetty/server/NetworkTrafficServerConnector.java#L49 ) which internally calls [AbstractConnectionFactory.java#L139 ](https://github.com/eclipse/jetty.project/blob/jetty-9.4.43.v20210629/jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnectionFactory.java#L139)
which furher adds SecureRequestCustomizer

core/src/main/java/io/confluent/rest/ApplicationServer.java
niteshmor: Good point. Pushed an update along the lines of your suggestion. Should've paid more attention in the first place :)

niteshmor:
This repo's build + tests pass. Downstream validation will fail, but that's unrelated. That also failed in build no 2. 
Merging.

PR #291, opened 2021-11-11
Test "custom methods" support in rest-utils

milosimpson:


core/src/test/java/io/confluent/rest/CustomMethodsTest.java
```diff
@@ -0,0 +1,165 @@
...
+    Response response = ClientBuilder.newClient()
+        .target(server.getURI())
+        .path("/foo/v2/collection/pants")
+        .request().get();
+    assertEquals(200, response.getStatus());
+    String responseBody = response.readEntity(String.class);
+    assertEquals("base pants", responseBody);
+  }
+
+  @Test
```
milosimpson, line 101: @codyaray

core/src/test/java/io/confluent/rest/CustomMethodsTest.java
```diff
@@ -0,0 +1,165 @@
...
+  public void testIncorrectBaseCall() {
+    // namely that you if you have an id of `pants:search` then it will
```
codyaray, line 103: yep, this totally makes sense.

core/src/test/java/io/confluent/rest/CustomMethodsTest.java
```diff
@@ -0,0 +1,165 @@
...
+  @Test
+  public void testCustomMethodCall() {
+
+    String[] tests = new String[]{
+            "pants",
+            "shoes",
+            "pants-shoes", // test hyphen
+            "pants:shoes", // test colon
+            "pants:search", // test colon + string that matches custom method
+            "pants:search:search", // why not
```
codyaray, line 124: cool that the custom method works even with colons in the IDs too.

but of course, we can't have colons in the IDs because the base method doesn't work (as you noted above)

codyaray (approved):
this is awesome, thank you so much!

PR #292, opened 2021-11-18
Upgrade Netty Version

xjin-Confluent:
This is to upgrade Netty to 4.1.68.Final.


xjin-Confluent:
KSQL downstream validation failed because of spotbug check. I will merge and monitor their branch build.

PR #293, opened 2021-12-01
KREST-3148 enable ip based rate limiting

ehumber:
Before this PR the code checked if the "global" rate limiter was turned on and if the "per port" configuration was set.  If so, the dos filters were applied globally, across all connections.  If not, a per connection rate limiter was created.

The global property is `dos.filter.track.global` which is set to true in cc-spec-kafka (although the in code default is false) and its intention is to do dos rate limiting across all connections, ie globally.

The per port property is `dos.filter.remote.port` which is set to false in cc-spec-kafka and also in code. The intent of this property is to apply the dos filters per ip-addres:port rather than just ip-address.

Before this PR, the dos filters were just applied globally, although the documentation for the `dos.filter.max.requests.per.sec` rate limit does state "per connection" this wasn't actually true in practice for our configuration

The intent of this code change is to have both a per connection `dos.filter.max.requests.per.sec` rate limit as well as the existing global rate limit across all connections.

The `dos.filter.max.requests.per.sec` filter is documented as applying per connection, so while it would be nice to use this for the global limit, as that is how it is already being used, I had to make a new global rate limiting property to maintain consistency.

If this is unset, it defaults to the existing value for `dos.filter.max.requests.per.sec`.

In all cases, the per ip address dos filter is applied (this is new, previously it was not in place if you had global rate limiting turned on).  If dos.filter.track.global is on and ip-addres:port is off, you also get the global limiter, as you did before.

This is all possible because of the `proxy.protocol.enabled` property (false in code, true in launch darkly), which was added to ensure that the ip address from the header is used for rate limiting.  This means that for both AWS and where a load balancer is used, the original ip address is used, not the load balancer one.

### DIFF
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -371,17 +371,16 @@ public class RestConfig extends AbstractConfig {
   private static final String DOS_FILTER_MAX_REQUESTS_PER_SEC_CONFIG =
       "dos.filter.max.requests.per.sec";
   private static final String DOS_FILTER_MAX_REQUESTS_PER_SEC_DOC =
-      "Maximum number of requests from a connection per second. Requests in excess of this are "
-          + "first delayed, then throttled. Default is 25.";
+      "Maximum number of requests per second for the REST instance. Requests in excess of this "
+          + "are first delayed, then throttled. Default is 25.";
   private static final int DOS_FILTER_MAX_REQUESTS_PER_SEC_DEFAULT = 25;
 
-  private static final String DOS_FILTER_MAX_REQUESTS_GLOBAL_PER_SEC_CONFIG =
-      "dos.filter.max.requests.global.per.sec";
-  private static final String DOS_FILTER_MAX_REQUESTS_GLOBAL_PER_SEC_DOC =
-      "Maximum number of requests per second for the rest instance. Requests in excess of this are "
-          + "first delayed, then throttled. If set to the default value of -1, then the value of "
-          + "dos.filter.max.requests.per.sec is used.";
-  private static final int DOS_FILTER_MAX_REQUESTS_GLOBAL_PER_SEC_DEFAULT = -1;
+  private static final String DOS_FILTER_MAX_REQUESTS_PER_CONNECTION_PER_SEC_CONFIG =
+      "dos.filter.max.requests.per.connection.per.sec";
+  private static final String DOS_FILTER_MAX_REQUESTS_PER_CONNECTION_PER_SEC_DOC =
+      "Maximum number of requests per second per ipaddress for the rest instance. "
+          + "Requests in excess of this are first delayed, then throttled.";
+  private static final int DOS_FILTER_MAX_REQUESTS_PER_CONNECTION_PER_SEC_DEFAULT = 25;
 
   private static final String DOS_FILTER_DELAY_MS_CONFIG = "dos.filter.delay.ms";
   private static final String DOS_FILTER_DELAY_MS_DOC =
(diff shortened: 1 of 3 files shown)


ehumber:
https://app.launchdarkly.com/default/stag/features/kafka_api.config.proxy_protocol.enabled/targeting will need to be set to true once this PR has gone through

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -375,6 +375,14 @@
           + "first delayed, then throttled. Default is 25.";
   private static final int DOS_FILTER_MAX_REQUESTS_PER_SEC_DEFAULT = 25;
 
+  private static final String DOS_FILTER_MAX_REQUESTS_GLOBAL_PER_SEC_CONFIG =
+      "dos.filter.max.requests.global.per.sec";
+  private static final String DOS_FILTER_MAX_REQUESTS_GLOBAL_PER_SEC_DOC =
+      "Maximum number of requests per second for the rest instance. Requests in excess of this are "
+          + "first delayed, then throttled. If set to the default value of -1, then the value of "
+          + "dos.filter.max.requests.per.sec is used.";
+  private static final int DOS_FILTER_MAX_REQUESTS_GLOBAL_PER_SEC_DEFAULT = -1;
```
rigelbm, line 384: I think we should:

 * Remove `DOS_FILTER_REMOTE_PORT_CONFIG`
 * Remove `DOS_FILTER_TRACK_GLOBAL_CONFIG`.
 * Add `DOS_FILTER_PER_CONNECTION_ENABLED_CONFIG`
 * Add `DOS_FILTER_PER_CONNECTION_MAX_REQUESTS_PER_SECOND_CONFIG`
 * Profit

rigelbm (approved):
LGTM

bob-barrett:
> https://app.launchdarkly.com/default/stag/features/kafka_api.config.proxy_protocol.enabled/targeting will need to be set to true once this PR has gone through

@ehumber Just to clarify, the https://app.launchdarkly.com/default/stag/features/kafka_api.config.proxy_protocol.enabled/targeting feature flag will apply to Kafka API, but not to embedded REST. ERP uses the Kafka feature flag, https://app.launchdarkly.com/default/devel/features/kafka.config.proxy_protocol_version/targeting. This is because ERP and the brokers share a load balancer, so we need to make sure that both of them have the feature enabled before enabling PROXY headers on the load balancer.

In order to send PROXY headers from the LBs, the https://app.launchdarkly.com/default/devel/features/kafka.enable_proxy_protocol_lb/targeting flag also needs to be set to true. This means that we need to make sure that that flag is enabled before enabling the IP rate limiting, otherwise ERP will still see the LB IP address and will probably overthrottle. I'm currently in the process of enabling this flag in production, but currently it's only been enabled for most AWS multi-tenant clusters, not any Dedicated ones.

ehumber:
Hi @bob-barrett 

Thanks for the clarification, and for spotting the comment hidden above!  

When I was doing the work for this I saw

https://confluentinc.atlassian.net/wiki/spaces/CNKAF/pages/2109484445/1-pager+Client+IP+Preservation+In+AWS

which says

` impose a connection rate limit on each IP address, but this requires Kafka to be aware of the client's IP. This is already the case in GCP and Azure, but in AWS, brokers only get the IP address of the load balancer.`

I think I pasted the wrong launch darkly setting above. I remember when I looked at the time I saw that all AWS clusters were enabled with the proxy flag:

image at https://app.launchdarkly.com/default/prod/features/kafka.config.proxy_protocol_version/targeting

so I thought IP forwarding was in place.

I'd missed seeing the flag https://app.launchdarkly.com/default/prod/features/kafka.enable_proxy_protocol_lb/targeting though :(

Do you have a link to something tracking the enabling of the https://app.launchdarkly.com/default/prod/features/kafka.enable_proxy_protocol_lb/targeting in production so that I know when it's OK to do ip based rate limiting in REST?

PR #294, opened 2021-12-02
Enforce Netty for children

xjin-Confluent:
This is to make sure that all children will enforce io.netty:netty-codec at 4.1.68.Final version.


xjin-Confluent:
@linouk23  I will merge for now as this is to unblock a fix.

linouk23:
I might be lacking access to approve PRs but LGTM via a comment 😁

PR #296, opened 2021-12-02
DGS-2581: Catch TimeoutException and map to 408 

xiaoyali97:
Catch TimeoutException and map to 408 when reading input stream from bad client

jira: https://confluentinc.atlassian.net/browse/DGS-2581


rayokota (approved):
Thanks @xiaoyali97 !  LGTM

PR #300, opened 2022-01-02
TRIVIAL: Change the Maven repository protocol: http → https

dongjinleekr:
From 3.8.1, maven blocks [the HTTP repositories for CVE-2021-26291](https://maven.apache.org/docs/3.8.1/release-notes.html#cve-2021-26291) so rest-utils the build fails (below).

This PR fixes this issue by updating the repository configuration from http to https.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,9 +17,9 @@
     <version>7.2.0-0</version>
     <organization>
         <name>Confluent, Inc.</name>
-        <url>http://confluent.io</url>
+        <url>https://confluent.io</url>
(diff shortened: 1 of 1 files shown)


ewencp (approved):
LGTM. We also have the confluent-snapshots repo. I think all references to that already use https, but you should test e.g. a ksql build to be sure.

dongjinleekr:
@ewencp Sure, I also checked ksqldb and confluent-common and found that they are already using `https`. :smiley: (below)

PR #303, opened 2022-01-19
KREST-3660-Junit5-ify rest-utils

ehumber:
Move to junit 5 only so that I can move kafka-rest to junit 5 (and that's because I need to move ce-kafka-rest to junit 5 only to address some failing tests caused by dependency mistmatches)

Downstream validation passes - eg https://jenkins.confluent.io/job/confluentinc-pr/job/rest-utils/job/PR-303/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ target/
 *.iml
 .idea/
 .ipr
+.DS_Store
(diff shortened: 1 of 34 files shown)


rigelbm (approved):
LGTM

PR #306, opened 2022-01-24
KREST-3660 Remove junit 4 dependencies that come in via jersey test

ehumber:
Running mvn dependency:tree shows that jersey pulls in junit 4.

There's no version of jersey that ships junit 5, but compatibility for using it came in at jersey 2.35, so I've updated the version and excluded junit4

https://github.com/eclipse-ee4j/jersey/pull/4824

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -22,6 +22,13 @@
         <dependency>
             <groupId>org.glassfish.jersey.containers</groupId>
             <artifactId>jersey-container-servlet</artifactId>
+            <version>${jersey.version}</version>
+            <exclusions>
+                <exclusion>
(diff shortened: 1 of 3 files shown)


rigelbm (commented):
That looks good, except for one thing: https://github.com/confluentinc/ce-kafka/blob/master/gradle/dependencies.gradle#L100

ehumber:
I've reverted to jersey 2.34.  I tested downstream in kafka-rest with junit 4 manually excluded and jersey 2.34 and the jersey tests are happy.

So, I think this is fine with 2.34, until I find some evidence that it is not :)

rigelbm (approved):
LGTM

PR #308, opened 2022-03-21
KREST-3446 ensure there is always a minimum value for the thread pool queue

ehumber:
This minimum calculation is copied for the jetty code which uses the same formula when no request queue is configured (which is actually the recommened way of starting up the application).

### DIFF
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -71,6 +71,8 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
   private final List<Application<?>> applications;
   private final SslContextFactory sslContextFactory;
 
+  private static int threadPoolRequestQueueCapacity;
+
   private List<NetworkTrafficServerConnector> connectors = new ArrayList<>();
 
   private static final Logger log = LoggerFactory.getLogger(ApplicationServer.class);
@@ -573,7 +575,7 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
    * @return the capacity of the queue in the pool.
    */
   public int getQueueCapacity() {
-    return config.getInt(RestConfig.REQUEST_QUEUE_CAPACITY_CONFIG);
+    return threadPoolRequestQueueCapacity;
   }
 
   static Handler wrapWithGzipHandler(RestConfig config, Handler handler) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -603,8 +605,16 @@ private static ThreadPool createThreadPool(RestConfig config) {
     log.info("Initial capacity {}, increased by {}, maximum capacity {}.",
             initialCapacity, growBy, maxCapacity);
 
+    if (initialCapacity > maxCapacity) {
+      threadPoolRequestQueueCapacity = Math.max(initialCapacity, 8) * 1024;
+      log.warn("request.queue.capacity is less than request.queue.capacity.init, invalid config. "
+          + "Setting request.queue.capacity to at least 1024 * request.queue.capacity.init.");
```
dimitarndimitrov, line 611: I think this violates the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

It's unintuitive and I think not very practical to have the `maxCapacity` be set to 10 if both it and the initial capacity are configured to be 10, but to more than 10k if it is configured to be 9 while the initial capacity is configured to be 10. Let's just set it to `initialCapacity` in that case - note that there's a specific intent here, unlike the `null` queue case where this very generous sizing originates from.

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -71,6 +71,8 @@
   private final List<Application<?>> applications;
   private final SslContextFactory sslContextFactory;
 
+  private static int threadPoolRequestQueueCapacity;
```
dimitarndimitrov, line 74: nit: If this is being returned by a public API, it probably should be `volatile`.

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -71,7 +71,7 @@
   private final List<Application<?>> applications;
   private final SslContextFactory sslContextFactory;
 
-  private static int threadPoolRequestQueueCapacity;
+  private static volatile  int threadPoolRequestQueueCapacity;
```
dimitarndimitrov, line 74: super-nit: I think there's one redundant space after the `volatile`... :D

core/src/test/java/io/confluent/rest/ApplicationServerTest.java
```diff
@@ -250,7 +250,7 @@ public void testInvalidThreadPoolConfigQueueCapacityValid() throws Exception {
 
     ApplicationServer applicationServer = new ApplicationServer(config);
     applicationServer.start();
-    assertEquals( 1024*9, applicationServer.getQueueCapacity());
+    assertEquals( 9, applicationServer.getQueueCapacity());
```
dimitarndimitrov, line 253: super-nit: Also one here...

dimitarndimitrov (approved):
+1.

PR #313, opened 2022-05-17
MMA-11901: Suppress json parse error if debug config is disabled

srpanwar-confluent:
Jira: https://confluentinc.atlassian.net/browse/MMA-11901

What
Rather than creating a new flag, this change piggybacks on the existing `debug` config to suppress the JSON parsing error message. We replace the actual message with a generic `Error parsing JSON` text.

### DIFF
--- a/core/src/main/java/io/confluent/rest/exceptions/SuppressedJsonParseExceptionMapper.java
+++ b/core/src/main/java/io/confluent/rest/exceptions/SuppressedJsonParseExceptionMapper.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2017 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/confluent/rest/exceptions/SuppressedJsonParseExceptionMapper.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+import javax.ws.rs.ext.ExceptionMapper;
+
+public class SuppressedJsonParseExceptionMapper implements ExceptionMapper<JsonParseException> {
+  public SuppressedJsonParseExceptionMapper() {}
+
+  public Response toResponse(JsonParseException exception) {
+    return Response.status(Status.BAD_REQUEST)
+        .entity("Error parsing JSON")
```
rigelbm, line 29: Can't we at least send exception.getMessage()? The user is losing critical information to understand what went wrong.

linouk23: cc @srpanwar-confluent

PR #315, opened 2022-06-09
KREST-4977 Allow limiting the number of active connections.

dimitarndimitrov:
This change enables configurable on startup connection limiting per server and per connector based on Jetty's `ConnectionLimit` listener (see its [Javadoc](https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/server/ConnectionLimit.html) and [this Jetty issue about its usage](https://github.com/eclipse/jetty.project/issues/2501#issue-319785646)).

That mechanism allows limiting the number of active connections (already opened + currently being opened) connections as with long-running connections, like with Kafka REST's Produce v3 streams, rate limits don't guarantee that we won't have an accumulation of connections past what the server can withstand.

This has been tested locally via the aforementioned Kafka REST Produce v3 streams.
After setting `server.connection.limit=3` and opening 3 Produce v3 streams, a subsequent Get Topics call did not establish a connection until the first opened Produce v3 stream timed out.
In the command output below notice the `Time Spent` column, indicating that this call took slightly less than the 30 second timeout which killed the first Produce v3 stream and opened a space for a new connection:
```
ddimitrov@maconfluent kafka-rest % curl http://localhost:8082/v3/clusters/lkc-kjg39m/topics/ | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   121    0   121    0     0      4      0 --:--:--  0:00:26 --:--:--    27
{
  "kind": "KafkaTopicList",
  "metadata": {
    "self": "http://localhost:8082/v3/clusters/lkc-kjg39m/topics",
    "next": null
  },
  "data": []
}
```

### DIFF
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -440,6 +440,26 @@ public class RestConfig extends AbstractConfig {
           + "Default is false.";
   private static final boolean DOS_FILTER_MANAGED_ATTR_DEFAULT = false;
 
+  private static final String SERVER_CONNECTION_LIMIT = "server.connection.limit";
+  private static final String SERVER_CONNECTION_LIMIT_DOC =
+      "Limits the number of active connections on that server to the configured number. Once that "
+          + "limit is reached further connections will not be accepted until the number of active "
+          + "connections goes below that limit again. Active connections here means all already "
+          + "opened connections plus all connections that are in the process of being accepted. "
+          + "If the limit is set to a non-positive number, no limit is applied. Default is 0.";
+  private static final int SERVER_CONNECTION_LIMIT_DEFAULT = 0;
+
+  // For rest-utils applications connectors correspond to configured listeners. See
+  // ApplicationServer#parseListeners for more details.
+  private static final String CONNECTOR_CONNECTION_LIMIT = "connector.connection.limit";
+  private static final String CONNECTOR_CONNECTION_LIMIT_DOC =
+      "Limits the number of active connections per connector to the configured number. Once that "
+          + "limit is reached further connections will not be accepted until the number of active "
+          + "connections goes below that limit again. Active connections here means all already "
+          + "opened connections plus all connections that are in the process of being accepted. "
+          + "If the limit is set to a non-positive number, no limit is applied. Default is 0.";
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -440,6 +440,26 @@ public class RestConfig extends AbstractConfig {
           + "Default is false.";
   private static final boolean DOS_FILTER_MANAGED_ATTR_DEFAULT = false;
 
+  private static final String SERVER_CONNECTION_LIMIT = "server.connection.limit";
+  private static final String SERVER_CONNECTION_LIMIT_DOC =
+      "Limits the number of active connections on that server to the configured number. Once that "
+          + "limit is reached further connections will not be accepted until the number of active "
+          + "connections goes below that limit again. Active connections here means all already "
+          + "opened connections plus all connections that are in the process of being accepted. "
+          + "If the limit is set to a non-positive number, no limit is applied. Default is 0.";
```
ehumber, line 449: Is this quite right? Looking at the setting code above, there is no limit set if the default of 0 is used too (not just if a negative number is used)

ehumber: Or I guess it technically is correct if 0 is considered non-positive.... But I think that requires a little processing when reading the line that I think people will miss.

ehumber (approved):

srpanwar-confluent:
@dimitarndimitrov the master builds for the last few days are failing. Can you verify this change is not the root cause?https://confluentinc.atlassian.net/browse/MMA-12033

cc @abhijeet2096-confluent @kkeshav-confluent

dimitarndimitrov:
> @dimitarndimitrov the master builds for the last few days are failing. Can you verify this change is not the root cause?https://confluentinc.atlassian.net/browse/MMA-12033
> 
> cc @abhijeet2096-confluent @kkeshav-confluent

Thanks, @srpanwar-confluent - one test from this change is failing on Jenkins. I'll take a look to see if I can find a nice way to fix it, although I feel that I might have to use a crude solution like increasing timeouts. I'll add you as a reviewer to the PR for the fix.

PR #317, opened 2022-06-28
ErrorHandler to suppress the default Jetty behavior of printing full stacktraces for 500s

dragonmushu:
An issue was noticed where any unhandled exceptions thrown by Jetty Handlers display the stack trace of the exception to the user. For internal server errors we should avoid showing stack traces to the user as it poses a security risk. The StackTraceErrorHandler removes stack traces and servlet information before passing response to user and is optional by consumers of the rest-utils library.

### DIFF
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -228,7 +228,7 @@ public class RestConfig extends AbstractConfig {
   @Deprecated
   public static final String SSL_CLIENT_AUTH_CONFIG = "ssl.client.auth";
   protected static final String SSL_CLIENT_AUTH_DOC =
-      "Whether or not to require the https client to authenticate via the server's trust store. " 
+      "Whether or not to require the https client to authenticate via the server's trust store. "
           + "Deprecated; please use " + SSL_CLIENT_AUTHENTICATION_CONFIG + " instead.";
   protected static final boolean SSL_CLIENT_AUTH_DEFAULT = false;
   public static final String SSL_ENABLED_PROTOCOLS_CONFIG = "ssl.enabled.protocols";
@@ -480,6 +480,15 @@ public class RestConfig extends AbstractConfig {
           + "https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt for more information. "
           + "Default is false.";
   protected static final boolean PROXY_PROTOCOL_ENABLED_DEFAULT = false;
+
+  public static final String ERROR_HANDLING_ENABLED_CONFIG = "error.handling.enabled";
+
+  protected static final String ERROR_HANDLING_ENABLED_DOC =
+      "If true, enable overall error handling for any uncaught errors in handlers pipeline. "
+          + "Exceptions usually display stack trace to the client and the error handler will "
+          + "disable the stack trace to the client.";
+
+  protected static final boolean ERROR_HANDLING_ENABLED_DEFAULT = true;
 
   public static ConfigDef baseConfigDef() {
     return baseConfigDef(
@@ -974,6 +983,12 @@ public class RestConfig extends AbstractConfig {
             NOSNIFF_PROTECTION_ENABLED_DEFAULT,
             Importance.LOW,
             NOSNIFF_PROTECTION_ENABLED_DOC
+        ).define(
+            ERROR_HANDLING_ENABLED_CONFIG,
+            Type.BOOLEAN,
+            ERROR_HANDLING_ENABLED_DEFAULT,
+            Importance.LOW,
+            ERROR_HANDLING_ENABLED_DOC
         );
   }
 
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -481,14 +481,13 @@ public class RestConfig extends AbstractConfig {
           + "Default is false.";
   protected static final boolean PROXY_PROTOCOL_ENABLED_DEFAULT = false;
 
-  public static final String ERROR_HANDLING_ENABLED_CONFIG = "error.handling.enabled";
+  public static final String SUPPRESS_STACK_TRACE_IN_RESPONSE = "suppress.stack.trace.response";
 
-  protected static final String ERROR_HANDLING_ENABLED_DOC =
+  protected static final String SUPPRESS_STACK_TRACE_IN_RESPONSE_DOC =
       "If true, enable overall error handling for any uncaught errors in handlers pipeline. "
-          + "Exceptions usually display stack trace to the client and the error handler will "
-          + "disable the stack trace to the client.";
+          + "This ensures that no stack traces are included in responses to clients.";
 
-  protected static final boolean ERROR_HANDLING_ENABLED_DEFAULT = true;
+  protected static final boolean SUPPRESS_STACK_TRACE_IN_RESPONSE_DEFAULT = true;
 
   public static ConfigDef baseConfigDef() {
     return baseConfigDef(
@@ -984,11 +983,11 @@ public class RestConfig extends AbstractConfig {
             Importance.LOW,
             NOSNIFF_PROTECTION_ENABLED_DOC
         ).define(
-            ERROR_HANDLING_ENABLED_CONFIG,
+            SUPPRESS_STACK_TRACE_IN_RESPONSE,
             Type.BOOLEAN,
-            ERROR_HANDLING_ENABLED_DEFAULT,
+            SUPPRESS_STACK_TRACE_IN_RESPONSE_DEFAULT,
             Importance.LOW,
-            ERROR_HANDLING_ENABLED_DOC
+            SUPPRESS_STACK_TRACE_IN_RESPONSE_DOC
         );
   }
 
--- a/core/src/main/java/io/confluent/rest/errorhandlers/NoJettyDefaultStackTraceErrorHandler.java
+++ b/core/src/main/java/io/confluent/rest/errorhandlers/NoJettyDefaultStackTraceErrorHandler.java
@@ -27,7 +27,7 @@ public class NoJettyDefaultStackTraceErrorHandler extends ErrorHandler {
         retrieveErrorMessage(code, message));
   }
 
-  private String retrieveErrorMessage(int code, String message) {
+  protected String retrieveErrorMessage(int code, String message) {
     switch (code) {
       case HttpStatus.INTERNAL_SERVER_ERROR_500:
         return HttpStatus.getMessage(code);
--- a/core/src/main/java/io/confluent/rest/errorhandlers/NoJettyDefaultStackTraceErrorHandler.java
+++ b/core/src/main/java/io/confluent/rest/errorhandlers/NoJettyDefaultStackTraceErrorHandler.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2022 Confluent Inc.
+ */
+
+package io.confluent.rest.errorhandlers;
+
+import java.io.IOException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.eclipse.jetty.http.HttpStatus;
+import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.server.handler.ErrorHandler;
+
+public class NoJettyDefaultStackTraceErrorHandler extends ErrorHandler {
+  private static final String JAVAX_ERROR_EXCEPTION_ATTRIBUTE = "javax.servlet.error.exception";
+
+  public NoJettyDefaultStackTraceErrorHandler() {
+    super();
+    setShowServlet(false);
+  }
+
+  @Override
+  protected void generateAcceptableResponse(Request baseRequest, HttpServletRequest request,
+      HttpServletResponse response, int code, String message) throws IOException {
+    request.setAttribute(JAVAX_ERROR_EXCEPTION_ATTRIBUTE, null);
+    super.generateAcceptableResponse(baseRequest, request, response, code,
+        retrieveErrorMessage(code, message));
+  }
+
+  private String retrieveErrorMessage(int code, String message) {
+    switch (code) {
+      case HttpStatus.INTERNAL_SERVER_ERROR_500:
+        return HttpStatus.getMessage(code);
+      default:
+        return message;
+    }
+  }
+}
--- a/core/src/test/java/io/confluent/rest/ErrorHandlerIntegrationTest.java
+++ b/core/src/test/java/io/confluent/rest/ErrorHandlerIntegrationTest.java
@@ -1,0 +1,117 @@
+package io.confluent.rest;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.confluent.rest.errorhandlers.NoJettyDefaultStackTraceErrorHandler;
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.core.Configurable;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.servlet.ServletContextHandler;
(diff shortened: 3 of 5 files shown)


core/src/test/java/io/confluent/rest/ErrorHandlerIntegrationTest.java
```diff
@@ -0,0 +1,117 @@
...
+    public void setupResources(Configurable<?> config, RestConfig appConfig) {
+      config.register(TestResource.class);
+    }
+  }
+
+  private static class DummyServerHandler extends ServletHandler {
+
+    public void doHandle(String target, Request baseRequest, HttpServletRequest request,
+        HttpServletResponse response) throws IOException, ServletException {
+      throw new RuntimeException(DUMMY_EXCEPTION);
```
milosimpson, line 104: Registering a full on ServletHander is cool.   But there should also be a test that more mimics the "normal" usage, i.e.
* have a specific endpoint that always throws 500, and verify behavior based on config.   Also have an endpoint that doesn't throw 500 and check behavior to prove that this doesn't break normal path things.
* test at the Jetty security layer, using the normal Jetty Servlet Handler.   I.e. configure a special basic auth login that blows up for a specific user, and that this approach covers that as well

milosimpson (commented):
Please to add testing that more closely mimics usage, i.e. I don't think we generally set/overwrite the servlet handler.

core/src/main/java/io/confluent/rest/errorhandlers/NoJettyDefaultStackTraceErrorHandler.java
```diff
@@ -0,0 +1,38 @@
...
+      HttpServletResponse response, int code, String message) throws IOException {
+    request.setAttribute(JAVAX_ERROR_EXCEPTION_ATTRIBUTE, null);
+    super.generateAcceptableResponse(baseRequest, request, response, code,
+        retrieveErrorMessage(code, message));
+  }
+
+  private String retrieveErrorMessage(int code, String message) {
+    switch (code) {
+      case HttpStatus.INTERNAL_SERVER_ERROR_500:
+        return HttpStatus.getMessage(code);
```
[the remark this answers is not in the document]
ehumber: No need, I tested this myself in Kafka-rest - the exception stack is still written to the logs, but not out to the client call :)

ehumber:
Hi @dragonmushu 

Thanks for the PR, it looks good :)

I tested the changes against kafka-rest to see how they worked with this application, and found that I needed to register the errorHandler on the Application, otherwise it's not invoked.

For example

` context.setErrorHandler(new NoJettyDefaultStackTraceErrorHandler());` in Application.java in rest-utils

I think this should be done in rest-utils as part of this PR, rather than for the individual applications as this error suppression something that should be enabled by default, not opted in to by the individual applications that use rest-utils.


To test this I enabled security on kafka REST, and then sent a badly formed auth header

eg admin:admin should be base64 encoded

```
curl -X POST  http://localhost:8082/v3/clusters/8iW5H_i_TAqqG4SFixSgcw/topics/summit/records/ -H 'Content-Type: application/json'  -d '{"records": [{},{}]}' -H 'Authorization: Basic  admin:admin'

```

Without this PR code change I get 

```
<title>Error 500 java.lang.IllegalArgumentException: Illegal base64 character 20</title>
</head>
<body><h2>HTTP ERROR 500 java.lang.IllegalArgumentException: Illegal base64 character 20</h2>
<table>
<tr><th>URI:</th><td>/v3/clusters/8iW5H_i_TAqqG4SFixSgcw/topics/summit/records/</td></tr>
<tr><th>STATUS:</th><td>500</td></tr>
<tr><th>MESSAGE:</th><td>java.lang.IllegalArgumentException: Illegal base64 character 20</td></tr>
<tr><th>CAUSED BY:</th><td>java.lang.IllegalArgumentException: Illegal base64 character 20</td></tr>
</table>
<h3>Caused by:</h3><pre>java.lang.IllegalArgumentException: Illegal base64 character 20
	at java.base/java.util.Base64$Decoder.decode0(Base64.java:746)
	at java.base/java.util.Base64$Decoder.decode(Base64.java:538)
	at java.base/java.util.Base64$Decoder.decode(Base64.java:561)
	at org.eclipse.jetty.security.authentication.BasicAuthenticator.validateRequest(BasicAuthenticator.java:82)
...

```

With this change (plus registering the error handler) I get
```
<title>Error 500 Server Error</title>
</head>
<body><h2>HTTP ERROR 500 Server Error</h2>
<table>
<tr><th>URI:</th><td>/v3/clusters/8iW5H_i_TAqqG4SFixSgcw/topics/summit/records/</td></tr>
<tr><th>STATUS:</th><td>500</td></tr>
<tr><th>MESSAGE:</th><td>Server Error</td></tr>
</table>

```

I think that's not very useful for our end users - it's a user error (and really, jetty shouldn't throw a 500 error here) - but we are going to have support tickets raised against us because there is no indication to the end user of what they are doing wrong, and a 500 looks like it's our bug.


For reference, if I remove the http status mapping code in retrieveErrorMessage I get

```
<title>Error 500 java.lang.IllegalArgumentException: Illegal base64 character 20</title>
</head>
<body><h2>HTTP ERROR 500 java.lang.IllegalArgumentException: Illegal base64 character 20</h2>
<table>
<tr><th>URI:</th><td>/v3/clusters/8iW5H_i_TAqqG4SFixSgcw/topics/summit/records/</td></tr>
<tr><th>STATUS:</th><td>500</td></tr>
<tr><th>MESSAGE:</th><td>java.lang.IllegalArgumentException: Illegal base64 character 20</td></tr>
</table>
```

Could we either

 - in the retrieveErrorMessage() method remove the 500 case mapping (probably the wrong answer, as all other 500 errors will end up with the first line of the stack printed and we don't want that)
 - put in a special case statement specifically looking for java.lang.IllegalArgumentException: Illegal base64 character (or java.lang.* as we don't care if we reveal we are using java) and return the original exception message in the title in this case.
 
 
 And as a side note for my reference, if auth is NOT enabled on Kafka REST, then the error returned is completely different, so Kafka REST needs to take a follow on jira to go look at why that is:
 
 ```
{"error_code":400,"message":"Unrecognized field \"records\" (class io.confluent.kafkarest.entities.v3.AutoValue_ProduceRequest$Builder), not marked as ignorable (6 known properties: \"value\", \"originalSize\", \"partitionId\", \"headers\", \"key\", \"timestamp\"])\n at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 21] (through reference chain: io.confluent.kafkarest.entities.v3.AutoValue_ProduceRequest$Builder[\"records\"])"}

```

**Comment by dragonmushu** [comment]
> * original



> Hi @dragonmushu
> 
> Thanks for the PR, it looks good :)
> 
> I tested the changes against kafka-rest to see how they worked with this application, and found that I needed to register the errorHandler on the Application, otherwise it's not invoked.
> 
> For example
> 
> ` context.setErrorHandler(new NoJettyDefaultStackTraceErrorHandler());` in Application.java in rest-utils
> 
> I think this should be done in rest-utils as part of this PR, rather than for the individual applications as this error suppression something that should be enabled by default, not opted in to by the individual applications that use rest-utils.
> 
> To test this I enabled security on kafka REST, and then sent a badly formed auth header
> 
> eg admin:admin should be base64 encoded
> 
> ```
> curl -X POST  http://localhost:8082/v3/clusters/8iW5H_i_TAqqG4SFixSgcw/topics/summit/records/ -H 'Content-Type: application/json'  -d '{"records": [{},{}]}' -H 'Authorization: Basic  admin:admin'
> ```
> 
> Without this PR code change I get
> 
> ```
> <title>Error 500 java.lang.IllegalArgumentException: Illegal base64 character 20</title>
> </head>
> <body><h2>HTTP ERROR 500 java.lang.IllegalArgumentException: Illegal base64 character 20</h2>
> <table>
> <tr><th>URI:</th><td>/v3/clusters/8iW5H_i_TAqqG4SFixSgcw/topics/summit/records/</td></tr>
> <tr><th>STATUS:</th><td>500</td></tr>
> <tr><th>MESSAGE:</th><td>java.lang.IllegalArgumentException: Illegal base64 character 20</td></tr>
> <tr><th>CAUSED BY:</th><td>java.lang.IllegalArgumentException: Illegal base64 character 20</td></tr>
> </table>
> <h3>Caused by:</h3><pre>java.lang.IllegalArgumentException: Illegal base64 character 20
> 	at java.base/java.util.Base64$Decoder.decode0(Base64.java:746)
> 	at java.base/java.util.Base64$Decoder.decode(Base64.java:538)
> 	at java.base/java.util.Base64$Decoder.decode(Base64.java:561)
> 	at org.eclipse.jetty.security.authentication.BasicAuthenticator.validateRequest(BasicAuthenticator.java:82)
> ...
> ```
> 
> With this change (plus registering the error handler) I get
> 
> ```
> <title>Error 500 Server Error</title>
> </head>
> <body><h2>HTTP ERROR 500 Server Error</h2>
> <table>
> <tr><th>URI:</th><td>/v3/clusters/8iW5H_i_TAqqG4SFixSgcw/topics/summit/records/</td></tr>
> <tr><th>STATUS:</th><td>500</td></tr>
> <tr><th>MESSAGE:</th><td>Server Error</td></tr>
> </table>
> ```
> 
> I think that's not very useful for our end users - it's a user error (and really, jetty shouldn't throw a 500 error here) - but we are going to have support tickets raised against us because there is no indication to the end user of what they are doing wrong, and a 500 looks like it's our bug.
> 
> For reference, if I remove the http status mapping code in retrieveErrorMessage I get
> 
> ```
> <title>Error 500 java.lang.IllegalArgumentException: Illegal base64 character 20</title>
> </head>
> <body><h2>HTTP ERROR 500 java.lang.IllegalArgumentException: Illegal base64 character 20</h2>
> <table>
> <tr><th>URI:</th><td>/v3/clusters/8iW5H_i_TAqqG4SFixSgcw/topics/summit/records/</td></tr>
> <tr><th>STATUS:</th><td>500</td></tr>
> <tr><th>MESSAGE:</th><td>java.lang.IllegalArgumentException: Illegal base64 character 20</td></tr>
> </table>
> ```
> 
> Could we either
> 
> * in the retrieveErrorMessage() method remove the 500 case mapping (probably the wrong answer, as all other 500 errors will end up with the first line of the stack printed and we don't want that)
> * put in a special case statement specifically looking for java.lang.IllegalArgumentException: Illegal base64 character (or java.lang.* as we don't care if we reveal we are using java) and return the original exception message in the title in this case.
> 
> On a side now, I will raise a follow on jira for Kafka REST produce as errors in the stream handling don't seem to go via this jetty exception handler
> 
> eg I get
> 
> ```
> curl -X POST  http://localhost:8082/v3/clusters/8iW5H_i_TAqqG4SFixSgcw/topics/summit/records/ -H 'Content-Type: application/json'  -d '{"records": [{},{}]}' -H 'Authorization: Basic  YWRtaW46YWRtaW4K'
> 
> :"Unrecognized field \"records\" (class io.confluent.kafkarest.entities.v3.AutoValue_ProduceRequest$Builder), not marked as ignorable (6 known properties: \"value\", \"originalSize\", \"partitionId\", \"headers\", \"key\", \"timestamp\"])\n at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 21] (through reference chain: io.confluent.kafkarest.entities.v3.AutoValue_ProduceRequest$Builder[\"records\"])"}
> ```
> 
> even with this fix.
> 
> So this PR only fixes the fill stack trace error reported at the end of https://confluentinc.atlassian.net/jira/software/c/projects/KREST/boards/417?modal=detail&selectedIssue=KREST-6129 not the shorter error.

Hi @ehumber Couple of things from your comment I will address. The first bit of actually registering the error handler in the pipeline: we actually made the conscious decision to be opt-in even though no service really should display stack trace because we wanted to cause minimal disturbance with experience. Also in case if a service already had their own error handler registered we did not want to override the error handler from that service. Couple of services I did see no error handler registered and thus this issue surfaced. But for example, I believe Schema Registry has their own error handler implemented which is very close to this implementation. 

The second point of setting the correct error code. This catch-all error handler was implemented for any issue that surfaced from the handlers that were uncaught (which included the securityhandler). We can not make assumptions of the exception but the intention was to make sure we suppress the stack trace and any extraneous info about the issue. We actually did discuss when we first got this bug that the actual status code thrown in this case when the auth token is malformed should be a 401 but the first step we wanted to do was register a catch-all error handler in the pipeline that scraped the stack trace and the actual java exceptions. After that is done we were planning to  into our security handler pipeline and make sure we catch the exception thrown by BasicAuth and throw the appropriate 401 for malformed token. Let me know if that answers your questions.

**Comment by ehumber** [comment]
Hi @dragonmushu 

Yup, that makes sense 

- how are you going to let all the relevant teams know they need to register the exception handler to suppress the stack traces?

- For throwing the 401 for the malformed token, can you let me know the jira for tracking it please, so I can add a link to our jira too?

**Comment by ehumber** [comment]
For reference the kafka-rest jira that prompted this is https://confluentinc.atlassian.net/browse/KREST-6129

**Review by ehumber — APPROVED** [review]

File: core/src/test/java/io/confluent/rest/ErrorHandlerIntegrationTest.java (hunk 1 above)
Comment (dragonmushu) [reply core/src/test/java/io/confluent/rest/ErrorHandlerIntegrationTest.java:121]: added test that mimicked error with security handler

**Comment by dragonmushu** [comment]
> Hi @dragonmushu
> 
> Yup, that makes sense
> 
> * how are you going to let all the relevant teams know they need to register the exception handler to suppress the stack traces?
> * For throwing the 401 for the malformed token, can you let me know the jira for tracking it please, so I can add a link to our jira too?

Hey @ehumber so I took your comment into consideration and decided to actually set the error handler on by default. Please do another pass of the pr. Just used a config flag to be able to enable/disable to allow teams to disable if they must. Also for your second comment: https://confluentinc.atlassian.net/browse/CIAM-2191 . I just created this ticket and where the problem exists but we have not implemented change.

File: core/src/test/java/io/confluent/rest/ErrorHandlerIntegrationTest.java [hunk 3]
```diff
@@ -34,9 +37,14 @@ public void tearDown() throws Exception {
...
 
   @Test
-  public void unhandledServerExceptionDisplaysStackTrace() throws Exception {
-    TestApplication application = new TestApplication(new RestConfig(RestConfig.baseConfigDef()),
-        false);
+  public void unhandledServerExceptionDisplaysStackTraceForInvalidAuthentication()
+      throws Exception {
+
+    Properties props = new Properties();
+    props.setProperty(RestConfig.ERROR_HANDLING_ENABLED_CONFIG, "false");
```
ehumber, line 44: I'm wondering if this is the right name for the property.

What the config really does is SUPPRESS_STACK_TRACE_IN_RESPONSE so something similar to that might be more appropriate. What do you think?

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -481,6 +481,15 @@ public class RestConfig extends AbstractConfig {
           + "Default is false.";
   protected static final boolean PROXY_PROTOCOL_ENABLED_DEFAULT = false;
 
+  public static final String ERROR_HANDLING_ENABLED_CONFIG = "error.handling.enabled";
```
ehumber, line 484: I'm wondering if this is the right name for the property.

What the config really does is SUPPRESS_STACK_TRACE_IN_RESPONSE so something similar to that might be more appropriate. What do you think?

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -481,6 +481,15 @@ public class RestConfig extends AbstractConfig {
...
+
+  protected static final String ERROR_HANDLING_ENABLED_DOC =
+      "If true, enable overall error handling for any uncaught errors in handlers pipeline. "
+          + "Exceptions usually display stack trace to the client and the error handler will "
+          + "disable the stack trace to the client.";
```
ehumber, line 489: If true, enable overall error handling for any uncaught errors in handlers pipeline. 

This ensures that no stack traces are included in responses sent to clients.

ehumber:
> 

For the enabling by default, you mentioned that some components already have custom error handlers, eg Schema Registry.  I'm assuming their error handler setting on the context will override this one, so it's still in effect?  We don't want to accidentally prevent their error handling from happening with this change.

core/src/main/java/io/confluent/rest/RestConfig.java
dragonmushu: yep i think that is better. I was trying to play with both stack trace and error handling and ended up shortening to error handling. I will add the one you mentioned.

dragonmushu:
> > 
> 
> For the enabling by default, you mentioned that some components already have custom error handlers, eg Schema Registry. I'm assuming their error handler setting on the context will override this one, so it's still in effect? We don't want to accidentally prevent their error handling from happening with this change.

yep verified that overwriting the error handler from Application will work

milosimpson (approved):
LGTM

PR #341, opened 2022-07-19
DGS-4220: fix request tag based metrics

ericwush:
This PR tries to fix the bug that a sensor created for a method + performance annotation pair is shared across the entire application, as opposed to each request tag set having its own copy of sensor. It is causing a problem when request tags are used. A sensor is either created or fetched depending on whether the name exists. For example, 
1. `methodX.request-size` sensor is created on application startup, let's call it sensorA
2. a request w/ tag1 comes in, sensorA is returned because the same sensor name is provided
3. a request w/ tag2 comes in, sensorA is again returned

However, each tag really should have its own sensor.

The fix sorts request tags to make sure the same sensor is used for the same set of request tags. Tags in `MetricName` is also sorted for `JmxReporter` to create consistent MBean names. I understand this is an implicit side effect, a more proper place to ensure the consistency is within `JmxReporter`.

This PR also fixes a bug that the error sensor is overwritten, so error rates are not recorded.

### DIFF
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -31,7 +31,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.reflect.Method;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.SortedMap;
@@ -54,6 +53,8 @@ import org.apache.kafka.common.utils.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static java.util.Collections.emptyMap;
+
 /**
  * Jersey ResourceMethodApplicationListener that records metrics for each endpoint by listening
  * for method start and finish events. It reports some common metrics for each such as rate and
@@ -66,7 +67,7 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
   public static final String REQUEST_TAGS_PROP_KEY = "_request_tags";
 
   protected static final String HTTP_STATUS_CODE_TAG = "http_status_code";
-  private static final String[] HTTP_STATUS_CODE_TEXT = {
+  protected static final String[] HTTP_STATUS_CODE_TEXT = {
       "unknown", "1xx", "2xx", "3xx", "4xx", "5xx"};
   private static final int PERCENTILE_NUM_BUCKETS = 200;
   private static final double PERCENTILE_MAX_LATENCY_IN_MS = TimeUnit.SECONDS.toMillis(10);
@@ -82,7 +83,7 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
     super();
     this.metrics = metrics;
     this.metricGrpPrefix = metricGrpPrefix;
-    this.metricTags = (metricTags != null) ? metricTags : Collections.emptyMap();
+    this.metricTags = (metricTags != null) ? metricTags : emptyMap();
     this.time = time;
   }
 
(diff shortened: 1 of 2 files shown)


ehumber:
@codyaray Is this something your team can look at reviewing?

ericwush:
@rayokota should be able to help with the review.

BTW, I have done a code search, looks like only our team is using request tags by setting `REQUEST_TAGS_PROP_KEY`. Just double check @ehumber your team doesn't use it, otherwise, the according alerts might need to be adjusted.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -139,16 +142,12 @@ public MethodMetrics metrics() {
...
-      allTags.putAll(context.metricTags);
-      allTags.putAll(requestTags);
-      return new MethodMetrics(context.method, context.performanceMetric,
-          context.metrics, context.metricGrpPrefix, allTags);
+      // The key will also be used to identify a unique sensor,
+      // so we want to pass the sorted tags to MethodMetrics
+      SortedMap<String, String> key = new TreeMap<>(requestTags);
+      requestMetrics.putIfAbsent(key, new MethodMetrics(context.method, context.performanceMetric,
+          context.metrics, context.metricGrpPrefix, context.metricTags, key));
+      return requestMetrics.get(key);
```
rayokota, line 150: Can this be replaced with `computeIfAbsent`?

rayokota (approved):
One comment, otherwise LGTM

ehumber:
> @rayokota should be able to help with the review.
> 
> BTW, I have done a code search, looks like only our team is using request tags by setting `REQUEST_TAGS_PROP_KEY`. Just double check @ehumber your team doesn't use it, otherwise, the according alerts might need to be adjusted.

I've looked, I can't see any usage of `REQUEST_TAGS_PROP_KEY` in the kafka-rest or ce-kafka-rest code.

ericwush:
Thanks @ehumber. Seems I need the code owner's approval to merge. Otherwise, could you merge it for me?

codyaray:
cc @rigelbm

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -184,57 +183,65 @@ private static class MethodMetrics {
...
+                         String metricGrpPrefix, Map<String, String> metricTags,
+                         Map<String, String> requestTags) {
       String metricGrpName = metricGrpPrefix + "-metrics";
+      // The tags will be used to generate MBean names if JmxReporter is used,
+      // sort to get consistent names
+      Map<String, String> allTags = new TreeMap<>(metricTags);
+      allTags.putAll(requestTags);
 
-      this.requestSizeSensor = metrics.sensor(getName(method, annotation, "request-size"));
+      this.requestSizeSensor = metrics.sensor(
```
ehumber, line 194: When making a similar change in kafka-rest, I was asked to ensure the sensors expire if they are unused, otherwise we will end up with a sensor leak.

1 hour was the suggested duration.

This didn't matter so much when one sensor was reused, but now there are multiple sensors I think it is important.

ehumber (approved):

ehumber:
@ericwush Approved but with one comment that needs addressing before you merge.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
ericwush: Good point, addressed, thanks!

ericwush:
Thanks @ehumber for the review. As discussed, I've addressed the feedback and also included the bug fix for error sensor.

ehumber:
LGTM

PR #349, opened 2022-08-09
Convert the response to json format and hidding the details of the co…

jliunyu:
…de path

For the V2 version, when produce a message, if the request body is an invalid json, we will meet this two problems.
1 ) The response is not a json message.
2) It contains two much information: for example
```
Unexpected character (''' (code 39)): was expecting double-quote to start field name
 at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 7, column: 7] (through reference chain: io.confluent.kafkarest.entities.v2.ProduceRequest["records"]->java.util.ArrayList[1])
```

The reasons the above problems are:
1) https://github.com/confluentinc/rest-utils/blob/master/core/src/main/java/io/confluent/rest/exceptions/JsonMappingExceptionMapper.java doesn't trigger when throws JsonMappingException, so the response doesn't convert to json.
2) Exception message is not been handled from JsonMappingException, all the details returned to users directly.

How to fix:
1) There are two types Jackson related exception: JsonParseException and the one we met JsonMappingException, we need to register both mappers here https://github.com/confluentinc/rest-utils/blob/master/core/src/main/java/io/confluent/rest/Application.java#L631, but we only registered JsonParseExceptionMapper
2) According to https://stackoverflow.com/a/58287443,
Starting with Jersey 2.26 (1, 2) it should be enough to annotate the custom exception mapper with a sufficiently high Priority
We are using version 2.34, so I added @Priority(1) to https://github.com/confluentinc/rest-utils/blob/master/core/src/main/java/io/confluent/rest/exceptions/JsonMappingExceptionMapper.java#L31
3) Remove exception messages from https://github.com/confluentinc/rest-utils/blob/master/core/src/main/java/io/confluent/rest/exceptions/JsonMappingExceptionMapper.java#L35

Manually test:
```
jliu@C02DM32YMD6T kafka-rest % curl -X POST \
     -H "Content-Type: application/vnd.kafka.json.v2+json" \
     -H "Accept: application/vnd.kafka.v2+json" \
     --data '{"records":[{"key":"abc","value":"alarm clock"},{"key":"htanaka","value":"batteries"},{"key":"awalther","value":"bookshelves"}, {
      "value": ["foo", "bar"],
      "partition": 1
    ]}' \
     "http://localhost:8082/topics/tst"
{"error_code":400,"message":"Unexpected close marker ']': expected '}'"}% 
```

Jenkins failed for 
```19:21:45  [INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.432 s - in io.confluent.rest.CsrfHandlingTest
19:21:45  [INFO] 
19:21:45  [INFO] Results:
19:21:45  [INFO] 
19:21:45  [ERROR] Failures: 
19:21:45  [ERROR]   MetricsResourceMethodApplicationListenerIntegrationTest.testSuccessMetrics:112->assertMetric:325 Actual: 9.0 ==> expected: <10.0> but was: <9.0>
19:21:45  [INFO] 
19:21:45  [ERROR] Tests run: 152, Failures: 1, Errors: 0, Skipped: 2
```
Looking why it failed

### DIFF
--- a/core/src/test/java/io/confluent/rest/JsonMappingExceptionMapperTest.java
+++ b/core/src/test/java/io/confluent/rest/JsonMappingExceptionMapperTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Confluent Inc.
+ * Copyright 2021 - 2022 Confluent Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package io.confluent.rest;
 
+import com.fasterxml.jackson.core.JsonLocation;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.confluent.rest.entities.ErrorMessage;
@@ -53,6 +54,18 @@ public class JsonMappingExceptionMapperTest {
     } catch (Exception e) {
       fail("A JsonMappingException is expected.");
     }
+  }
+
+  @Test
+  public void testJsonMappingExceptionRemoveDetailsFromMessage() {
+      JsonMappingException jsonMappingException = new JsonMappingException("Json mapping error (for Object starting at:", JsonLocation.NA);
+      jsonMappingException.prependPath(new JsonMappingException.Reference("path"), 0);
+
+      Response response = mapper.toResponse(jsonMappingException);
+      assertEquals(400, response.getStatus());
+      ErrorMessage out = (ErrorMessage)response.getEntity();
+      assertEquals(400, out.getErrorCode());
+      assertEquals("Json mapping error", out.getMessage());
   }
 
   class User {
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -629,6 +629,7 @@ public abstract class Application<T extends RestConfig> {
     config.register(jsonProvider);
     if (registerExceptionMapper) {
       config.register(JsonParseExceptionMapper.class);
+      config.register(JsonMappingExceptionMapper.class);
     }
   }
 
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Confluent Inc.
+ * Copyright 2014 - 2022 Confluent Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


ehumber:
Looks good to me Jing - swap the PR over to be Ready for review once you've added the unit test

core/src/test/java/io/confluent/rest/JsonMappingExceptionMapperTest.java
```diff
@@ -55,6 +55,22 @@ public void testJsonMappingException() {
...
+  public void testJsonMappingExceptionWithInvalidJson() {
+    try {
+      String json = "{\"name\":{'key:value}}";
+      ObjectMapper mapper = new ObjectMapper();
+      mapper.reader().forType(User.class).readValue(json);
+    } catch (JsonMappingException e) {
+      Response response = mapper.toResponse(e);
+      assertEquals(400, response.getStatus());
+      ErrorMessage out = (ErrorMessage)response.getEntity();
+      assertEquals(400, out.getErrorCode());
```
ehumber, line 68: Can you check that the error message is as expected too please, not just the code.

ehumber (approved):

jliunyu:
If the exception is `JsonParseException`, we got the similar problem

Before my fix:
```
curl -X POST \
     -H "Content-Type: application/vnd.kafka.json.v2+json" \
     -H "Accept: application/vnd.kafka.v2+json" \
     --data '{"name": {key:value}}' \
     "http://localhost:8082/topics/tst"
Unexpected character ('k' (code 107)): was expecting double-quote to start field name
 at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 12]%  
```
I added a new exception mapper called `JsonParseExceptionMapper` and unit test for it, it's similar with `JsonMappingExceptionMapper`, 

After my fix:
```
curl -X POST \
     -H "Content-Type: application/vnd.kafka.json.v2+json" \
     -H "Accept: application/vnd.kafka.v2+json" \
     --data '{"name": {key:value}}' \
     "http://localhost:8082/topics/tst"
{"error_code":400,"message":"Unexpected character ('k' (code 107)): was expecting double-quote to start field name"}%  
```

rigelbm (approved):
LGTM

Thanks @jliunyu!!!

rigelbm:
This should be merged to 5.4.x, not master. After merging to 5.4.x, you can use https://jenkins.confluent.io/job/tools-autotasks/job/tools-autotasks/job/apply-pint-merge/build to merge it up.

jliunyu:
Discard this one because the issue is fixed by https://github.com/confluentinc/rest-utils/pull/350

PR #357, opened 2022-10-13
APIF-2959: Prepare rest-utils API for per-listener SSL configs.

rigelbm:
This PR prepares the rest-utils API for per-listener SSL configs. Changes:

 * [new] `SslConfig`: This class encapsulates the many SSL params in rest-utils.
 * [new] `SslFactory`: `ApplicationServer.createSslContextFactory` method extracted out.
 * Moved all the listener parsing logic from `ApplicationServer` to `RestConfig`.
 * [new] `RestConfig.getBaseSslConfig()`: Returns the unprefixed (base) SSL configs.
 * [new] `RestConfig.getSslConfigs()`: Returns the per-listener SSL configs.

Note that this PR does not implements per-listener SSL config. It only introduces the APIs, so that downstream applications can adapt before the changes. As such, all the SSL configs returned are the same for all listeners.

### DIFF
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -16,26 +16,33 @@
 
 package io.confluent.rest;
 
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
+import static java.util.Collections.unmodifiableList;
+import static org.apache.kafka.clients.CommonClientConfigs.METRICS_CONTEXT_PREFIX;
+
 import io.confluent.rest.extension.ResourceExtension;
 import io.confluent.rest.metrics.RestMetricsContext;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.time.Duration;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Collectors;
+import javax.ws.rs.core.UriBuilder;
+import javax.ws.rs.core.UriBuilderException;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Importance;
 import org.apache.kafka.common.utils.Time;
-
-import static java.util.Collections.emptyList;
-import static org.apache.kafka.clients.CommonClientConfigs.METRICS_CONTEXT_PREFIX;
 
 public class RestConfig extends AbstractConfig {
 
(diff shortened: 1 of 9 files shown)


rigelbm:
The downstream failure in `secret-registry` is unrelated to these changes. Some change in Kafka broke `secret-registry` master.

ehumber:
@rigelbm Yes I can take a look at this change as you requested, but before I do that, can you find someone in your team who is more familiar with the per listener SSL requirement than I am and get them to review the concept of the change (if not the rest-utils/java specifics). 

It would be a good opportunity to get someone outside the kafka-rest team familiar with rest-utils - maybe have an interactive review session and talk them through the changes, and why you've done what you've done.

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -1139,6 +1168,72 @@ public final Map<String, String> getMap(String propertyName) {
...
+  }
+
+  private static NamedURI constructNamedURI(
+      String listener, Map<String,String> listenerProtocolMap, List<String> supportedSchemes) {
+    URI uri;
+    try {
+      uri = new URI(listener);
+    } catch (URISyntaxException e) {
+      throw new ConfigException(
+          "Listener '" + listener + "' is not a valid URI.");
```
ivanhyf, line 1212: maybe carry the message from URISyntaxException to the config exception for better debuggability.

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -1139,6 +1168,72 @@ public final Map<String, String> getMap(String propertyName) {
...
+    List<NamedURI> namedUris =
+        uris.stream().filter(uri -> uri.getName() != null).collect(Collectors.toList());
+    List<NamedURI> unnamedUris =
+        uris.stream().filter(uri -> uri.getName() == null).collect(Collectors.toList());
+
+    if (namedUris.stream().map(a -> a.getName()).distinct().count() != namedUris.size()) {
+      throw new ConfigException(
+          "More than one listener was specified with same name. Listener names must be unique.");
+    }
+    if (namedUris.isEmpty() && unnamedUris.isEmpty()) {
```
ivanhyf, line 1197: this should be same as uris.isEmpty.()
ok to keep this as is since this is code refactoring.

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -1139,6 +1168,72 @@ public final Map<String, String> getMap(String propertyName) {
...
+    if (listeners.isEmpty() || listeners.get(0).isEmpty()) {
+      listeners = singletonList(defaultScheme + "://0.0.0.0:" + deprecatedPort);
+    }
+
+    List<NamedURI> uris = listeners.stream()
+        .map(listener -> constructNamedURI(listener, listenerProtocolMap, supportedSchemes))
+        .collect(Collectors.toList());
+    List<NamedURI> namedUris =
+        uris.stream().filter(uri -> uri.getName() != null).collect(Collectors.toList());
+    List<NamedURI> unnamedUris =
```
ivanhyf, line 1190: We may not need to construct this list.

ivanhyf (approved):
Some minor comments for the old code.

core/src/main/java/io/confluent/rest/RestConfig.java
rigelbm: Done.

core/src/main/java/io/confluent/rest/RestConfig.java
rigelbm: Done.

core/src/main/java/io/confluent/rest/RestConfig.java
rigelbm: Done.

PR #358, opened 2022-10-28
Use LINEAR instead of CONSTANT bucket sizing for latency percentiles.

cjohnson-confluent:
This will give greater precision when latency is closer to zero
  (and less precision when it's further). With 200 buckets to cover
  10 seconds, CONSTANT sizing gave no precision below 50 ms.

Ref: https://github.com/confluentinc/ce-kafka/blob/5f154af1ffc138ac5ab7b4de1829e3c247f32015/clients/src/main/java/org/apache/kafka/common/metrics/stats/Histogram.java#L164-L168


cjohnson-confluent:
Build failure is unrelated.

ivanhyf (approved):
cc: @rigelbm

PR #359, opened 2022-10-31
CIAM-2346 Searchable Log Format

cyrusv:
Why
----
Cloud RBAC team needs better searchability of our logs. Currently, in Kibana, there's no structured way to search for HTTP-500. If you search for the freetext "500", you also get results where "500" is anywhere in the freetext. Same with latency. We also got paged for latency in DD, and we wanted to search our request logs by latency, but could not.

Ideally, we would use a JSON format (CIAM-735). I explored this, but it seems like a lot of work (we will start taking on at some point this quarter. See "Alternatives" section for more). 

We just want a low-hanging-fruit way to unblock ourselves on these searches.

How
----
Simply extended the log format we have now.

I do worry a little that our log format, since it conforms to some standard, is considered an API for our on-prem users. Hopefully someone can tell me I'm being too paranoid, and it if they are programmatically parsing these log formats they will be fine. I guess I could make it configurable?

Alternatives
----
Configurable log format, and provide a nice JSON format. I was hopeful I could do this, but it is not fun to escape characters that Jetty might put into our JSON. I think we will have to write (or find in open source) an implementation of `RequestLog` that knows how to format and escape.JSON strings. More work than I wanted to take on.


ehumber (approved):

ehumber:
@rigelbm should probably look at this too

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -134,7 +134,11 @@ public Application(T config, String path, String listenerName) {
     logWriter.setLoggerName(config.getString(RestConfig.REQUEST_LOGGER_NAME_CONFIG));
 
     // %{ms}T logs request time in milliseconds
-    requestLog = new CustomRequestLog(logWriter, CustomRequestLog.EXTENDED_NCSA_FORMAT + " %{ms}T");
+    requestLog = new CustomRequestLog(logWriter, requestLogFormat());
+  }
+
+  protected String requestLogFormat() {
+    return CustomRequestLog.EXTENDED_NCSA_FORMAT + " %{ms}T";
```
rigelbm, line 141: LGTM....

But be careful how you override this. The request log format is part of the public "API" of the application, as in, there are people out there that depend on it (on on-prem).

rigelbm (approved):

core/src/main/java/io/confluent/rest/Application.java
cyrusv: Thanks for the reminder -- we plan to only override this in our cloud deployment

PR #361, opened 2022-11-03
Add ApplicationServer.getSslContextFactories method.

rigelbm:
And use it to configure the per-listener SSL factories, even if they are all the same.

This is needed so that https://github.com/confluentinc/blueway/pull/2445 works without changing behaviour on C3 backend.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -225,6 +225,7 @@ public abstract class Application<T extends RestConfig> {
   /**
    * expose SslContextFactory
    */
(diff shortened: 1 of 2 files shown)


rigelbm:
No Idea why the Secret Scan is marked as red, if in Jenkins all steps are green:

```
19:44:37  time="2022-11-03T19:44:37Z" level=info msg="scan time: 347 milliseconds 52 microseconds"
19:44:37  time="2022-11-03T19:44:37Z" level=info msg="No leaks found"
```

ivanhyf (approved):

PR #364, opened 2022-12-09
KREST-8429 alert on 429s

ehumber:
At the moment, 429 errors don't always trigger metrics to be written.

And even then, they are only logged as 4xx errors, not specifically as 429s.

This PR ensures that 429 responses always log a metric, and that the metrics are logged individually.

I think this code change is useful to more than just kafka rest, so I've done this logger override in rest-utils (it would be possible to do it in kafka-rest too if necessary)

429s caused by the jetty dos filter were not logged because the dos filtering occurs before the MetricsResourceMethodApplicationListener (which does metrics for all the endpoints) is called.

I got around this by highjacking the CustomRequestLog (which always fires, and logs the response status) and added metrics reporting to this.

The side benefit is that we can now pretty easily add request/response bodies to error case logging (not done in this PR, but we should do this soon if possible).

We also only logged any 4XX metric under the 4XX tag, so it wasn't possible to alert on just 429s.

This is probably the only 4XX metric we do want to pull out separately, as it can impact other users eg with a noisy neighbour

Existing metrics can be seen by exec-ing into a kafka pod and running `curl localhost:7778` (note these do not include the tenant id)
eg:

```
kafka_rest_jersey_metrics_request_size_max NaN
kafka_rest_jersey_metrics_request_count 0.0
kafka_rest_jersey_metrics_request_latency_95 NaN
kafka_rest_jersey_metrics_request_latency_99 NaN
kafka_rest_jersey_metrics_request_total 0.0
kafka_rest_jersey_metrics_request_error_total{http_status_code="5xx",} 0.0
kafka_rest_jersey_metrics_request_error_total{http_status_code="2xx",} 0.0
kafka_rest_jersey_metrics_request_error_total{http_status_code="1xx",} 0.0
kafka_rest_jersey_metrics_request_error_total{http_status_code="4xx",} 0.0
kafka_rest_jersey_metrics_request_error_total{http_status_code="3xx",} 0.0
```
and produce metrics which look like (note these include the tenant id)
```
kafka_rest_produce_api_metrics_record_error_rate{tenant="lkc-exve7g9",} 0.0
kafka_rest_produce_api_metrics_record_rate_limited_rate{tenant="lkc-exve7g9",} 0.0 <-------- NOTE THAT THIS LOGS METRICS FOR PRODUCE REQUEST RATE LIMITER 429s, as they are returned as 200 success but with a sub-code that gives the result of theproduce
kafka_rest_produce_api_metrics_request_byte_rate{tenant="lkc-exve7g9",} 10.219242616416018
kafka_rest_produce_api_metrics_request_latency_avg{tenant="lkc-exve7g9",} 29.0
kafka_rest_produce_api_metrics_response_rate{tenant="lkc-exve7g9",} 0.2178926152561146
```

This PR adds
```
kafka_rest_jersey_metrics_request_error_count{http_status_code="429",} 0.0
kafka_rest_jersey_metrics_request_error_rate{http_status_code="429",} 0.0
kafka_rest_jersey_metrics_request_error_total{http_status_code="429",} 0.0
```

I decided not to include the LKC tag here so we don't hugely add to our datadog bill (we'd need to do this in kafka-rest if we did want them added)

Tested in a standalone kafka-rest and also in CPD.


pom.xml
```diff
@@ -256,6 +256,16 @@
                 <artifactId>junit-jupiter-api</artifactId>
                 <version>${junit.jupiter.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.mockito</groupId>
+                <artifactId>mockito-core</artifactId>
+                <version>3.11.0</version>
```
trnguyencflt, line 262: we have mockito version defined in common pom, should we use that instead
`<mockito.version>4.6.1</mockito.version>`

core/src/main/java/io/confluent/rest/RestCustomRequestLog.java
```diff
@@ -0,0 +1,85 @@
...
+      String sensorTags =
+          instanceMetricsTags.keySet().stream()
+              .map(key -> ":" + instanceMetricsTags.get(key))
+              .collect(Collectors.joining());
+      String sensorName = sensorNamePrefix + ":request-errors" + sensorTags;
+      fourTwoNineSensor = metrics.sensor(sensorName,
+          null, SENSOR_EXPIRY_SECONDS, RecordingLevel.INFO, (Sensor[]) null);
+
+      fourTwoNineSensor.add(getMetricName(metrics, "request-error-rate",
+          "The average number of requests per second that resulted in 429 error responses",
```
trnguyencflt, line 60: 429 error responses -> 429 HTTP error responses, to be the same as the other two descriptions.

core/src/main/java/io/confluent/rest/RestCustomRequestLog.java
```diff
@@ -0,0 +1,85 @@
...
+import org.eclipse.jetty.server.CustomRequestLog;
+import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.server.Response;
+
+public class RestCustomRequestLog extends CustomRequestLog {
+
+  private static final long SENSOR_EXPIRY_SECONDS = TimeUnit.HOURS.toSeconds(1);
+  private static final String GROUP_NAME = "jersey-metrics";
+
+  Sensor fourTwoNineSensor = null;
```
trnguyencflt, line 41: should this sensor is private? (and maybe final as well)

trnguyencflt (commented):
Only two minor comments on visibility of fourTwoNineSensor and version of mockito in pom. Otherwise look good to me.

pom.xml
```diff
@@ -256,6 +256,16 @@
...
             </dependency>
+            <dependency>
+                <groupId>org.mockito</groupId>
+                <artifactId>mockito-core</artifactId>
+                <version>${mockito.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.mockito</groupId>
+                <artifactId>mockito-inline</artifactId>
+                <version>3.11.0</version>
```
trnguyencflt, line 267: this also should be with ${mockito.version}

trnguyencflt:
@ehumber only one comment left with mockito-inline version (should be ${mockito.version}), pre-approved.

trnguyencflt (approved):

jliunyu (approved):

core/src/main/java/io/confluent/rest/RestCustomRequestLog.java
```diff
@@ -0,0 +1,85 @@
...
+  private static final long SENSOR_EXPIRY_SECONDS = TimeUnit.HOURS.toSeconds(1);
+  private static final String GROUP_NAME = "jersey-metrics";
+
+  private Sensor fourTwoNineSensor = null;
+
+  public RestCustomRequestLog(final Writer writer, final String formatString, Metrics metrics,
+      Map<String, String> metricTags, String jmxPrefix) {
+    super(writer, formatString);
+
+    if (metrics != null) {
```
msn-tldr, line 47: Another approach to add custom-metrics at the Jetty-layer, would be add another jetty-handler much like RequestLogHandler.java (used in [Application.java](https://github.com/confluentinc/rest-utils/blob/2e8e92703a2d386fa75bf43ccfd677c85b2ffc44/core/src/main/java/io/confluent/rest/Application.java#L376)). This would avoid _extending_ CustomRequestLog for non-logging related code.

What do you think?

PS - Since this PR is ready, it might be worth following up with another PR. OR Add a Jira tracker & a comment where in next person who _extends_ RestCustomRequestLog for non-logging related logic, can add the handler.

pom.xml
ehumber: Ooops 😊

core/src/main/java/io/confluent/rest/RestCustomRequestLog.java
ehumber: That's a good call Mayank, thanks for the idea.  We shouldn't mix functions. 

Code has been updated.

core/src/test/java/io/confluent/rest/Rest429CustomResponseHandlerTest.java
```diff
@@ -0,0 +1,225 @@
...
+    when(response2.getStatus()).thenReturn(429);
+    when(request2.getRequestURL()).thenReturn(new StringBuffer("http://lkc-efg567.v4.network.address/clusters/lkc-efg567/topics"));
+    when(request2.getRemoteHost()).thenReturn("remoteHost");
+    when(response2.getCommittedMetaData()).thenReturn(metaDataResponse2);
+    when(metaDataResponse2.getStatus()).thenReturn(429);
+    when(response2.getHttpChannel()).thenReturn(httpChannel2);
+    when(httpChannel2.getBytesWritten()).thenReturn(12345L);
+
+    //send calls
+    Rest429CustomResponseHandler customRequestLog = new Rest429CustomResponseHandler(metrics, Collections.emptyMap(),
```
msn-tldr, line 154: nit: customRequestLog -> rest429CustomResponseHandler

core/src/main/java/io/confluent/rest/RestCustomRequestLog.java
msn-tldr: Gr8, tx! Pointed a minor nit, otherwise good to go.

PR #365, opened 2023-01-07
Adapt Test to handle Exception class changing its parent

milosimpson:
- in KAFKA-14417 ConcurrentTransactionsException was changed from an ApiException to be a RetriableException (which is itself an ApiException)
 - Doing if/else logic based on instanceof check so the test can handle the ConcurrentTransactionsException being of either heritage as Jenkins sees the updated class but my local build only sees the old one

### DIFF
--- a/core/src/test/java/io/confluent/rest/KafkaExceptionMapperTest.java
+++ b/core/src/test/java/io/confluent/rest/KafkaExceptionMapperTest.java
@@ -33,6 +33,7 @@ import org.apache.kafka.common.errors.InvalidRequestException;
 import org.apache.kafka.common.errors.NotCoordinatorException;
 import org.apache.kafka.common.errors.NotEnoughReplicasException;
 import org.apache.kafka.common.errors.PolicyViolationException;
+import org.apache.kafka.common.errors.RetriableException;
 import org.apache.kafka.common.errors.SaslAuthenticationException;
 import org.apache.kafka.common.errors.SecurityDisabledException;
 import org.apache.kafka.common.errors.TimeoutException;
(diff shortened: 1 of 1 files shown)


milosimpson:
The actual PR where the ExceptionClass was changed in Dec 6th.
https://github.com/apache/kafka/pull/12915/files#diff-1746d0b93646b9a55a932b23f4052734815876066003ab45cb237b4a0379a309

omkreddy (approved):
LGTM

milosimpson:
Need a code owner to merge it.   There are two overlapping issues.
1) the AK change to ConcurrentTransactionsException
2) some CVE Scanning issue that causes build failure, which has been addressed in https://github.com/confluentinc/common/pull/496

rest-utils master build can get past the CVE, but fails a test due to ConcurrentTransactionsException
This PR should pass the ConcurrentTransactionsException, but fails the CVE check first.

Even tho this PR branch is up to date with rest-utils master except for this change, the PR builder still is pulling the old CVE logic.

The pressing thing is that rest-utils master has been failing for 20 days due to the ConcurrentTransactionsException change, which is preventing downstream projects from picking up newer upstream Kafka dependencie.

core/src/test/java/io/confluent/rest/KafkaExceptionMapperTest.java
```diff
@@ -133,8 +134,13 @@ public void testKafkaExceptions() {
     //test couple of kafka exception
     verifyMapperResponse(new CommitFailedException(), Status.INTERNAL_SERVER_ERROR,
         KAFKA_ERROR_ERROR_CODE);
-    verifyMapperResponse(new ConcurrentTransactionsException("some message"), Status.INTERNAL_SERVER_ERROR,
-        KAFKA_ERROR_ERROR_CODE);
+
+    Exception cte = new ConcurrentTransactionsException("some message");
+    if(cte instanceof RetriableException) {
```
ijuma, line 139: Can we add a comment explaining why we do this? And nit: there should be a space after `if`.

milosimpson: Added context to the change

ijuma (approved):
LGTM

PR #366, opened 2023-01-08
Adapt test to the new ConcurrentTransactionsException heritage

milosimpson:
- now that the change from KAFKA-14417 ConcurrentTransactionsException 
   has propagated thru the system


ijuma (approved):
LGTM if the tests pass.

PR #368, opened 2023-01-17
KREST-8707 fix Rest429CustomResponseHandler metrics for all applications

trnguyencflt:
This PR
- removes Rest429CustomResponseHandler for 429 metrics, and 
- uses MetricsResourceMethodApplicationListener for 429 metrics.

The motivation is that:
- MetricsResourceMethodApplicationListener has done metrics correctly for 1xx,2xx,..., so it is natural to put 429 here
- Rest429CustomResponseHandler extends RequestLogHandler, which by default bind to all available applications, for which we have at least two in CC (external and internal), that causes the metrics to be duplicated across all applications.


core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -265,32 +268,7 @@ public MethodMetrics(ResourceMethod method, PerformanceMetric annotation, Metric
               "The 99th percentile request latency in ms", allTags), 99));
       this.requestLatencySensor.add(percs);
 
-      for (int i = 0; i < errorSensorByStatus.length; i++) {
```
trnguyencflt, line 268: this is moved to a helper function to reduce code complexity

jliunyu (approved):
LGTM

trnguyencflt:
thanks @jliunyu

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -327,22 +337,39 @@ public void finished(long requestSize, long responseSize, long latencyMs) {
      * Indicate that a request has failed with an exception.
      */
     public void exception(final RequestEvent event) {
-      //map the http status codes down to their classes (2xx, 4xx, 5xx)
-      // use the containerResponse status as it has the http status after ExceptionMappers
-      // are applied
-      int idx = event.getContainerResponse() != null
```
msn-tldr, line 333: +1 on getting rid of custom logic to get http status codes.

msn-tldr:
@trnguyencflt 

My understanding is this -
1. This https://github.com/confluentinc/rest-utils/blob/master/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java is registered at the Jersey layer.
2. BUT the global-dos & per-ip connection filters run at the Jetty [layer](https://github.com/confluentinc/rest-utils/blob/5f6b356a42be54c25a2a5a68b3c99d92cc165f88/core/src/main/java/io/confluent/rest/Application.java#L370).
So capturing 429s at MethodApplicationListener.java isn't enough, as it would miss 429s at Jetty layer. Probably that's why @ehumber chose to do as a separate request handler. What do you think?

trnguyencflt:
@msn-tldr interesting take, that might be the reason that @ehumber chose to use RequestLogHandler. 

However, considering the issues of the current implementation of Rest429CustomResponseHandler
- it reports metrics for all Jersey applications and jetty - which cause the current issue
- and in the tag of the metrics itself, which is type=jersey-metrics, so the context of this is only account for jersey layer.

So I would say, this PR fits for the current purpose, because, the goal of this is to provide correct jersey layer 429 metrics for each jersey application (internal + external)

About jetty layer dos metrics, we would need a different metrics context (type=jetty) and scope that into a ticket.

What do you think?

msn-tldr:
@trnguyencflt It *doesn't solve* the original problem. Main purpose is to detect the noisy neighbor when it happens. It won't detect the below scenario since its at Jetty layer -

```
Client 1 : using multiple ip addresses

Uses up all of the 25 requests/s using 4 different client IPs

Client 2

sends 1 request a second to REST

Jetty still has no space left to allow this through in it’s global DoS filter
```
So we still a need solution for the original problem 😞

msn-tldr:
> this PR only addresses the 429 metrics of jersey layer, which is part of the noisy neighbour (per LKC).
then there will be another metrics on actual Jetty layer, where 429 can be measured by dos filter.

@trnguyencflt Make sense :)

There is little value in adding an alert on it in Datadog. As per LKC limit being tripped, is OK for us, since its a customer over utilising their own quota. It shouldn't affect other customers. Would the recently added datadog alert trigger on this metric? If yes, that monitor should be disabled for now.

Can you add a jira for tracking `another metrics on actual Jetty layer, where 429 can be measured by dos filter.`? And that datadog monitor should alert on this.

trnguyencflt:
@msn-tldr 

> There is little value in adding an alert on it in Datadog. As per LKC limit being tripped, is OK for us, since its a customer over utilising their own quota. It shouldn't affect other customers. Would the recently added datadog alert trigger on this metric? If yes, that monitor should be disabled for now.

AFAIK, the Datadog monitor is only visible to Confluent and alert only sends to us so we can still have a leeway on this. Also monitor is applied on PKC level IIRC so it's still worth having the monitor so that we understand what is going on on Jersey layer. What should be changed is the wording on the monitor and metrics dashboard to distinguish the current metrics (so that people will not be confused).

 > Can you add a jira for tracking another metrics on actual Jetty layer, where 429 can be measured by dos filter.? And that datadog monitor should alert on this.
Definitely 😃

PR #371, opened 2023-02-24
CIAM-2420: Add max buffer header size configs

keahung:
One of the query parameters for the rbac list endpoints is getting too large (it is a list of all the internal service accounts in an org). As a quick fix, make the jetty response/request max header size configurable in `ApplicationServer`.

### DIFF
--- a/core/src/test/java/io/confluent/rest/ApplicationServerTest.java
+++ b/core/src/test/java/io/confluent/rest/ApplicationServerTest.java
@@ -131,30 +131,36 @@ public class ApplicationServerTest {
     assertThat(makeGetRequest("/app2/index.html"), is(Code.OK));
   }
 
-  List<URL> getListeners() {
+  List<URL> getListeners(ApplicationServer<TestRestConfig> server) {
     return Arrays.stream(server.getConnectors())
-            .filter(ServerConnector.class::isInstance)
-            .map(ServerConnector.class::cast)
-            .map(connector -> {
-              try {
-                String protocol = new HashSet<>(connector.getProtocols())
-                        .stream()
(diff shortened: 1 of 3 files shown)


rigelbm (approved):
LGTM

cyrusv (approved):
Nice test

milosimpson (approved):
Nice

core/src/test/java/io/confluent/rest/ApplicationServerTest.java
```diff
@@ -241,6 +247,37 @@ public void testInvalidThreadPoolConfigQueueCapacityValid() throws Exception {
 
   }
 
+  @Test
```
confluent-alec-kendall, line 250: Only now getting a chance to review. I think a longer, more descriptive and explicit method name would be helpful. For instance, here's how I would rewrite this method name:
- applicationServer_whenMaxHeaderSizeIncreased_shouldNotThrowExceptionWhenRequestHeaderExceedsDefault()

There are a few benefits to doing this. 

1. Easier and faster to find something / know exactly what's tested
Consider the two scenarios where you're identifying a potential source of a bug or writing a test for a newly discovered bug. In both cases, explicit unit test names and testing a single thing can greatly improve developer productivity.

2. Forces you to really consider what you're testing
This has a number of benefits. One is that it prevents the test from containing more than it needs. For instance, I'd argue that the part of your test that tests whether the default server rejects a request > default max header size is outside the scope of the test. Why? Because at this point you're testing the default functionality of the server, rather than the code changes you wrote - unit test scope creep. I definitely understand where you're coming from though. So, food for thought

PR #373, opened 2023-03-08
CIAM-2673: initialize start time in metrics before onEvent() is called

confluent-altang:
We're noticing severe latency spikes that correspond with API errors. Upon investigation, it seems that the `started` variable in `MetricsRequestEventListener` does not get initialized in `onEvent()` when an error occurs, resulting in a total latency calculation of `Time.milliseconds() - 0`. This CR initializes `started` in the constructor to avoid false latency reports.

### DIFF
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -412,7 +412,8 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
     public MetricsRequestEventListener(final Map<Method, RequestScopedMetrics> metrics, Time time) {
       this.metrics = metrics;
       this.time = time;
-      // CIAM-2673: if an exception occurs, MATCHING_START is never reached, resulting in false latency metrics
+      // CIAM-2673: if an exception occur in a filter that runs before this method listener,
+      // MATCHING_START is never reached, resulting in false latency metrics
       this.started = time.milliseconds();
     }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -412,6 +412,8 @@ private static class MetricsRequestEventListener implements RequestEventListener
     public MetricsRequestEventListener(final Map<Method, RequestScopedMetrics> metrics, Time time) {
       this.metrics = metrics;
       this.time = time;
+      // CIAM-2673: if an exception occurs, MATCHING_START is never reached, resulting in false latency metrics
+      this.started = time.milliseconds();
```
dragonmushu, line 416: High level I think this should work, but I would like to confirm when the EventsListener is instantiated. Please add tests to verify this works correctly.

https://eclipse-ee4j.github.io/jersey.github.io/apidocs/2.34/jersey/org/glassfish/jersey/server/monitoring/ApplicationEventListener.html

It looks like it should be created once per request at RequestEvent.Type.START

cyrusv: Is this better than doing it after line 420?

dragonmushu: RequestEvent.Type.START won't be passed there: https://eclipse-ee4j.github.io/jersey.github.io/apidocs/2.34/jersey/org/glassfish/jersey/server/monitoring/RequestEventListener.html

`RequestEvent.Type.Start` is only handled by the ApplicationEventListener in the `onRequest` method.

confluent-altang:
Will not merge. Requires further investigation into integration testing CIAM-2673. See Jira ticket for more details.

PR #374, opened 2023-03-09
Make SNI host check explicit

trnguyencflt:
The default value of SniHostCheck in SecureRequestCustomizer class is true. That means this option is always enabled in our Kafka REST.

This PR is just making sure that we explicitly set this to true (for example, some time later there is a breaking change in jetty library if we ever update it).

### DIFF
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -228,7 +228,10 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
     for (NamedURI listener : listeners) {
       if (listener.getUri().getScheme().equals("https")) {
         if (httpConfiguration.getCustomizer(SecureRequestCustomizer.class) == null) {
-          httpConfiguration.addCustomizer(new SecureRequestCustomizer());
+          SecureRequestCustomizer secureRequestCustomizer = new SecureRequestCustomizer();
+          // Explicitly making sure that SNI is checked against Host in HTTP request
+          secureRequestCustomizer.setSniHostCheck(true);
+          httpConfiguration.addCustomizer(secureRequestCustomizer);
         }
       }
       addConnectorForListener(httpConfiguration, httpConnectionFactory, listener,


core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -228,7 +228,10 @@ private void configureConnectors() {
     for (NamedURI listener : listeners) {
       if (listener.getUri().getScheme().equals("https")) {
         if (httpConfiguration.getCustomizer(SecureRequestCustomizer.class) == null) {
-          httpConfiguration.addCustomizer(new SecureRequestCustomizer());
+          SecureRequestCustomizer secureRequestCustomizer = new SecureRequestCustomizer();
+          // Explicitly making sure that SNI is checked against Host in HTTP request
+          secureRequestCustomizer.setSniHostCheck(true);
```
apoorvmittal10, line 233: Wouldn't it be better to write the `assert` instead of explicit check as the config is enabled by default?

```suggestion
          assert secureRequestCustomizer.isSniHostCheck();
```

trnguyencflt: @apoorvmittal10 that also works, however I'm not sure if we're allowed to use assert in source (not tests) in Confluent.

apoorvmittal10: I can see several examples of that in [ce-kafka](https://github.com/search?q=repo%3Aconfluentinc%2Fce-kafka%20%22assert%20%22&type=code). Moreover it's java `assert` keyword unlike any `Assertion` from test library. I think it should be safe to use that.
cc: @AndrewJSchofield

AndrewJSchofield: Given that precedent, I would say it's OK.

apoorvmittal10 (approved):

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -228,7 +228,10 @@ private void configureConnectors() {
     for (NamedURI listener : listeners) {
       if (listener.getUri().getScheme().equals("https")) {
         if (httpConfiguration.getCustomizer(SecureRequestCustomizer.class) == null) {
-          httpConfiguration.addCustomizer(new SecureRequestCustomizer());
+          SecureRequestCustomizer secureRequestCustomizer = new SecureRequestCustomizer();
+          // Explicitly making sure that SNI is checked against Host in HTTP request
+          assert secureRequestCustomizer.isSniHostCheck();
```
kirktrue, line 233: Since I don't know this code base, this comment might not be relevant, but...

Can we change the `assert` to an `if` statement with some logging? IIRC, the message the `assert` keyword spits out is very terse. I would think it would be helpful to add some logging for the user to correct the situation. Or maybe that's not possible in this case?

kirktrue (changes requested):
Sorry. I'm not much help because I don't know the code base enough to know if this solves the problem. Just a minor nitpick about `assert`.

core/src/main/java/io/confluent/rest/ApplicationServer.java
trnguyencflt: thanks for pointing out, assert can spit out an uneasy error message, what is best I think is I'll change it to `Preconditions.checkArgument` for which we can write a nicer error message.

trnguyencflt (commented):
@kirktrue could you please have a look if this is ok to merge? thanks

kirktrue (approved):

PR #379, opened 2023-03-30
Remove stacktrace in response when server throws BadMessageException

trnguyencflt:
This will make the response nicer, in term of no stack trace will show up for error on server

- [x] test to show this works

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -22,7 +22,6 @@ import static java.util.Collections.emptyMap;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.confluent.rest.auth.AuthUtil;
-import io.confluent.rest.errorhandlers.NoJettyDefaultStackTraceErrorHandler;
 import io.confluent.rest.exceptions.ConstraintViolationExceptionMapper;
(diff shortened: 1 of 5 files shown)


AndrewJSchofield (commented):
I'd like to see a test where the new exception mapper is invoked as a result of a REST call and then demonstrates that there's a JSON response.

trnguyencflt:
@AndrewJSchofield I'm deploying this to a test cluster and will include the response for a REST request, would that be acceptable?

AndrewJSchofield:
> @AndrewJSchofield I'm deploying this to a test cluster and will include the response for a REST request, would that be acceptable?

I do see that the effect of this might be to make `kafka-rest` behave better, so a test in that repo which shows the benefit of this work is acceptable.

trnguyencflt:
@AndrewJSchofield thanks for pointing out about kafka-rest, I'm testing this with kafka-rest atm, and it turns out the error is raised at lower level than application, so using registerExceptionMappers doesn't work.

Will update this PR and raise another PR for kafka-rest once I figure out the correct error handler.

trnguyencflt:
@AndrewJSchofield this PR should be ready to review now, I have added tests to validate that it should work (remove stack trace from error response).
Could you please have a look? thanks

PR #384, opened 2023-04-11
AUTHN-1586 Add optional latency SLO/SLA metrics in annotation

bcuii:
What is it about?
---
This is to collect server-side metrics for [AUTHN-1586](https://confluentinc.atlassian.net/browse/AUTHN-1586), and is useful if others need it.

Add optional metrics about the count of requests that meet or violate latency SLO/SLA in the `Performance` annotation. 

How?
---
Add three config parameters.
1. (boolean) `METRICS_ENABLE_LATENCY_SLO_SLA_CONFIG`
2. (long) `METRICS_LATENCY_SLO_MS_CONFIG`
3. (long) `METRICS_LATENCY_SLA_MS_CONFIG`

If `METRICS_ENABLE_LATENCY_SLO_SLA_CONFIG` is enabled, add four sensors to count the number of request whose response latency below(above) SLO(SLA) threshold.

[AUTHN-1586]: https://confluentinc.atlassian.net/browse/AUTHN-1586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -164,6 +164,16 @@ public class RestConfig extends AbstractConfig {
       + " be used to specify additional tags during deployment like data center, instance "
       + "details, etc.";
   protected static final String METRICS_TAGS_DEFAULT = "";
+  public static final String METRICS_ENABLE_LATENCY_SLO_SLA_CONFIG =
+      "metrics.enable.latency.slo.sla";
+  protected static final String METRICS_ENABLE_LATENCY_SLO_SLA_DOC = "Enable latency SLO and SLA";
+  protected static final boolean METRICS_ENABLE_LATENCY_SLO_SLA_DEFAULT = false;
+  public static final String METRICS_LATENCY_SLO_MS_CONFIG = "metrics.latency.slo.ms";
+  protected static final String METRICS_LATENCY_SLO_MS_DOC = "Latency SLO (in ms)";
+  protected static final long METRICS_LATENCY_SLO_MS_DEFAULT = 5;
+  public static final String METRICS_LATENCY_SLA_MS_CONFIG = "metrics.latency.sla.ms";
+  protected static final String METRICS_LATENCY_SLA_MS_DOC = "Latency SLA (in ms)";
+  protected static final long METRICS_LATENCY_SLA_MS_DEFAULT = 50;
 
   public static final String SSL_KEYSTORE_RELOAD_CONFIG = "ssl.keystore.reload";
   protected static final String SSL_KEYSTORE_RELOAD_DOC =
@@ -667,6 +677,24 @@ public class RestConfig extends AbstractConfig {
             METRICS_TAGS_DEFAULT,
             Importance.LOW,
             METRICS_TAGS_DOC
+        ).define(
+            METRICS_ENABLE_LATENCY_SLO_SLA_CONFIG,
+            Type.BOOLEAN,
+            METRICS_ENABLE_LATENCY_SLO_SLA_DEFAULT,
+            Importance.LOW,
+            METRICS_ENABLE_LATENCY_SLO_SLA_DOC
+        ).define(
+            METRICS_LATENCY_SLO_MS_CONFIG,
+            Type.LONG,
+            METRICS_LATENCY_SLO_MS_DEFAULT,
+            Importance.LOW,
+            METRICS_LATENCY_SLO_MS_DOC
+        ).define(
+            METRICS_LATENCY_SLA_MS_CONFIG,
+            Type.LONG,
+            METRICS_LATENCY_SLA_MS_DEFAULT,
+            Importance.LOW,
+            METRICS_LATENCY_SLA_MS_DOC
         ).define(
             SSL_KEYSTORE_RELOAD_CONFIG,
             Type.BOOLEAN,
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -166,13 +166,16 @@ public class RestConfig extends AbstractConfig {
   protected static final String METRICS_TAGS_DEFAULT = "";
   public static final String METRICS_ENABLE_LATENCY_SLO_SLA_CONFIG =
       "metrics.enable.latency.slo.sla";
-  protected static final String METRICS_ENABLE_LATENCY_SLO_SLA_DOC = "Enable latency SLO and SLA";
+  protected static final String METRICS_ENABLE_LATENCY_SLO_SLA_DOC = "Whether to enable metrics about the "
+      + "count of requests that meet or violate latency SLO/SLA in the Performance annotation";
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -164,6 +164,19 @@ public class RestConfig extends AbstractConfig {
       + " be used to specify additional tags during deployment like data center, instance "
       + "details, etc.";
   protected static final String METRICS_TAGS_DEFAULT = "";
+  public static final String METRICS_ENABLE_LATENCY_SLO_SLA_CONFIG =
+      "metrics.enable.latency.slo.sla";
```
AndrewJSchofield, line 168: We would typically use a name ending in `enable` for a config. Such as `metrics.latency.slo.sla.enable`.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -292,6 +324,23 @@ public MethodMetrics(ResourceMethod method, PerformanceMetric annotation, Metric
...
+        PerformanceMetric annotation, Metrics metrics, Map<String, String> requestTags,
+        String metricGrpName, Map<String, String> allTags) {
+      for (int i = 0; i < LATENCY_SLO_SLA_SENSOR_NAMES.length; i++) {
+        final Sensor sensor = metrics.sensor(
+            getName(method, annotation, LATENCY_SLO_SLA_SENSOR_NAMES[i], requestTags),
+            null, SENSOR_EXPIRY_SECONDS, Sensor.RecordingLevel.INFO, (Sensor[]) null);
+        MetricName metricName = new MetricName(
+            getName(method, annotation, LATENCY_SLO_SLA_SENSOR_NAMES[i] + "-total"), metricGrpName,
+            (i % 2 == 0 ? "Below" : "Above")
+                + " latency SLA request count, using a cumulative counter",
```
AndrewJSchofield, line 337: This seems like it could be improved. The array of names is a bit awkward in practice, as you can see when the "latency SLA" string is being used even for SLO-related metrics. I would restructure this code and remove the array of names.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -331,6 +380,20 @@ public void finished(long requestSize, long responseSize, long latencyMs) {
       requestSizeSensor.record(requestSize);
       responseSizeSensor.record(responseSize);
       requestLatencySensor.record(latencyMs);
+
+      if (enableLatencySloSla) {
+        if (latencyMs < latencySloMs) {
+          responseLatencySloSlaSensor.get(LATENCY_SLO_SLA_SENSOR_NAMES[0]).record();
```
AndrewJSchofield, line 386: Again, it would be more expressive if the sensor names had separate constants so that it was not necessary to refer back to the array to work out what the members actually are.

AndrewJSchofield (changes requested):
I wonder whether there ought to be some tests too.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
bcuii: Thanks for the comments! 
Yea, it does seem kind of awkward to look back and forth to the array. I followed the [previous people' approach](https://github.com/confluentinc/rest-utils/blob/master/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java#L199-L225) and separated them because they have separate content and hopefully it's more readable.

core/src/test/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListenerIntegrationTest.java
bcuii, line 1: I chose to verify the sum of the metric values for two sensors rather than picking one sensor because simulating latency is tricky. Given the unstable CI environment, where the latency is uncertain each time, if I pick one metric and make an assertion, it sometimes leads to test failure.
If needed, we could mock the latency by making a fake Clock or fake Time anyway, but that would have a larger scope code change. In addition, the results of this part code are quite deterministic.

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -383,15 +405,15 @@ public void finished(long requestSize, long responseSize, long latencyMs) {
 
       if (enableLatencySloSla) {
         if (latencyMs < latencySloMs) {
-          responseLatencySloSlaSensor.get(LATENCY_SLO_SLA_SENSOR_NAMES[0]).record();
+          responseLatencySloSlaSensors.get("response-below-latency-slo").record();
```
AndrewJSchofield, line 408: I think these names ought to be in final static strings. Apart from that, I think this PR is looking good.

AndrewJSchofield (changes requested):

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
bcuii: Make sense.

core/src/test/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListenerIntegrationTest.java
```diff
@@ -67,11 +58,16 @@ public class MetricsResourceMethodApplicationListenerIntegrationTest {
...
-  public void setUp() throws Exception {
+  public void setUp(TestInfo info) throws Exception {
     TestMetricsReporter.reset();
     Properties props = new Properties();
     props.setProperty("debug", "false");
     props.put(RestConfig.METRICS_REPORTER_CLASSES_CONFIG, "io.confluent.rest.TestMetricsReporter");
+
+    if (info.getDisplayName().contains("testMetricLatencySloSlaEnabled")) {
+      props.put(RestConfig.METRICS_LATENCY_SLO_SLA_ENABLE_CONFIG, "true");
+    }
```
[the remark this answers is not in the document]
bcuii: Agreed, set up a somewhat extreme configuration to make the results more deterministic, getting rid of the influence of the CI environment.

niteshmor (approved):
Looks good from my side once we have CI build succeed, but I'm not a code-owner.

AndrewJSchofield (approved):
lgtm

niteshmor:
> Looks good from my side once we have CI build succeed, but I'm not a code-owner.

@bcuii  looks like there are checkstyle errors (line length) in a local build.

bcuii:
> > Looks good from my side once we have CI build succeed, but I'm not a code-owner.
> 
> @bcuii looks like there are checkstyle errors (line length) in a local build.

Thx! Fixed checkstyle issue.

PR #395, opened 2023-05-11
Don't use DosFilter.Listener.onRequestOverLimit default to avoid verbosity

trnguyencflt:
The default function DosFilter.Listener.onRequestOverLimit will log this line 
```
[WARN] 2023-05-11 09:36:40,954 [qtp787738572-1439] org.eclipse.jetty.servlets.DoSFilter - DOS ALERT: Request rejected ip=127.0.0.1, overlimit=OverLimit@11dd96d5[type=IP, id=127.0.0.1, duration=PT0.361698758S, count=100], session=null, user=null
```
whenever a request is overlimited by DosFilter, this add unnecessary verbosity.

This PR disables this log line programmatically.

There is an alternative to disable this log line is by setting log level of `org.eclipse.jetty.servlets.DoSFilter` class to ERROR level.


AndrewJSchofield (approved):
lgtm.

An alternative would be to change the logging configuration perhaps.

msn-tldr (approved):
LGTM

PR #397, opened 2023-06-02
KREST-10243 Add ability to run multiple dosfilter-listeners and setup a custom-log on the app.

msn-tldr:
This PR introduces 2 changes majorly

1. `JettyDosFilterMultiListener` is type of `DosFilter.Listener` that can hold multiple such listeners and run them. This is needed as Jetty's `DosFilter.setListener()` only accepts and runs `1 listener`. Whereas Kafka-rest needs to run another listener, other than the common one for all apps `Jetty429MetricsDosFilterListener`. This other listener will add metadata to the `HttpServletRequest` for custom-logging enhancement for kafka-rest(to be followed in another PR).
* `Application` exposes setters to set-up multi-listeners for the global & non-global dos-filters.
2. `Application`'s `requestLog` was final. Make it non-final, so child-applications can reset it, if needed. This would be used by `KafkaRestApplication` in `kafka-rest`.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -129,7 +129,7 @@ public abstract class Application<T extends RestConfig> {
     this(config, path, listenerName, null);
   }
 
-  public Application(T config, String path, String listenerName, RequestLog inputRequestLog) {
+  public Application(T config, String path, String listenerName, RequestLog customRequestLog) {
     this.config = config;
     this.path = Objects.requireNonNull(path);
     this.listenerName = listenerName;
@@ -140,13 +140,13 @@ public abstract class Application<T extends RestConfig> {
         this.getMetricsTags(),
         config.getString(RestConfig.METRICS_JMX_PREFIX_CONFIG));
 
-    if (inputRequestLog == null) {
+    if (customRequestLog == null) {
       Slf4jRequestLogWriter logWriter = new Slf4jRequestLogWriter();
       logWriter.setLoggerName(config.getString(RestConfig.REQUEST_LOGGER_NAME_CONFIG));
       // %{ms}T logs request time in milliseconds
       requestLog = new CustomRequestLog(logWriter, requestLogFormat());
     } else {
-      requestLog = inputRequestLog;
+      requestLog = customRequestLog;
     }
   }
 
@@ -829,10 +829,6 @@ public abstract class Application<T extends RestConfig> {
    * A rate-limiter that applies a single limit to the entire server.
    */
   private static final class GlobalDosFilter extends DoSFilter {
-
-    public GlobalDosFilter() {
-      super();
-    }
 
     @Override
     protected String extractUserId(ServletRequest request) {
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -21,7 +21,6 @@ import static io.confluent.rest.RestConfig.WEBSOCKET_SERVLET_INITIALIZERS_CLASSE
 import static java.util.Collections.emptyMap;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.annotations.VisibleForTesting;
 import io.confluent.rest.auth.AuthUtil;
 import io.confluent.rest.exceptions.ConstraintViolationExceptionMapper;
 import io.confluent.rest.exceptions.GenericExceptionMapper;
@@ -66,6 +65,7 @@ import org.eclipse.jetty.security.authentication.BasicAuthenticator;
 import org.eclipse.jetty.security.authentication.LoginAuthenticator;
 import org.eclipse.jetty.server.CustomRequestLog;
 import org.eclipse.jetty.server.Handler;
+import org.eclipse.jetty.server.RequestLog;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.Slf4jRequestLogWriter;
 import org.eclipse.jetty.server.handler.HandlerCollection;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -136,16 +140,32 @@ public Application(T config, String path, String listenerName) {
...
     } else {
-      requestLog = customRequestLog;
+      requestLog = inputRequestLog;
     }
   }
 
+  /**
+   * Set DosFilter.listeners for global-dosfilter
+   */
+  public void setGlobalDosfilterListeners(
```
trnguyencflt, line 156: Not sure we need this public api as those listeners are only triggered on DoSFilter, for which rest-utils is the only repo that uses dosfilter.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -136,16 +140,32 @@ public Application(T config, String path, String listenerName) {
...
+      List<Listener> listeners) {
+    this.globalDosfilterListeners = Objects.requireNonNull(listeners);
+  }
+
+  /**
+   * Set DosFilter.listeners for non-global-dosfilter
+   */
+  public void setNonGlobalDosfilterListeners(
```
trnguyencflt, line 164: same as above comment

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -124,8 +129,7 @@ public Application(T config, String path, String listenerName) {
     this(config, path, listenerName, null);
   }
 
-  @VisibleForTesting
-  Application(T config, String path, String listenerName, CustomRequestLog customRequestLog) {
+  public Application(T config, String path, String listenerName, RequestLog inputRequestLog) {
```
trnguyencflt, line 132: I think the name `customRequestLog` suites this better.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -803,6 +830,10 @@ public void onShutdown() {
    */
   private static final class GlobalDosFilter extends DoSFilter {
 
+    public GlobalDosFilter() {
```
trnguyencflt, line 833: I don't think this constructor is needed.

core/src/main/java/io/confluent/rest/JettyDosFilterMultiListener.java
```diff
@@ -0,0 +1,58 @@
...
+  public Action onRequestOverLimit(HttpServletRequest request, OverLimit overlimit,
+      DoSFilter dosFilter) {
+    // KREST-10418: we don't use super function to get action object because
+    // it will log a WARN line, in order to reduce verbosity
+    Action action = Action.fromDelay(dosFilter.getDelayMs());
+    for (DoSFilter.Listener listener : listeners) {
+      try {
+        listener.onRequestOverLimit(request, overlimit, dosFilter);
+      } catch (Exception ex) {
+        log.error("DosFilter.Listener threw exception {}", ex);
```
trnguyencflt, line 52: this might result in very verbose log if many requests are limited and one of the listener has error, I think debug level is more suitable (as we can dynamically change logging level).

Also could you log the classname of the listener instead of general `DosFilter.Listener`, so that we understand from which listener threw the exception.

core/src/main/java/io/confluent/rest/Application.java
msn-tldr: I chose the option to offer this as a public API as user of `rest-utils` would want to do custom-actions on when each of those filters trigger. For example, `kafka-rest` will use it to configure dosfilter.listeners that add request-attributes for when requests get-ratelimited by Jetty filters(see this [code](https://github.com/confluentinc/kafka-rest/blob/1cedf144e133223ab1e60c630ee2014661a0f87b/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestMain.java#L57) WIP PR). These request-attributes are then logged in the custom logger.

core/src/main/java/io/confluent/rest/Application.java
msn-tldr: see above.

core/src/main/java/io/confluent/rest/Application.java
msn-tldr: Somehow snuck-in, was only for local dev, good spot.

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -103,7 +105,7 @@
 
   protected ApplicationServer<?> server;
   protected Metrics metrics;
-  protected final CustomRequestLog requestLog;
+  protected RequestLog requestLog;
```
trnguyencflt, line 108: I would prefer this field is final then the constructor with `RequestLog` public instead (in previous commit), was there a problem with that approach?

msn-tldr: https://github.com/confluentinc/kafka-rest/pull/1168/files#diff-0dbc682fe5c6635c3e6dc268c10b1adc54544e47ddd2e86815175935fcba24e7R143

`KafkaRestApplication` won't be able to set `requestLog` either from `KafkaRestApplication` ctor, or configureHandler().

trnguyencflt: You could do this with static method though, for example
```
  public KafkaRestApplication(KafkaRestConfig config, String path, String listenerName) {
    super(config, path, listenerName, createRequestLog(config, CustomRequestLog.EXTENDED_NCSA_FORMAT + " %{ms}T"));

    restResourceExtensions =
        config.getConfiguredInstances(
            KafkaRestConfig.KAFKA_REST_RESOURCE_EXTENSION_CONFIG, RestResourceExtension.class);
    config.setMetrics(metrics);
  }
  
  private static RequestLog createRequestLog(KafkaRestConfig config, String requestLogFormat) {
    Slf4jRequestLogWriter logWriter = new Slf4jRequestLogWriter();
    logWriter.setLoggerName(config.getString(RestConfig.REQUEST_LOGGER_NAME_CONFIG));
    // %{ms}T logs request time in milliseconds
    return new CustomRequestLog(logWriter, requestLogFormat);
  }
```

trnguyencflt (approved):
LGTM

PR #400, opened 2023-06-26
Initial draft of a simple filter

TemirAskhat:
This PR is to provide Jetty layer count metrics, i.e those requests are not rejected by any of the filters. 
* A general filter that won't filter requests on the jetty layer 
* Request-total-count
* Under group name of jetty-metrics, with the tag http_status_code:xxx

### DIFF
--- a/core/src/main/java/io/confluent/rest/filters/JettyRequestsSimpleFilter.java
+++ b/core/src/main/java/io/confluent/rest/filters/JettyRequestsSimpleFilter.java
@@ -19,7 +19,6 @@ import org.slf4j.LoggerFactory;
 public class JettyRequestsSimpleFilter implements Filter {
   private static final Logger log = LoggerFactory.getLogger(JettyRequestsSimpleFilter.class);
   private JettyAllRequestsListener _listener;
-
   @Override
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/confluent/rest/filters/JettyRequestsSimpleFilter.java
```diff
@@ -0,0 +1,43 @@
...
+  private static final Logger log = LoggerFactory.getLogger(JettyRequestsSimpleFilter.class);
+  private JettyAllRequestsListener _listener;
+  @Override
+  public void init(final FilterConfig filterConfig) throws ServletException {
+  }
+  public void setListener(JettyAllRequestsListener listener) {
+    this._listener = (JettyAllRequestsListener) Objects.requireNonNull(listener, "Listener may not be null");
+  }
+  @Override
+  public void doFilter(final ServletRequest servletRequest, final ServletResponse servletResponse,
```
trnguyencflt, line 29: to reiterate @msn-tldr comment, just this function is enough, basically, you will need to migrate the code from JettyAllRequestsListener to here.

PR #401, opened 2023-06-30
KREST-10790 Handle TopicDeletionDisabledException as 400

msn-tldr:
`TopicDeletionDisabledException` is only thrown by broker if
1. Its topic-deletion is disabled all together on user's kafka-cluster(uncommon setting).
2. Its a mirrored-topic, so can't be explicitly deleted.

In both cases, its a user-error, should be treated as such with `400` by kafka-rest


trnguyencflt (approved):
LGTM

PR #403, opened 2023-07-07
KREST-10989: Handle ThrottlingQuotaExceededException exception as 429

apoorvmittal10:
ThrottlingQuotaExceededException denotes quota breach. This error code should be treated as 429 instead of 500.


msn-tldr (approved):
LGTM

PR #405, opened 2023-07-20
Make sure that ErrorMessage is always interpreted as json in Response

trnguyencflt:
As title said, this makes sure that we explicitly specify application/json media type for ErrorMessage, otherwise, with Options request, it can return 500

### DIFF
--- a/core/src/test/java/io/confluent/rest/ErrorMessageIntegrationTest.java
+++ b/core/src/test/java/io/confluent/rest/ErrorMessageIntegrationTest.java
@@ -1,0 +1,184 @@
+/*
+ * Copyright 2014 - 2023 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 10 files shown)


msn-tldr (approved):
LGTM

core/src/test/java/io/confluent/rest/ErrorMessageIntegrationTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+  @AfterEach
+  public void tearDown() throws Exception {
+    server.stop();
+    server.join();
+    client.close();
+  }
+
+  @Test
+  @DisplayName("testOPTIONS_WithoutMediaType_return500")
```
ehumber, line 71: Why test this path?  If we've got the fix right then all endpoints return properly (ie not 500) because we've set mediaType everywhere it's needed.

ehumber:
Does WebApplicationExceptionMapper https://github.com/confluentinc/rest-utils/blob/86e536588148b8ffc644187e0a7afa9d2279b9bc/core/src/main/java/io/confluent/rest/exceptions/WebApplicationExceptionMapper.java#L69 need the same change?

ehumber (approved):

core/src/test/java/io/confluent/rest/ErrorMessageIntegrationTest.java
trnguyencflt: This test is just demonstration that without MediaType, we would get 500, I put a comment on this

trnguyencflt:
> Does WebApplicationExceptionMapper
> 
> https://github.com/confluentinc/rest-utils/blob/86e536588148b8ffc644187e0a7afa9d2279b9bc/core/src/main/java/io/confluent/rest/exceptions/WebApplicationExceptionMapper.java#L69
> 
> need the same change?

We don't need it here as it's already set in `createResponse` call above it.

PR #406, opened 2023-08-01
Implementation of RateLimitNetworkTrafficListener

trnguyencflt:
This PR implements a method to slowdown the reading rate from HTTP/TCP socket in Jetty. The approach is based on the fact that we can use NetworkTrafficListener to count for the reading from TCP sockets on all connections. This introduces 3 new configurations (there will be changed in cc-spec-kafka):
1. `network.traffic.rate.limit.enable`: whether to enable this rate limit
2. `network.traffic.rate.limit.backend`: backend of rate limiting, either guava or resilience4j
3. `network.traffic.rate.limit.bytes.per.sec`: the rate to limit, default to 20MiB

Reference JIRA https://confluentinc.atlassian.net/browse/KREST-11222

### DIFF
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -129,7 +129,7 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
   }
 
   private void attachMetricsListener(Metrics metrics, Map<String, String> tags) {
-    NetworkTrafficListener metricsListener = new MetricsListener(metrics, "jetty", tags);
+    MetricsListener metricsListener = new MetricsListener(metrics, "jetty", tags);
     for (NetworkTrafficServerConnector connector : connectors) {
       connector.addNetworkTrafficListener(metricsListener);
     }
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -36,6 +36,7 @@ import org.eclipse.jetty.http.HttpCompliance;
 import org.eclipse.jetty.http.HttpVersion;
 import org.eclipse.jetty.http2.server.HTTP2CServerConnectionFactory;
 import org.eclipse.jetty.http2.server.HTTP2ServerConnectionFactory;
+import org.eclipse.jetty.io.NetworkTrafficListener;
 import org.eclipse.jetty.jmx.MBeanContainer;
 import org.eclipse.jetty.server.ConnectionFactory;
 import org.eclipse.jetty.server.ConnectionLimit;
@@ -128,9 +129,20 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
   }
 
   private void attachMetricsListener(Metrics metrics, Map<String, String> tags) {
-    MetricsListener metricsListener = new MetricsListener(metrics, "jetty", tags);
-    for (NetworkTrafficServerConnector connector : connectors) {
-      connector.addNetworkTrafficListener(metricsListener);
+    if (config.getNetworkTrafficRateLimitEnable()) {
+      NetworkTrafficListener rateLimitListener = new RateLimitNetworkTrafficListener(config);
+      for (NetworkTrafficServerConnector connector : connectors) {
+        connector.addNetworkTrafficListener(rateLimitListener);
+      }
+    }
+  }
+
+  private void attachNetworkTrafficRateLimitListener() {
+    if (config.getNetworkTrafficRateLimitEnable()) {
+      NetworkTrafficListener rateLimitListener = new RateLimitNetworkTrafficListener(config);
+      for (NetworkTrafficServerConnector connector : connectors) {
+        connector.addNetworkTrafficListener(rateLimitListener);
+      }
     }
   }
 
@@ -197,6 +209,7 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
     HandlerCollection wsHandlers = new HandlerCollection();
     for (Application<?> app : applications) {
       attachMetricsListener(app.getMetrics(), app.getMetricsTags());
+      attachNetworkTrafficRateLimitListener();
       addJettyThreadPoolMetrics(app.getMetrics(), app.getMetricsTags());
       handlers.addHandler(app.configureHandler());
       wsHandlers.addHandler(app.configureWebSocketHandler());
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -129,11 +129,9 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
   }
 
   private void attachMetricsListener(Metrics metrics, Map<String, String> tags) {
-    if (config.getNetworkTrafficRateLimitEnable()) {
-      NetworkTrafficListener rateLimitListener = new RateLimitNetworkTrafficListener(config);
-      for (NetworkTrafficServerConnector connector : connectors) {
-        connector.addNetworkTrafficListener(rateLimitListener);
-      }
+    NetworkTrafficListener metricsListener = new MetricsListener(metrics, "jetty", tags);
+    for (NetworkTrafficServerConnector connector : connectors) {
+      connector.addNetworkTrafficListener(metricsListener);
     }
   }
 
--- a/core/src/main/java/io/confluent/rest/RateLimitNetworkTrafficListener.java
+++ b/core/src/main/java/io/confluent/rest/RateLimitNetworkTrafficListener.java
@@ -21,8 +21,17 @@ import io.confluent.rest.ratelimit.NetworkTrafficRateLimiterFactory;
 import java.net.Socket;
 import java.nio.ByteBuffer;
 import org.eclipse.jetty.io.NetworkTrafficListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+/**
+ * This class is used as a way to apply back-pressure on all the network traffic sockets when
+ * incoming rate exceeds a threshold
+ */
 public class RateLimitNetworkTrafficListener implements NetworkTrafficListener {
+
+  private static final Logger log = LoggerFactory.getLogger(RateLimitNetworkTrafficListener.class);
+
   private final NetworkTrafficRateLimiter rateLimiter;
 
   public RateLimitNetworkTrafficListener(RestConfig restConfig) {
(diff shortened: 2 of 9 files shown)


core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -128,12 +129,21 @@ public List<Application<?>> getApplications() {
   }
 
   private void attachMetricsListener(Metrics metrics, Map<String, String> tags) {
-    MetricsListener metricsListener = new MetricsListener(metrics, "jetty", tags);
+    NetworkTrafficListener metricsListener = new MetricsListener(metrics, "jetty", tags);
```
AndrewJSchofield, line 132: This change of class seems very odd. Is this intentional?

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -1066,6 +1087,24 @@ private static ConfigDef incompleteBaseConfigDef() {
...
+            Importance.LOW,
+            NETWORK_TRAFFIC_RATE_LIMIT_ENABLE_DOC
+        ).define(
+            NETWORK_TRAFFIC_RATE_LIMIT_BACKEND_CONFIG,
+            Type.STRING,
+            NETWORK_TRAFFIC_RATE_LIMIT_BACKEND_DEFAULT,
+            Importance.LOW,
+            NETWORK_TRAFFIC_RATE_LIMIT_BACKEND_DOC
+        ).define(
+            NETWORK_TRAFFIC_RATE_LIMIT_BYTES_PER_SEC_CONFIG,
```
AndrewJSchofield, line 1103: I would specify an `atLeast` on this config. Just make sure it's non-negative for example.

trnguyencflt: added the validator for this config

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -197,6 +207,7 @@ protected final void doStart() throws Exception {
     HandlerCollection wsHandlers = new HandlerCollection();
     for (Application<?> app : applications) {
       attachMetricsListener(app.getMetrics(), app.getMetricsTags());
+      attachNetworkTrafficRateLimitListener();
```
ehumber, line 210: Does the order you add them here match the order that the listeners run in?  We want the metrics listener to fire before the rate limiting one, so that we get metrics about everything trying to connect.

I the right ordering matches what you've done, but I wanted to ask if you'd checked.

trnguyencflt: Yes, underneath it is a list of listeners so those are added first will be run first.

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -526,6 +527,28 @@ public class RestConfig extends AbstractConfig {
...
+      "network.traffic.rate.limit.enable";
+  protected static final String NETWORK_TRAFFIC_RATE_LIMIT_ENABLE_DOC =
+      "Whether to enable network traffic rate-limiting. Default is false";
+  protected static final boolean NETWORK_TRAFFIC_RATE_LIMIT_ENABLE_DEFAULT = false;
+
+  protected static final String NETWORK_TRAFFIC_RATE_LIMIT_BACKEND_CONFIG =
+      "network.traffic.rate.limit.backend";
+  protected static final String NETWORK_TRAFFIC_RATE_LIMIT_BACKEND_DOC =
+      "The rate-limiting backend to use. The options are 'guava', and 'resilience4j'."
+          + "Default is 'guava'.";
```
ehumber, line 540: For reference, the kafka-rest default may be guava (and good we match it), but we actually run with the config set to resliance4j, so launch darkly should match.

ehumber (approved):

core/src/main/java/io/confluent/rest/RestConfig.java
trnguyencflt: Yeah, I'll make the config in LD set to resilience4j

msn-tldr: What is the reason to keeping both guava/resilience4j as 2 options? I never understood why kafka-rest gives an option. 
IMO, ideally whatever the reason, would have converged to 1 option by now, no reason to have LD/config.

core/src/main/java/io/confluent/rest/RateLimitNetworkTrafficListener.java
```diff
@@ -0,0 +1,39 @@
...
+
+package io.confluent.rest;
+
+import io.confluent.rest.ratelimit.NetworkTrafficRateLimiter;
+import io.confluent.rest.ratelimit.NetworkTrafficRateLimiterFactory;
+import java.net.Socket;
+import java.nio.ByteBuffer;
+import org.eclipse.jetty.io.NetworkTrafficListener;
+
+public class RateLimitNetworkTrafficListener implements NetworkTrafficListener {
```
msn-tldr, line 25: A comment along the lines of // This is way to apply back-pressure to socket when incoming rate exceeds a threshold.

core/src/test/java/io/confluent/rest/RateLimitNetworkTrafficListenerTest.java
```diff
@@ -0,0 +1,217 @@
...
+
+  @Test
+  @DisplayName("NetworkTrafficRateLimitDisabled")
+  public void testNetworkTrafficRateLimitDisabled_unlimited() throws Exception {
+    long startTime = System.nanoTime();
+    // send 1000 POST requests in 1 second
+    hammerAtConstantRate(app.getServer().getURI(), "/resource", Duration.ofMillis(1), 10, 1000);
+    double durationMillis = (System.nanoTime() - startTime) / 1_000_000.0;
+    // with no rate limit, 1000 requests should finish less than 2 seconds
+    assertThat("Duration must be greater than 1 second", durationMillis >= 1000);
```
msn-tldr, line 102: So my understanding is this.
1. Rate-limit disabled,  1 sec <= time <= 2 sec
2. Rate-limit enabled, time > 2 sec.

Is it possible to increase the time gap between 1 & 2.? Example if rate-limit is enabled, slow down is more drastic > 10 sec. That way test-cases will give high confidence that rate-limit logic is working as expected.

msn-tldr (changes requested):
@trnguyencflt This mostly looks good, thanks for doing the change!

Most of the comments are nits, but is it possible to improve the test? see my comment above.

core/src/test/java/io/confluent/rest/RateLimitNetworkTrafficListenerTest.java
trnguyencflt: @msn-tldr I can do that but that means the CI will take longer 😄

trnguyencflt: Is somewhere around 5seconds good enough to show the effect?

core/src/main/java/io/confluent/rest/RateLimitNetworkTrafficListener.java
```diff
@@ -0,0 +1,39 @@
...
+  private final NetworkTrafficRateLimiter rateLimiter;
+
+  public RateLimitNetworkTrafficListener(RestConfig restConfig) {
+    rateLimiter = NetworkTrafficRateLimiterFactory.create(restConfig);
+  }
+
+  @Override
+  public void incoming(final Socket socket, final ByteBuffer bytes) {
+    int cost = bytes.limit() - bytes.position();
```
msn-tldr, line 34: Also, would there be scenarios where operationally would want to know if back-pressure is being applied?

Worthwhile to add `log.trace` for now for future debugging, if rateLimiter.rateLimit() > 0 i.e. it had to block to get permits.

core/src/test/java/io/confluent/rest/RateLimitNetworkTrafficListenerTest.java
msn-tldr: > Is somewhere around 5seconds good enough to show the effect?
Works. 

> I can do that easily (by changing the configuration value) but that means the CI will take longer 😄
Its OK if test takes upto 1 minute, as long signal is high confidence.

core/src/main/java/io/confluent/rest/RateLimitNetworkTrafficListener.java
trnguyencflt: For sure, I can add a debug line (to avoid too verbose in prod)

core/src/test/java/io/confluent/rest/RateLimitNetworkTrafficListenerTest.java
trnguyencflt: done 👍

trnguyencflt:
@msn-tldr I have addressed the comments, could you please have a look again?

core/src/test/java/io/confluent/rest/RateLimitNetworkTrafficListenerTest.java
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  @Test
+  @DisplayName("NetworkTrafficRateLimitEnabled")
+  public void testNetworkTrafficRateLimitEnabled_Guava_slowdown() throws Exception {
+    long startTime = System.nanoTime();
+    // send 1000 POST requests in 1 second
+    hammerAtConstantRate(app.getServer().getURI(), "/resource", Duration.ofMillis(1), 10, 1000);
+    double durationMillis = (System.nanoTime() - startTime) / 1_000_000.0;
+    // with rate limiting, 1000 requests should finish in more than 2 seconds
```
msn-tldr, line 113: nit: update comment to reflect finish >= 10 seconds.

core/src/test/java/io/confluent/rest/RateLimitNetworkTrafficListenerTest.java
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  @Test
+  @DisplayName("NetworkTrafficRateLimitEnabled_Resilience4j")
+  public void testNetworkTrafficRateLimitEnabled_Resilience4j_slowdown() throws Exception {
+    long startTime = System.nanoTime();
+    // send 1000 POST requests in 1 second
+    hammerAtConstantRate(app.getServer().getURI(), "/resource", Duration.ofMillis(1), 10, 1000);
+    double durationMillis = (System.nanoTime() - startTime) / 1_000_000.0;
+    // with rate limiting, 1000 requests should finish in more than 2 seconds
```
msn-tldr, line 125: nit: update comment to reflect finish >= 10 seconds.

msn-tldr (approved):
LGTM, left feedback on updating comments before merging.

core/src/main/java/io/confluent/rest/RestConfig.java
trnguyencflt: it is some historical reason I think when it first developed

PR #408, opened 2023-08-11
KREST-11276 - Tighten SSL auto cert reload

AndrewJSchofield:
Flaky test for SSL cert auto reload seems to be failing because the code itself isn't rock-solid :)

### DIFF
--- a/core/src/main/java/io/confluent/rest/SslFactory.java
+++ b/core/src/main/java/io/confluent/rest/SslFactory.java
@@ -51,6 +51,7 @@ public final class SslFactory {
                 // Need to reset the key store path for symbolic link case
                 sslContextFactory.setKeyStorePath(sslConfig.getKeyStorePath());
                 sslContextFactory.reload(scf -> log.info("SSL cert auto reload begun"));
+                log.info("SSL cert auto reload done");
               }
           );
           log.info("Enabled SSL cert auto reload for: " + watchLocation);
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/confluent/rest/SslFactory.java
```diff
@@ -50,12 +50,12 @@ public static SslContextFactory createSslContextFactory(SslConfig sslConfig) {
           FileWatcher.onFileChange(watchLocation, () -> {
                 // Need to reset the key store path for symbolic link case
                 sslContextFactory.setKeyStorePath(sslConfig.getKeyStorePath());
-                sslContextFactory.reload(scf -> log.info("Reloaded SSL cert"));
+                sslContextFactory.reload(scf -> log.info("SSL cert auto reload begun"));
               }
```
omkreddy, line 54: Can we add a log after SSL reload completes

AndrewJSchofield: Couldn't figure out how to do it.

omkreddy: Can't we add a log after `sslContextFactory.reload` line

AndrewJSchofield: Is it a properly synchronous call with all side-effects guaranteed to have been completed by the time it returns? Why does the reload method take a `Consumer` if it's actually no use at all? I will add a log line. I doubt it is really that informative in practice.

trnguyencflt (approved):

msn-tldr (approved):

PR #409, opened 2023-08-18
KREST-11276 - Fixing cert reload with multiple registered listeners

apoorvmittal10:
Though Multiple FileWatchers are created for multiple listeners but they execute on `static` single thread pool. As the `run` method continuously executes hence only one among multiple submitted FileWatcher will run.

```
jmap -histo 1 | grep -e FileWatcher
4360:             2             64  io.confluent.rest.FileWatcher
8205:             1             16  io.confluent.rest.FileWatcher$1
```

When SslFactory registers multiple listeners for cert dynamic reload then only one among listener will get new certificate.

Current PR creates individual thread executor which guarantees execution of all FileWatcher and cert reload on all listener. 

We might do better here in future with registering all file change notifications. Rather than creating multiple file watcher services we can register for change notification over common single executor.

### DIFF
--- a/core/src/main/java/io/confluent/rest/FileWatcher.java
+++ b/core/src/main/java/io/confluent/rest/FileWatcher.java
@@ -31,19 +31,10 @@ import java.nio.file.WatchEvent;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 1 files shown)


AndrewJSchofield (approved):

msn-tldr (approved):
lgtm

PR #410, opened 2023-09-07
KREST-7955 421 misdirected request if host does not match SNI

trnguyencflt:
This PR adds SniHandler that checks for SNI against the host of the request, if they don't match, response with 421.

This handler is controlled by configuration to turn off/on, default is disabled.
`sni.check.enabled`

[Jira ticket](https://confluentinc.atlassian.net/browse/KREST-7955)

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -413,12 +413,14 @@ public abstract class Application<T extends RestConfig> {
     requestLogHandler.setRequestLog(requestLog);
     context.insertHandler(requestLogHandler);
 
-    HandlerCollection handlerCollection = new HandlerCollection();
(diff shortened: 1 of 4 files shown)


msn-tldr (approved):
lgtm

msn-tldr (changes requested):
@trnguyencflt Thinking more on it, it requires more work to be ready. This new feature should have "config" that acts as a feature-flag to safely roll this out.

trnguyencflt:
@msn-tldr 

> This new feature should have "config" that acts as a feature-flag to safely roll this out

This will be done on the subsequent cc-spec change (when http2 enable -> enable sni check).

For this feature, it is complete on it's own code because we have a configuration to switch it on/off.

msn-tldr (approved):
lgtm

PR #411, opened 2023-09-28
KREST-11705 Map RecordTooLargeException to KAFKA_BAD_REQUEST_ERROR_CODE

trnguyencflt:
RecordTooLargeException is not handled in KafkaExceptionMapper, thus leading to 50002 error.

Now we handle it and map it to 413 REQUEST_ENTITY_TOO_LARGE


msn-tldr (approved):
lgtm

PR #413, opened 2023-10-09
CNC-397: unpin jetty to use the version defined in common

janjwerner-confluent:
Jetty version is defined in common. Unpin the version here and use the one defined globally

PR #416, opened 2023-10-11
update jetty version from 9.4.51.v20230217 to 9.4.53.v20231009

ankurjoshi9:
Jetty version requires update to 9.4.53.v20231009.


msn-tldr:
@ankurjoshi9 
there is an ongoing PR to get Jetty version from common Vs rest-utils. See this
https://github.com/confluentinc/rest-utils/pull/413#issuecomment-1757251743
I think you should update the 1 in common then, that would take care of other uses of Jetty in confluent too, not just users of rest-utils 😃

PR #420, opened 2023-10-18
CIAM-2673: initialize start time in metrics before onEvent() is called

dragonmushu:
We have been getting paged once or twice every week of latency metrics that fire. When investigating usually these metrics are completely off. The root cause of the issue is the way we calculate the latency metric where the start time is stamped after resource matching. If we have filters that execute before matching, then the start time is not stamped and at the end when global metrics are calculated:

`final long elapsed = time.milliseconds() - started;`

The final time is just the current elapsed time since epoch. To fix this we need to move the start time stamping before resource matching. https://eclipse-ee4j.github.io/jersey.github.io/apidocs/2.34/jersey/org/glassfish/jersey/server/monitoring/ApplicationEventListener.html#onRequest-org.glassfish.jersey.server.monitoring.RequestEvent-

@confluent-altang Had already worked on this: https://github.com/confluentinc/rest-utils/pull/373. I just added a test and removed another line of code.

### DIFF
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -536,7 +536,8 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
       this.metrics = metrics;
       this.time = time;
       this.enableGlobalStatsRequestTags = enableGlobalStatsRequestTags;
-      // CIAM-2673: if an exception occurs, MATCHING_START is never reached, resulting in false latency metrics
+      // CIAM-2673: if an exception occur in a filter that runs before this method listener,
+      // MATCHING_START is never reached, resulting in false latency metrics
       this.started = time.milliseconds();
     }
 
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -536,6 +536,8 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
       this.metrics = metrics;
       this.time = time;
       this.enableGlobalStatsRequestTags = enableGlobalStatsRequestTags;
+      // CIAM-2673: if an exception occurs, MATCHING_START is never reached, resulting in false latency metrics
+      this.started = time.milliseconds();
     }
 
     @Override
--- a/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
+++ b/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
@@ -544,7 +544,6 @@ public class MetricsResourceMethodApplicationListener implements ApplicationEven
     @Override
     public void onEvent(RequestEvent event) {
       if (event.getType() == RequestEvent.Type.MATCHING_START) {
-        started = time.milliseconds();
         final ContainerRequest request = event.getContainerRequest();
         wrappedRequestStream = new CountingInputStream(request.getEntityStream());
         request.setEntityStream(wrappedRequestStream);
--- a/core/src/test/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListenerIntegrationTest.java
+++ b/core/src/test/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListenerIntegrationTest.java
@@ -14,10 +14,20 @@ import io.confluent.rest.exceptions.WebApplicationExceptionMapper;
 
 import javax.ws.rs.core.Response.Status;
 import org.apache.kafka.common.metrics.KafkaMetric;
+import org.eclipse.jetty.security.AbstractLoginService;
+import org.eclipse.jetty.security.ConstraintMapping;
+import org.eclipse.jetty.security.ConstraintSecurityHandler;
+import org.eclipse.jetty.security.IdentityService;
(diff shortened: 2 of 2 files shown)


core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
```diff
@@ -536,12 +536,14 @@ private MetricsRequestEventListener(final Map<Method, RequestScopedMetrics> metr
       this.metrics = metrics;
       this.time = time;
       this.enableGlobalStatsRequestTags = enableGlobalStatsRequestTags;
+      // CIAM-2673: if an exception occur in a filter that runs before this method listener,
```
milosimpson, line 539: An instance of this inner class get created for every request?

Also in situations where "MATCHING_START is never reached" ... RequestEvent.Type.FINISHED still get called?

Curious about alternatives... I wonder if maybe we just don't emit a metric  if `RequestEvent.Type.MATCHING_START` is never hit?

milosimpson (commented):
Questions and comments.

core/src/test/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListenerIntegrationTest.java
```diff
@@ -560,6 +561,21 @@ public void testMetricLatencySloSlaEnabled() {
...
+    makeFilterError();
+
+    Map<String, String> allMetrics = TestMetricsReporter.getMetricTimeseries()
+        .stream()
+        .collect(Collectors.toMap(
+            x -> x.metricName().name(),
+            x -> x.metricValue().toString(),
+            (a, b) -> a));
+
+    assertTrue(Double.valueOf(allMetrics.get("request-latency-avg")) < 1000);
```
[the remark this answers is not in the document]
dragonmushu: yep I had tested without the code change...and the metric had a value of 1E12 :)

core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java
dragonmushu: Yeah we create a container for every request to perform the metrics calculations. https://eclipse-ee4j.github.io/jersey.github.io/apidocs/2.34/jersey/org/glassfish/jersey/server/monitoring/RequestEventListener.html

For RequestEventListener:

> This will register the instance for listening of request events for one particular request. Once the processing of the request is finished, the instance will be ignored by the Jersey runtime and not used for processing of further requests.

Specifically for request starting:

> void onEvent([RequestEvent](https://eclipse-ee4j.github.io/jersey.github.io/apidocs/2.34/jersey/org/glassfish/jersey/server/monitoring/RequestEvent.html) event)
>
>The method is called when new request event occurs. This method will never be called for method [RequestEvent.Type.START](https://eclipse-ee4j.github.io/jersey.github.io/apidocs/2.34/jersey/org/glassfish/jersey/server/monitoring/RequestEvent.Type.html#START) as this event is handled by [ApplicationEventListener](https://eclipse-ee4j.github.io/jersey.github.io/apidocs/2.34/jersey/org/glassfish/jersey/server/monitoring/ApplicationEventListener.html).

dragonmushu: Yeah its possible to not emit any metric if the resource is not matched.

core/src/test/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListenerIntegrationTest.java
```diff
@@ -560,6 +561,21 @@ public void testMetricLatencySloSlaEnabled() {
     assertEquals(0, Double.valueOf(allMetrics.get("response-above-latency-sla-total")).intValue());
   }
 
+  @Test
+  public void testValidLatencyMetricsForErrorBeforeResourceMatching() {
```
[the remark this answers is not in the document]
dragonmushu: updated

milosimpson (approved):
Ok

confluent-altang (approved):
I forgot I did this lmao. LGTM!

core/src/test/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListenerIntegrationTest.java
```diff
@@ -682,6 +716,12 @@ public String fourTwoNine() {
       throw new StatusCodeException(Status.TOO_MANY_REQUESTS, new RuntimeException("kaboom"));
     }
 
+    @GET
+    @Path("/filterError")
+    @PerformanceMetric("filterError")
+    public String filterError() {
+      return "should never get here";
```
msn-tldr, line 723: why not add `junit.fail("should never reach here")` to make sure test fails if it reaches here.
https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html#fail(java.lang.String)

dragonmushu: Actually @msn-tldr I tried this and realized the execution of the test won't stop. The fail just prints and is caught. Either way I think its ok since we have a client call that checks if the endpoint returns 200 ok. I tested and removed the filter error and the test fails due to this assertion.

msn-tldr: @dragonmushu u are right. The goal is to validate in the test that pre-matching filter throws before reaching resource. You check that is returns non-200 code is achieves that.

msn-tldr (approved):

PR #435, opened 2023-11-24
KREST-12705 Relax assertion condition for flaky tests

trnguyencflt:
As explained in code comment, due to the timing of exactly 1 second in DosFilter, sometime we might get one more response200 in CI build, so the assertion is changed to handle the flakiness.


msn-tldr (approved):
lgtm

PR #440, opened 2023-11-28
KREST-11812 Replace EOL javaee/jaxb-api with jarkartaee/jaxb-api

trnguyencflt:
As the title said, this replaces an EoL dependency with a maintained one.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -27,10 +27,9 @@
             <groupId>org.glassfish.jersey.inject</groupId>
             <artifactId>jersey-hk2</artifactId>
         </dependency>
-        <!-- Jersey dependency that was available in the JDK before Java 9 -->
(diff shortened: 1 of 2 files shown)


core/pom.xml
```diff
@@ -27,10 +27,9 @@
             <groupId>org.glassfish.jersey.inject</groupId>
             <artifactId>jersey-hk2</artifactId>
         </dependency>
-        <!-- Jersey dependency that was available in the JDK before Java 9 -->
```
trnguyencflt, line 30: this comment is misleading, it was accidentally copied from the below dependency, so remove it

msn-tldr (approved):
lgtm, nice equivalent ee4j libs are available, and Jenkins/tests also passes.

PR #445, opened 2023-12-14
KREST-13041 Test that SSL cipher suites can be restricted

msn-tldr:
Add a test that when `ssl.cipher.suite` is set to restrict ssl ciphers, only restricted ciphers work with REST.

### DIFF
--- a/core/src/test/java/io/confluent/rest/SslTest.java
+++ b/core/src/test/java/io/confluent/rest/SslTest.java
@@ -243,6 +243,8 @@ public class SslTest {
       app.start();
 
       // Correct Cipher, returns 200.
+      // TLSv1.3 should be specified with TLS_AES_128_GCM_SHA256 only for test-setup, as this cipher is specific to TLSv1.3.
+      // NOTE - the behaviour of restricting cipher in REST is agnostic of TLS protocols.
       int statusCode = makeGetRequest(uri + "/test",
           clientKeystore.getAbsolutePath(), SSL_PASSWORD, SSL_PASSWORD, "TLSv1.3", "TLS_AES_128_GCM_SHA256");
       assertEquals(200, statusCode, EXPECTED_200_MSG);
--- a/core/src/test/java/io/confluent/rest/SslTest.java
+++ b/core/src/test/java/io/confluent/rest/SslTest.java
@@ -383,7 +383,7 @@ public class SslTest {
       String clientKeystorePassword,
       String clientKeyPassword)
       throws Exception {
-    return makeGetRequest(url, clientKeystoreLocation, clientKeystorePassword, clientKeyPassword, "TLSv1.2", null);
+    return makeGetRequest(url, clientKeystoreLocation, clientKeystorePassword, clientKeyPassword, null, null);
   }
 
   // returns the http response status code.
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/confluent/rest/SslTest.java
```diff
@@ -226,6 +227,36 @@ public void testHttpsOnly() throws Exception {
     }
   }
 
+  @Test
```
trnguyencflt, line 230: [minor] can this test be parameterized on tls or ssl versions?

msn-tldr: REST logic to restrict cipher suites is agnostic of protocol, so parameterising on ssl version is unnecessary at the moment.

trnguyencflt: > is agnostic of protocol
that's why paramterize tests on ssl version proves this statement, what do you think?

msn-tldr: Parameterised test would actually add more tests. That signals its important to test for different tls version 1/2/3. 

Using "tlsv1.3" is accidental & orthogonal to the behaviour being tested. Only needed for setup, better to add that as a comment, done now.

trnguyencflt: > Parameterised test would actually add more tests. 

That would be fine, tests in this class running fast, so we would not mind having more tests.

> That signals its important to test for different tls version 1.1/2/3.
This is actually the point of parameterize tests, because there is some cipher that is only applicable for tls version 1.3 as shown in this test. That would help us when we actually restrict the cipher because we would know the error if the restricted cipher isn't compatible with the protocol.

core/src/test/java/io/confluent/rest/SslTest.java
```diff
@@ -226,6 +227,38 @@ public void testHttpsOnly() throws Exception {
...
+    // Restrict HTTPs Server to cipher TLS_AES_128_GCM_SHA256
+    props.put(RestConfig.SSL_CIPHER_SUITES_CONFIG, "TLS_AES_128_GCM_SHA256");
+    TestRestConfig config = new TestRestConfig(props);
+    SslTestApplication app = new SslTestApplication(config);
+    try {
+      app.start();
+
+      // Correct Cipher, returns 200.
+      // TLSv1.3 should be specified with TLS_AES_128_GCM_SHA256 only for test-setup, as this cipher is specific to TLSv1.3.
+      // NOTE - the behaviour of restricting cipher in REST is agnostic of TLS protocols.
```
trnguyencflt, line 247: On this point, there are ciphers suite that only compatible with certain TLS version, for example if we restrict to cipher suite `TLS_AES_128_GCM_SHA256` then only TLS1.3 is supported on the server.

core/src/test/java/io/confluent/rest/SslTest.java
```diff
@@ -226,6 +227,38 @@ public void testHttpsOnly() throws Exception {
...
+      int statusCode = makeGetRequest(uri + "/test",
+          clientKeystore.getAbsolutePath(), SSL_PASSWORD, SSL_PASSWORD, "TLSv1.3", "TLS_AES_128_GCM_SHA256");
+      assertEquals(200, statusCode, EXPECTED_200_MSG);
+      assertMetricsCollected();
+      // Incorrect Cipher, throws an exception.
```
trnguyencflt, line 252: could you also add tests where client specifies the right cipher but tls version < 1.3?

msn-tldr: So there are 3 cases w.r.t ciphers
1. Request fails as cipher is invalid, due to incompatible tls protocol.
2. Request fails as Cipher is invalid, as its not part of the restricted ciphers on the server.
3. Request succeeds, as Cipher is valid, compatible with tls protocol, and part of restricted cipher.

2. & 3. are captured here.

> could you also add tests where client specifies the right cipher but tls version < 1.3?

I think this case 1. But isn't this case unrelated/independent to config `ssl.cipher.suite`, so outside the scope of the original ticket, wdyt?

msn-tldr: Anyways i have added test-case for 1 as its easy enough.

trnguyencflt:
Regarding the scope of the ticket https://confluentinc.atlassian.net/browse/KREST-13040, this PR is sufficient, about TLS protocol support, it can be considered in other scope.

trnguyencflt (approved):

PR #448, opened 2023-12-18
[skip secret scan] KSECURITY-2171: Added support for PEM based keystore.

smjn:
### What
- We have added support for pem stores (pem bundle comprising of certs and secret key).
- This pem bundle will be converted to BCFKS in FIPS mode and PKCS in non-FIPS mode to support SSL operations. Support has been added in `SslFactory` class.

### Why
- Currently rest-utils uses jks keystore & truststores by default or PKCS keystore if configured.
- These 2 types are not BC FIPS compliant. The PKCS does not even load in FIPS approved mode and JKS load but without support for secret keys.
- In order for things to work, we must use BCFKS ks type wherever needed. BCFKS type keystore support is not present in upstream control plane services, hence direct injection of this type is not possible.

### Testing
- We have added new test class `SslFactoryTest` with tests for pem based keystores and truststores.

### Issues
- It was observed that bc*-fips libs do not play well with existing bcpkix* libs which are being used in tests. This is yet to be solved.

### Misc
- In this PR, we have added some static files comprising of private key, root CA and an intermediate CA. The instructions to generate these using openssl could be found in:
  - https://openssl-ca.readthedocs.io/en/latest/create-the-root-pair.html
  - https://openssl-ca.readthedocs.io/en/latest/create-the-intermediate-pair.html
- Basic commands to generate key & cert is:
```
RSA keypair
$ openssl genrsa 2048 > ca-key.pem

X509 cert for CA:
$ openssl req -new -x509 -nodes -days 365000 \
   -key ca-key.pem \
   -out ca-cert.pem
```


core/src/main/java/io/confluent/rest/SslFactory.java
```diff
@@ -31,14 +34,40 @@ public final class SslFactory {
 
   private SslFactory() {}
 
+  private static boolean useBcfks(String provider) {
+    if(provider == null) {
+      return false;
+    }
+    return provider.equals(SslConfigs.FIPS_SSL_PROVIDER);
```
trnguyencflt, line 41: this function can be simplify as Objects.equals(SslConfigs.FIPS_SSL_PROVIDER, provider), because SslConfigs.FIPS_SSL_PROVIDER is not null.

core/src/main/java/io/confluent/rest/SslFactory.java
```diff
@@ -31,14 +34,40 @@ public final class SslFactory {
...
+    return useBcfks(config.getProvider()) ? SslConfigs.FIPS_KEYSTORE_TYPE : config.getKeyStoreType();
+  }
+
   public static SslContextFactory createSslContextFactory(SslConfig sslConfig) {
     SslContextFactory.Server sslContextFactory = new SslContextFactory.Server();
 
     if (!sslConfig.getKeyStorePath().isEmpty()) {
-      sslContextFactory.setKeyStorePath(sslConfig.getKeyStorePath());
-      sslContextFactory.setKeyStorePassword(sslConfig.getKeyStorePassword());
+      if (useBcfks(sslConfig.getProvider())) {
```
trnguyencflt, line 52: this block looks fine, will wait for more details

core/src/main/java/io/confluent/rest/SslFactory.java
```diff
@@ -31,14 +34,40 @@ public final class SslFactory {
...
+  }
+
+  private static String getKeyStoreType(SslConfig config) {
+    return useBcfks(config.getProvider()) ? SslConfigs.FIPS_KEYSTORE_TYPE : config.getKeyStoreType();
+  }
+
   public static SslContextFactory createSslContextFactory(SslConfig sslConfig) {
     SslContextFactory.Server sslContextFactory = new SslContextFactory.Server();
 
     if (!sslConfig.getKeyStorePath().isEmpty()) {
```
trnguyencflt, line 51: do we need to do the same for truststore?

core/src/main/java/io/confluent/rest/SslFactory.java
trnguyencflt, line 1: Please also take into account that we automatic reload of keystore see (`if (sslConfig.getReloadOnKeyStoreChange())`) part, so when keystore is reloaded, it has to be also converted again.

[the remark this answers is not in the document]
smjn: true

core/src/main/java/io/confluent/rest/SslFactoryPemHelper.java
```diff
@@ -0,0 +1,61 @@
...
+import java.security.KeyStore;
+import java.util.Objects;
+
+/**
+ * This class provides util methods to handle Pem type keystore. The primary aim
+ * is to encapsulate the DefaultSslEngineFactory invocation in a separate class.
+ * Based on the provider type, an intermediate keystore of appropriate type will be
+ * created.
+ */
+public class SslFactoryPemHelper {
```
trnguyencflt, line 31: this could be a static inner class in SslFactory

core/src/main/java/io/confluent/rest/SslFactoryPemHelper.java
```diff
@@ -0,0 +1,61 @@
...
+  public static boolean isPemSecurityStore(String keyStoreType) {
+    return Objects.equals(DefaultSslEngineFactory.PEM_TYPE, keyStoreType);
+  }
+
+  private static void checkAndThrowRuntimeException(String keyStoreType, String provider) {
```
trnguyencflt, line 36: this used only once, you can inline this function in `getKeyStoreFromPem`

core/src/main/java/io/confluent/rest/SslFactoryPemHelper.java
smjn: @trnguyencflt 
Wanted to avoid any references to DefaultSslEngineFactory in SslFactory. All code prone to modification has been encapsulated in the helper class.

core/src/test/java/io/confluent/rest/SslFactoryTest.java
```diff
@@ -0,0 +1,227 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class SslFactoryTest {
+  protected static final String CA1 = "-----BEGIN CERTIFICATE-----\n"
```
trnguyencflt, line 24: Could these certificate and private key be generated by code? It is not ideal to have those static value stored here.

smjn: Yes will look into it

core/src/test/resources/certs/cert1.pem
trnguyencflt, line 1: is there expiration date on this cert? test might fail if it runs after expired date.

core/src/test/java/io/confluent/rest/SslFactoryTest.java
```diff
@@ -0,0 +1,112 @@
...
+  private static String KEY;
+  private static String ENCRYPTED_KEY;
+  private static final String PEM_TYPE = "PEM";
+  private static final String PEM_KEYSTORE_NAME_PREFIX = "pemKeyStore";
+  private static final String PEM_KEYSTORE_NAME_SUFFIX = ".pem";
+
+  static {
+    try {
+      Path path = Paths.get(Objects.requireNonNull(
+          SslFactory.class.getClassLoader().getResource("certs/cert1.pem").toURI()));
```
trnguyencflt, line 38: as already commented, please use code to generate certificates/private keypair whenever possible.

core/pom.xml
```diff
@@ -118,6 +118,7 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
+            <version>7.7.0-901-ce</version>
```
trnguyencflt, line 121: please don't use fixed version

test/pom.xml
```diff
@@ -46,5 +46,15 @@
             <artifactId>junit-jupiter-api</artifactId>
             <version>${junit.jupiter.version}</version>
         </dependency>
+        <dependency>
```
trnguyencflt, line 49: you don't need to change this file

smjn: This happened automatically I think. I just added to the pom.xml in rest-utils-test

core/pom.xml
smjn: This is due to local testing - I had to generate the artefacts locally as the DefaultSslEngineFactoy change PR is still not merged

test/pom.xml
trnguyencflt: this pom file isn't related to core/pom.xml

core/src/test/java/io/confluent/rest/SslFactoryTest.java
smjn: @trnguyencflt 
The code to generate is quite complicated and will require bouncy castle libs.
Most recommendations are to use openssl locally and use the values produced.

pom.xml
```diff
@@ -289,6 +289,18 @@
...
+            <dependency>
+                <groupId>org.bouncycastle</groupId>
+                <artifactId>bc-fips</artifactId>
+                <version>1.0.2.4</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.bouncycastle</groupId>
+                <artifactId>bctls-fips</artifactId>
+                <version>1.0.13</version>
```
trnguyencflt, line 301: we should use the latest version, which is 1.0.17

core/src/test/java/io/confluent/rest/SslFactoryTest.java
trnguyencflt: > The code to generate is quite complicated and will require bouncy castle libs.

Do you mean it requires non-fips libs to generate?

Is this the same method that we use in ce-kafka PR?

trnguyencflt: > Most recommendations are to use openssl locally and use the values produced.

Could you please add the instruction to run openssl in a README file in resources/certs?

pom.xml
smjn: This is for version compatibility with ce-kafka

core/src/test/java/io/confluent/rest/SslFactoryTest.java
smjn: @trnguyencflt 

* cert1.pem and cert2.pem expiry is 2120 (**97 yrs** should be enough to not need regeneration)

* Yes, current code in ce-kafka uses non-fips (BC) provider to generate artifacts (`CertificateUtils`). As part of aforementioned PR in ce-kafka I've added changed to support FIPS provider - but having that will bring in a new dependency (ce-fips-tests), which can be avoided this way.

* The instructions to create root pair (root CA and private key) and intermediate certs are described in:
https://openssl-ca.readthedocs.io/en/latest/create-the-root-pair.html
https://openssl-ca.readthedocs.io/en/latest/create-the-intermediate-pair.html

pom.xml
trnguyencflt: that's weird, AFAIK, bctls-fips should be on version 1.0.17 for fips 140-2 https://confluentinc.atlassian.net/wiki/spaces/PM/pages/941887331/FedRAMP+Overview+Encryption+FIPS+Validation#Java

core/src/main/java/io/confluent/rest/SslFactory.java
```diff
@@ -16,29 +16,97 @@
...
+        sslContextFactory.setTrustStorePath(sslConfig.getTrustStorePath());
+        if (!setPathOnly) {
+          sslContextFactory.setTrustStorePassword(sslConfig.getTrustStorePassword());
+          sslContextFactory.setTrustStoreType(sslConfig.getTrustStoreType());
+        }
+      }
+    }
+  }
+
+  private static FileWatcher.Callback getCallback(SslConfig sslConfig,
```
trnguyencflt, line 86: should we call this function `onFileChangeCallback`?

fips-tests/src/main/java/io/confluent/rest/DummyClass.java
trnguyencflt, line 1: Could you fix the license format to be the same as in other file?

fips-tests/src/test/java/io/confluent/rest/SslFactoryFipsTest.java
trnguyencflt, line 1: Could you please add license header

core/src/test/java/io/confluent/rest/SslFactoryTest.java
trnguyencflt, line 1: could you please add license header?

fips-tests/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+        <groupId>io.confluent</groupId>
+        <artifactId>rest-utils-parent</artifactId>
+        <version>7.7.0-0</version>
+    </parent>
+
+    <artifactId>rest-utils-fips-tests</artifactId>
+    <packaging>jar</packaging>
+    <name>rest-utils-fips-tests</name>
+
+    <dependencies>
```
trnguyencflt, line 16: are the dependencies minimal? (i.e no redundant dep)

core/src/main/java/io/confluent/rest/SslFactory.java
```diff
@@ -16,29 +16,97 @@
 
 package io.confluent.rest;
 
-import java.nio.file.Path;
```
msn-tldr, line 19: @smjn What is the role out plan for the pem keystores in prod to make sure REST doesn't break?

msn-tldr: Also is this change backwards compatible? Say in "release 10" this change is released to prod, but for reason(bug) the REST is rolled back to "release 9", would it be safe?

smjn: @msn-tldr 
Pem support is additional (on top of current pkcs and jks). Current functionality should remain exactly same. Only when specific NEW Fips clusters are provisioned should the pem codeflow get invoked. This depends on the value of `ssl.provider`. Using Sunjsse or non-fips providers, the codeflow should remain same.

fips-tests/pom.xml
smjn: AFAIK, they are all required. I will do one more iteration and raise revision.

smjn: @trnguyencflt removed unnecessary deps.

core/src/main/java/io/confluent/rest/SslFactory.java
```diff
@@ -16,29 +16,97 @@
...
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
+
 public final class SslFactory {
 
   private static final Logger log = LoggerFactory.getLogger(SslFactory.class);
+  private static AtomicReference<Exception> watcherExecException = new AtomicReference<>(null);
 
-  private SslFactory() {}
+  public static Optional<Exception> lastLoadFailure() {
```
trnguyencflt, line 37: this is only used in tests, could you annotate with @VisibleForTesting?

core/src/main/java/io/confluent/rest/SslFactoryPemHelper.java
```diff
@@ -0,0 +1,390 @@
...
+    private final Password password;
+    protected final Password keyPassword;
+    private Long fileLastModifiedMs;
+    private KeyStore keyStore;
+    protected final boolean useBcfks;
+
+    FileBasedStore(String type, String path, Password password, Password keyPassword,
+                   boolean isKeyStore, boolean useBcfks) {
+      Objects.requireNonNull(type, "type must not be null");
+      this.type = type;
```
trnguyencflt, line 120: Objects.requireNonNull(type, "type must not be null");
      this.type = type;

can be reduced to this.type = Objects.requireNonNull(type, "type must not be null");

core/src/main/java/io/confluent/rest/SslFactoryPemHelper.java
```diff
@@ -0,0 +1,390 @@
...
+            : new PemStore(storeContents, useBcfks);
+        return pemStore.keyStore;
+      } catch (Exception e) {
+        log.error("Failed to load store, isKeyStore : {}, path {}", isKeyStore, path, e);
+        throw new InvalidConfigurationException("Failed to load PEM SSL keystore " + path, e);
+      }
+    }
+  }
+
+  public static class PemStore implements SecurityStore {
```
trnguyencflt, line 209: The logic in this class and the whole file in general are very complex, can this be simplified by using classes from upstream kafka?

smjn: Currently we are not dependant on this module in rest-utils and the code is part of ce-kafka not ak hence decided to replicate the code directly.

trnguyencflt: > hence decided to replicate the code directly.

there is issue with this method as it will double the maintenance cost, is it possible to extract a module in ce-kafka so that rest-utils can import?

msn-tldr: > there is issue with this method as it will double the maintenance cost, is it possible to extract a module in ce-kafka so that rest-utils can import?

1 challenge that @smjn also alluded is that `rest-utils` is open-source, where-as ce-kafka is not.

@smjn adding to @trnguyencflt why not put the common code in a module in AK, or common or some other open source repo? And `rest-utils` and `ce-kafka` could import it. It reduces duplicate code.

smjn: We can revisit this post FEDRAMP deadlines.

trnguyencflt:
@smjn have you tested this change in a cluster?

smjn:
> @smjn have you tested this change in a cluster?

@trnguyencflt Yes tested on devel GCP cluster using the process you devised.
```
root@kafka-0:/opt/confluent/share/java/ce-kafka-http-server# jar -tvf rest-utils-7.7.0-9999.jar |grep -i pemhelper
  3000 Fri Feb 16 11:30:34 UTC 2024 io/confluent/rest/SslFactoryPemHelper$FileBasedPemStore.class
  2853 Fri Feb 16 11:30:34 UTC 2024 io/confluent/rest/SslFactoryPemHelper.class
  7927 Fri Feb 16 11:30:34 UTC 2024 io/confluent/rest/SslFactoryPemHelper$PemStore.class
   338 Fri Feb 16 11:30:34 UTC 2024 io/confluent/rest/SslFactoryPemHelper$SecurityStore.class
  2810 Fri Feb 16 11:30:34 UTC 2024 io/confluent/rest/SslFactoryPemHelper$PemParser.class
  4509 Fri Feb 16 11:30:34 UTC 2024 io/confluent/rest/SslFactoryPemHelper$FileBasedStore.class
  2813 Fri Feb 16 11:30:34 UTC 2024 io/confluent/rest/SslFactoryPemHelper$SecurityFactory.class
```

omkreddy (approved):
LGTM

core/src/main/java/io/confluent/rest/SslFactoryPemHelper.java
trnguyencflt: @smjn could you please a JIRA ticket for this? Thanks 👍

trnguyencflt (approved):

PR #458, opened 2024-01-19
Set nodeLabel for nightly master build

trnguyencflt:
PR build is with JDK17 while master build is with JDK8 on Jenkins, this creates a gap where we have this test `SslTest.test_WhenCalledWithRestrictedCipher_ConnFails` consistently fails in [Jenkins](https://jenkins.confluent.io/job/confluentinc/job/rest-utils/job/master/9796/console) while it passes in PRs. Attempt to reproduce this failure in local machine is also unsuccessful (it passes with JDK8 on my Macbook).

Therefore, it is better just set the same JDK across PRs and master builds, also we're migrating from Jenkins to Semaphore (which uses JDK17 by default) anyway, so this is to reduce the toll of oncall resolving failed builds.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,5 +8,6 @@ common {
   nanoVersion = true
   pinnedNanoVersions = true
   mvnSkipDeploy = true
+  nodeLabel = 'docker-debian-jdk17'
 }
 //change


msn-tldr:
> Therefore, it is better just set the same JDK across PRs and master builds, also we're migrating from Jenkins to Semaphore
Where was this announced? 
I wonder if semaphore uses jdk17, how would nightly builds cover java 11/8 that are currently supported as well
https://docs.confluent.io/platform/current/installation/versions-interoperability.html#java

trnguyencflt:
> Where was this announced?
There are several announcements about the progress in engineering channel.

> I wonder if semaphore uses jdk17, how would nightly builds cover java 11/8 that are currently supported as well
Interesting question, I think the support means that release engineer will regularly run the package build with Java 8, 11, and 17. About what is used in Semaphore, possibly it isn't related.

soondenana:
> > I wonder if semaphore uses jdk17, how would nightly builds cover java 11/8 that are currently supported as well
> 
> Interesting question, I think the support means that release engineer will regularly run the package build with Java 8, 11, and 17. About what is used in Semaphore, possibly it isn't related.

I think the compilation will happen with target as Java 8, so bytecode will be java8 compatible.

trnguyencflt:
> I think the compilation will happen with target as Java 8, so bytecode will be java8 compatible.

If this is the case then the packaging might fail because it starts at jdk8.

The frustrating thing is that this isn't reproducible locally, so I think what is best is to disable this test for now.
@msn-tldr @soondenana what do you think?

msn-tldr:
@soondenana 
> I think the compilation will happen with target as Java 8, so bytecode will be java8 compatible.

I read it as jenkins using jdk17 would target Java8, as bytecode will compatible with java8, 11, 17.. so on. So this should ok to merge. Is that correct?

soondenana:
> @soondenana
> 
> > I think the compilation will happen with target as Java 8, so bytecode will be java8 compatible.
> 
> I read it as jenkins using jdk17 would target Java8, as bytecode will compatible with java8, 11, 17.. so on. So this should ok to merge. Is that correct?

Yes, here is the relevant snipped in Kafka build file: https://github.com/confluentinc/ce-kafka/blob/master/build.gradle#L429-L444

soondenana:
Just checked rest-utils build too, the compiler setting is fetched from the common project that has `java.version` as 1.8 (https://github.com/confluentinc/common/blob/master/pom.xml#L59) and sets that as source and target for the compiler: https://github.com/confluentinc/common/blob/master/pom.xml#L748-L756

trnguyencflt:
> Just checked rest-utils build too, the compiler setting is fetched from the common project that has `java.version` as 1.8 (https://github.com/confluentinc/common/blob/master/pom.xml#L59) and sets that as source and target for the compiler: https://github.com/confluentinc/common/blob/master/pom.xml#L748-L756

@soondenana @msn-tldr that means it should be fine to use jdk17 for the build process?

msn-tldr (approved):
lgtm

PR #464, opened 2024-02-22
Add doLog param for constructor

trnguyencflt:
Ref: https://github.com/confluentinc/ce-kafka/pull/13010

### DIFF
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -1125,11 +1125,15 @@ public class RestConfig extends AbstractConfig {
 
   private static Time defaultTime = Time.SYSTEM;
 
-  public RestConfig(ConfigDef definition, Map<?, ?> originals) {
-    super(definition, originals);
+  public RestConfig(ConfigDef definition, Map<?, ?> originals, boolean doLog) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -1125,13 +1127,18 @@ private static ConfigDef incompleteBaseConfigDef() {
 
   private static Time defaultTime = Time.SYSTEM;
 
-  public RestConfig(ConfigDef definition, Map<?, ?> originals) {
-    super(definition, originals);
+  public RestConfig(ConfigDef definition, Map<?, ?> originals, boolean doLog) {
```
msn-tldr, line 1130: why not use the config `CONFIGS_LOGGING_ENABLED_PROP` as the original PR proposed?

Using the config has 1 pro, that it can be modified at run-time to toggle logging on/off.

trnguyencflt: CONFIGS_LOGGING_ENABLED_PROP is internal (ce-kafka only), so it will be used in upstream of this project (ce-kafka-rest and ce-kafka-http-server).

msn-tldr (approved):

PR #468, opened 2024-02-28
Remove test jar dependency from fips-tests module

trnguyencflt:
We have problem in CP packaging that fips-tests module needs rest-utils jar in order to compile, but in CP packaging rest-utils is compiled with `mvn -Dmaven.test.skip=true`, so no test jar is produced (because we don't release package with test jar), which breaks the build.

This PR breaks the deadlock by removing test jar dependency from fips-tests with the price of we have similar test code between `SslFactoryTest` and `SslFactoryFipsTest`.

However, there is no much we can do on this as fips and non-fips cannot be used together in a module.

### DIFF
--- a/fips-tests/src/test/java/io/confluent/rest/SslFactoryFipsTest.java
+++ b/fips-tests/src/test/java/io/confluent/rest/SslFactoryFipsTest.java
@@ -16,10 +16,19 @@
 
 package io.confluent.rest;
 
+import java.io.FileWriter;
+import java.security.KeyStore;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
 import org.apache.kafka.common.config.types.Password;
+import org.apache.kafka.common.errors.InvalidConfigurationException;
+import org.apache.kafka.test.TestUtils;
 import org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider;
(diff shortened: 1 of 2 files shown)


smjn (approved):
LGTM

Zhongyin-Zhang (approved):
LGTM

ehumber (approved):

fips-tests/src/test/java/io/confluent/rest/SslFactoryFipsTest.java
```diff
@@ -72,23 +91,144 @@ public void setUp() {
...
-  @Override
   protected String getEncryptedKey() {
     return KEY;
   }
 
   protected Password getKeyPassword() {
     return null;
   }
+
+  @Test
```
msn-tldr, line 111: @trnguyencflt 
My understanding is that both`SslFactoryTest.java` and `SslFactoryFipsTest.java` need to be kept in sync in terms of test-cases, is that correct? If true, can you add this as a comment in each.

trnguyencflt: yup 👍 however, the PR is already merged, I'll add comment in the next PR 👍

msn-tldr: works, thanks!

PR #472, opened 2024-03-13
FEDRAMP-5315 Support HTTP/2 for BouncyCastle FIPS enabled

trnguyencflt:


core/src/main/java/io/confluent/rest/alpn/server/BouncyCastleServerALPNProcessor.java
```diff
@@ -0,0 +1,35 @@
...
+
+package io.confluent.rest.alpn.server;
+
+import com.google.auto.service.AutoService;
+import javax.net.ssl.SSLEngine;
+import org.eclipse.jetty.alpn.java.server.JDK9ServerALPNProcessor;
+import org.eclipse.jetty.io.ssl.ALPNProcessor;
+
+/**
+ * A server ALPN processor that works with BouncyCastle's JSSE provider.
```
soondenana, line 25: nit: Can you please copy some of the description/background from PR description here. Then we don't have to look back at the PR to figure out details of what is going on.

soondenana (approved):
Thanks for adding this

core/src/main/java/io/confluent/rest/alpn/server/BouncyCastleServerALPNProcessor.java
trnguyencflt: Done 👍

trnguyencflt:
@soondenana could you please give a thumb up to this again, I pushed a new commit for updating comment, so the approval is lost. Thanks 😃

soondenana (approved):

PR #477, opened 2024-03-22
FEDRAMP-5315 Support HTTP/2 for BouncyCastle FIPS enabled (#472)

trnguyencflt:
It has been found that with BouncyCastle (bc) SSL provider FIPS driver, Jetty HTTP2 isn't working because there isn't ALPNProcessor.Server service available.

```
java.lang.IllegalStateException: Connection rejected: No ALPN Processor for org.bouncycastle.jsse.provider.ProvSSLEngine_9 from [org.eclipse.jetty.alpn.java.server.JDK9ServerALPNProcessor@36def543]
at org.eclipse.jetty.alpn.server.ALPNServerConnectionFactory.newServerConnection(ALPNServerConnectionFactory.java:111)
at org.eclipse.jetty.server.NegotiatingServerConnectionFactory.newConnection(NegotiatingServerConnectionFactory.java:108)
at org.eclipse.jetty.server.SslConnectionFactory.newConnection(SslConnectionFactory.java:131)
at org.eclipse.jetty.server.ServerConnector$ServerConnectorManager.newConnection(ServerConnector.java:618)
at org.eclipse.jetty.io.ManagedSelector.createEndPoint(ManagedSelector.java:386)
at org.eclipse.jetty.io.ManagedSelector.access$2100(ManagedSelector.java:65)
at org.eclipse.jetty.io.ManagedSelector$Accept.run(ManagedSelector.java:890)
at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:883)
at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1034)
at java.base/java.lang.Thread.run(Thread.java:829)
```

This has also been reported in [bc-java](https://github.com/bcgit/bc-java/discussions/1226#discussion-4372175).

However, bc-java already [has support for ALPN](https://github.com/bcgit/bc-java/commit/977a0e4e484afc3a5300724688ba31c96710187a), so the main reason it isn't working is that there isn't an implementation of `ALPNProcessor.Server` available in Jetty that uses BC ALPN.

As [suggested](https://github.com/bcgit/bc-java/discussions/1226#discussioncomment-8481568) by the maintainer of bc-java. This PR takes the approach of leveraging `JDK9ServerALPNProcessor` implementation to make a service, BouncyCastleServerALPNProcessor, that is compatible with bc SSL provider.

Note that in [previous PR](https://github.com/confluentinc/rest-utils/pull/472), we extended `JDK9ServerALPNProcessor` class, but making `jetty-alpn-java-server` to provided scope, which affects downstream projects, for this PR, we make the dependency at compile scope, which should resolve the issue downstream.

### DIFF
--- a/core/src/main/java/io/confluent/rest/alpn/server/BouncyCastleServerALPNProcessor.java
+++ b/core/src/main/java/io/confluent/rest/alpn/server/BouncyCastleServerALPNProcessor.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2024 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.confluent.rest.alpn.server;
+
+import com.google.auto.service.AutoService;
+import javax.net.ssl.SSLEngine;
+import org.eclipse.jetty.alpn.java.server.JDK9ServerALPNProcessor;
+import org.eclipse.jetty.io.ssl.ALPNProcessor;
+
+/**
+ * This class implements Server ALPN processor and is available as a service so that Jetty http/2
+ * can work with BouncyCastle's JSSE provider FIPS driver.
+ * Support for ALPN in BouncyCastle's JSSE
+ * provider is available since bc-fips-1.0.2, but there isn't an implementation of
+ * ALPNProcessor.Server available in Jetty server.
+ * This class leverages JDK9ServerALPNProcessor in
+ * Jetty server to make a service that implements ALPNProcessor.Server and is compatible with
+ * BouncyCastle's JSSE provider FIPS driver.
+ */
+@SuppressWarnings("checkstyle:AbbreviationAsWordInName")
+@AutoService(ALPNProcessor.Server.class)
+public class BouncyCastleServerALPNProcessor extends JDK9ServerALPNProcessor {
+
+  @Override
+  public boolean appliesTo(SSLEngine sslEngine) {
+    return sslEngine.getClass().getName().startsWith("org.bouncycastle.jsse.provider");
+  }
+}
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/confluent/rest/alpn/server/BouncyCastleServerALPNProcessor.java
```diff
@@ -0,0 +1,110 @@
...
+import org.eclipse.jetty.util.JavaVersion;
+import org.eclipse.jetty.util.log.Log;
+import org.eclipse.jetty.util.log.Logger;
+
+/**
+ * This class implements Server ALPN processor and is available as a service so that Jetty http/2
+ * can work with BouncyCastle's JSSE provider FIPS driver. Support for ALPN in BouncyCastle's JSSE
+ * provider is available since bc-fips-1.0.2, but there isn't an implementation of
+ * ALPNProcessor.Server available in Jetty server. The implementation of this class is similar to
+ * JDK9ServerALPNProcessor in Jetty server to make a service that implements ALPNProcessor.Server
```
msn-tldr, line 37: can you add a link to the version of JDK9ServerALPNProcessor that inspires this implementation, to the comments? 

That would be helpful for future modifications.

core/src/main/java/io/confluent/rest/alpn/server/BouncyCastleServerALPNProcessor.java
```diff
@@ -0,0 +1,110 @@
+/*
```
ehumber, line 1: If this is copied from somewhere, I think we need to credit the original source.  And also check the license on that original source file to ensure it's not eg GPL.

ehumber (changes requested):
See the comment about licensing and the original source of the code.

core/src/main/java/io/confluent/rest/alpn/server/BouncyCastleServerALPNProcessor.java
trnguyencflt: it is Eclipse Public License v1.0 https://www.tldrlegal.com/license/eclipse-public-license-1-0-epl-1-0, similar to GPL, so a bit more complicated.

trnguyencflt: changed to use class inheritance method, so there is no license concern.

core/src/main/java/io/confluent/rest/alpn/server/BouncyCastleServerALPNProcessor.java
trnguyencflt: I've changed to class extension method, so we don't need to link any more 👍

msn-tldr:
@trnguyencflt can you add in a comment how did you get around the license-ing issue?

ehumber (approved):

PR #483, opened 2024-05-30
AMF-991: Support for max.connection.duration.ms

niteshmor:
This pull request adds support for terminating long running client connections after a specified duration.

### DIFF
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -478,6 +478,13 @@ public class RestConfig extends AbstractConfig {
           + "If the limit is set to a non-positive number, no limit is applied. Default is 0.";
   private static final int SERVER_CONNECTION_LIMIT_DEFAULT = 0;
 
+  public static final String MAX_CONNECTION_DURATION_MS = "max.connection.duration.ms";
+  public static final String MAX_CONNECTION_DURATION_MS_DOC =
+      "The maximum duration in milliseconds that a connection can be open. "
+          + "If a connection is open for longer than this duration, it will be closed. "
+          + "If set to 0, no limit is applied. Default is 0.";
+  protected static final long MAX_CONNECTION_DURATION_MS_DEFAULT = 0;
+
   // For rest-utils applications connectors correspond to configured listeners. See
   // ApplicationServer#parseListeners for more details.
   private static final String CONNECTOR_CONNECTION_LIMIT = "connector.connection.limit";
@@ -1050,6 +1057,11 @@ public class RestConfig extends AbstractConfig {
             SERVER_CONNECTION_LIMIT_DEFAULT,
             Importance.LOW,
             SERVER_CONNECTION_LIMIT_DOC
+        ).define(MAX_CONNECTION_DURATION_MS,
+            Type.LONG,
+            MAX_CONNECTION_DURATION_MS_DEFAULT,
+            Importance.LOW,
+            MAX_CONNECTION_DURATION_MS_DOC
         ).define(
             CONNECTOR_CONNECTION_LIMIT,
             Type.INT,
@@ -1256,6 +1268,10 @@ public class RestConfig extends AbstractConfig {
 
   public final int getServerConnectionLimit() {
     return getInt(SERVER_CONNECTION_LIMIT);
+  }
+
+  public final long getMaxConnectionDuration() {
+    return getLong(MAX_CONNECTION_DURATION_MS);
   }
 
   public final int getConnectorConnectionLimit() {
--- a/core/src/main/java/io/confluent/rest/filters/ConnectionDurationFilter.java
+++ b/core/src/main/java/io/confluent/rest/filters/ConnectionDurationFilter.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2024 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.confluent.rest.filters;
+
+import io.confluent.rest.RestConfig;
+import java.io.IOException;
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.eclipse.jetty.io.Connection;
+import org.eclipse.jetty.server.HttpChannel;
+import org.eclipse.jetty.server.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Simple server-side request filter that limits the connection age.
+ *
+ *
+ * <p>Long-running client connections can be problematic for multiple server instances behind
+ * an NLB due to uneven load distribution (especially in case of HTTP/2.0 that specifically
+ * encourages long-lived connections). This filter closes any connection that receives a request
+ * after the connection has been open for {@link RestConfig#MAX_CONNECTION_DURATION_MS} ms.</p>
+ */
+public class ConnectionDurationFilter implements Filter {
+
+  private static final Logger log = LoggerFactory.getLogger(ConnectionDurationFilter.class);
+  private static long MAX_CONNECTION_DURATION_MS = -1;
+
+  @Override
+  public void init(FilterConfig filterConfig) throws ServletException {
+    String maxConnectionDuration = filterConfig.getInitParameter(
+        RestConfig.MAX_CONNECTION_DURATION_MS);
+    if (maxConnectionDuration != null) {
+      MAX_CONNECTION_DURATION_MS = Long.parseLong(maxConnectionDuration);
+    }
+  }
+
+  @Override
+  public void doFilter(
+    ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain)
+    throws IOException, ServletException {
+
+    HttpServletResponse httpServletResponse = (HttpServletResponse) servletResponse;
+
+    HttpChannel channel = ((Response) httpServletResponse).getHttpChannel();
+
+    long connectionCreationTime = channel.getConnection().getCreatedTimeStamp();
+    long connectionAge = System.currentTimeMillis() - connectionCreationTime;
+
+    if (connectionAge > MAX_CONNECTION_DURATION_MS) {
+      log.debug("Connection from remote peer {} has been active for {}ms. Closing the connection.",
+          channel.getRemoteAddress(), connectionAge);
+      channel.getEndPoint().close();
+    } else {
+      log.trace("Connection from remote peer {} is {}ms old. Leaving the connection as is",
+          channel.getRemoteAddress(), connectionAge);
+    }
+
+    filterChain.doFilter(servletRequest, servletResponse);
+
+  }
+
+  @Override
+  public void destroy() {
+
+  }
+
+}
--- a/core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
+++ b/core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
@@ -1,0 +1,225 @@
+/*
+ * Copyright 2024 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.confluent.rest;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.google.common.collect.ImmutableMap;
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.Configurable;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response.Status;
+import javax.ws.rs.core.UriBuilder;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.impl.conn.BasicHttpClientConnectionManager;
+import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
+import org.apache.http.util.EntityUtils;
+import org.eclipse.jetty.server.Server;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+
+/*
+ *  Note on client support --
+ *  The Apache HttpClient library v4.x does not support HTTP/2.0, so we use
+ *  OkHttp for HTTP/2.0 tests.
+ */
+public class ConnectionDurationFilterTest {
+
+  static Server server;
+  private static final long MAX_CONNECTION_DURATION_MS = 100;
+
+  @BeforeAll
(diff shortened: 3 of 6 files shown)


core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -478,6 +478,13 @@ public class RestConfig extends AbstractConfig {
           + "If the limit is set to a non-positive number, no limit is applied. Default is 0.";
   private static final int SERVER_CONNECTION_LIMIT_DEFAULT = 0;
 
+  public static final String MAX_CONNECTION_DURATION_MS = "max.connection.duration.ms";
+  public static final String MAX_CONNECTION_DURATION_MS_DOC =
+      "The maximum duration in milliseconds that a connection can be open. "
```
ehumber, line 483: Is it worth clarifying this is the TCP connection the requests come over, to help differentiate this config from https://github.com/confluentinc/rest-utils/blob/a04ed80fe446463302f9a1d88c2d4ee4de589ae5/core/src/main/java/io/confluent/rest/RestConfig.java#L444 ?

core/src/main/java/io/confluent/rest/filters/ConnectionDurationFilter.java
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * <p>Long-running client connections can be problematic for multiple server instances behind
+ * an NLB due to uneven load distribution (especially in case of HTTP/2.0 that specifically
+ * encourages long-lived connections). This filter closes any connection that receives a request
+ * after the connection has been open for {@link RestConfig#MAX_CONNECTION_DURATION_MS} ms.</p>
+ */
+public class ConnectionDurationFilter implements Filter {
+
+  private static final Logger log = LoggerFactory.getLogger(ConnectionDurationFilter.class);
+  private static long MAX_CONNECTION_DURATION_MS = -1;
```
trnguyencflt, line 45: is there a reason why MAX_CONNECTION_DURATION_MS is static?

core/src/main/java/io/confluent/rest/filters/ConnectionDurationFilter.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  @Override
+  public void doFilter(
+      ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain)
+      throws IOException, ServletException {
+
+    HttpServletResponse httpServletResponse = (HttpServletResponse) servletResponse;
+
+    HttpChannel channel = ((Response) httpServletResponse).getHttpChannel();
```
trnguyencflt, line 63: we use several unchecked cast here, are they all safe?

core/src/main/java/io/confluent/rest/filters/ConnectionDurationFilter.java
```diff
@@ -0,0 +1,86 @@
...
+
+    long connectionCreationTime = channel.getConnection().getCreatedTimeStamp();
+    long connectionAge = System.currentTimeMillis() - connectionCreationTime;
+
+    if (connectionAge > MAX_CONNECTION_DURATION_MS) {
+      log.debug("Connection from remote peer {} has been active for {}ms. Closing the connection.",
+          channel.getRemoteAddress(), connectionAge);
+      channel.getEndPoint().close();
```
trnguyencflt, line 71: should we also use the filterChain to stop processing?

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
```diff
@@ -0,0 +1,225 @@
...
+  private static final long MAX_CONNECTION_DURATION_MS = 100;
+
+  @BeforeAll
+  public static void setUp() throws Exception {
+    MyAddressApplication application =
+        new MyAddressApplication(
+            new MyAddressConfig(
+                ImmutableMap.of(
+                    "listeners", "http://localhost:8080",
+                    "http2.enabled", "true",
```
trnguyencflt, line 68: is this applicable for http1.1?

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
```diff
@@ -0,0 +1,225 @@
...
+    CloseableHttpClient client = createPersistentClient();
+    List<String> addressList = new ArrayList<>();
+
+    // Send a few requests with a delay between consecutive requests that is longer
+    // than the connection duration threshold.
+    // Each request should succeed and the server should return a different address for each request.
+    for (int i=0; i<5; i++) {
+
+      HttpGet request = createRequest(server.getURI());
+      assertEquals("HTTP/1.1", request.getProtocolVersion().toString());
```
trnguyencflt, line 92: why the request is http1.1 when your server has http2 enabled?

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
trnguyencflt, line 1: I don't really see the test where the client connection is terminated due to exceeding max duration, could you please give some pointeR?

core/src/main/java/io/confluent/rest/filters/ConnectionDurationFilter.java
```diff
@@ -0,0 +1,86 @@
...
+import org.eclipse.jetty.server.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Simple server-side request filter that limits the connection age.
+ *
+ *
+ * <p>Long-running client connections can be problematic for multiple server instances behind
+ * an NLB due to uneven load distribution (especially in case of HTTP/2.0 that specifically
```
msn-tldr, line 38: Even HTTP 1.1 uses Persistent connections by default. So this could be helpful there.

core/src/main/java/io/confluent/rest/filters/ConnectionDurationFilter.java
msn-tldr: +1

core/src/main/java/io/confluent/rest/RestConfig.java
msn-tldr: CMIIW. As i understand, this config only applies for connections that are not-idle, i.e i.e. receiving http requests. So should this be renamed to MAX_NON_IDLE_TCP_CONNECTION_DURATION_MS ?
And a comment explaining that for idle connections MAX_IDLE_TIMEOUT_MS should be configured appropriately.

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
```diff
@@ -0,0 +1,225 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+
+/*
+ *  Note on client support --
+ *  The Apache HttpClient library v4.x does not support HTTP/2.0, so we use
+ *  OkHttp for HTTP/2.0 tests.
+ */
+public class ConnectionDurationFilterTest {
```
msn-tldr, line 56: I don't see a test where HTTP requests fail as long running TCP connection closes unexpectedly. Is that going to be added?

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
```diff
@@ -0,0 +1,225 @@
...
+
+  @AfterAll
+  public static void tearDown() throws Exception {
+    server.stop();
+  }
+
+  @Test
+  public void connectionDurationTest_ApacheHttpClient_Ephemeral_HTTP1() throws Exception {
+
+    CloseableHttpClient client = createPersistentClient();
```
msn-tldr, line 83: Should this be creating an ephemeral client given the test is called ephemeral? Naming should be consistent.

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
```diff
@@ -0,0 +1,225 @@
...
+    client.close();
+
+    // check that we actually used 5 different addresses under the hood
+    assertEquals(5, addressList.stream().distinct().count());
+  }
+
+  @Test
+  public void connectionDurationTest_ApacheHttpClient_Persistent_HTTP1() throws Exception {
+
+    CloseableHttpClient client = createEphemeralClient();
```
msn-tldr, line 113: Naming should be consistent, similar to comment on line 83.

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
```diff
@@ -0,0 +1,225 @@
...
+
+      CloseableHttpResponse response = client.execute(request);
+      assertEquals(Status.OK.getStatusCode(), response.getStatusLine().getStatusCode());
+
+      String responseBody = EntityUtils.toString(response.getEntity());
+      assertTrue(responseBody.startsWith("127.0.0.1:"));
+
+      response.close();
+      addressList.add(responseBody);
+      Thread.sleep(MAX_CONNECTION_DURATION_MS + 10);
```
msn-tldr, line 132: Going by the test name, is this test creating "persistent" tcp connections? So should the request fail given the connection remains idle for longer than MAX_CONNECTION_DURATION_MS

core/src/main/java/io/confluent/rest/RestConfig.java
niteshmor: Good points on documentation update.  Will fix. 

On the naming of the config item, `MAX_NON_IDLE_TCP_CONNECTION_DURATION_MS` could be confusing -- one might wonder whether the counter start when the connection is non idle and gets reset for idle connection, and all sorts of similar things.   The config is an absolute maximum limit for how long a single connection is allowed to be open, regardless of whether that's idle or not.

core/src/main/java/io/confluent/rest/filters/ConnectionDurationFilter.java
niteshmor: Typo.  will fix.
thanks for pointing out.

core/src/main/java/io/confluent/rest/filters/ConnectionDurationFilter.java
niteshmor: Good point.  The typecasts in general should be safe given that rest-utils is specifically targeting the HTTP request serving, but I will add some checks just to be on the safe side and to avoid any issues in the future.

core/src/main/java/io/confluent/rest/filters/ConnectionDurationFilter.java
niteshmor: Not sure I know how to do it.  The "don't keep processing the request" sounds reasonable (which is what I think you are referring to), but unfortunately Filterchain only exposes a `doFilter` method.   
Do you have some suggestion on how to go about it?

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
niteshmor: yes.  
"http2.enabled" simply says whether http 2.0 is supported or not.
http 1 is supported by default

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
niteshmor: That's a limitation of the Apache http client. 
HTTP client version 4.x only supports http 1.x.
HTTP 2.x support was added in version 5.x (see [here](https://stackoverflow.com/a/59566041)).

Using 5.x http client would require a major version upgrade of that package  -- there are a number of major changes (including relocated packages, etc) making 4.x -> 5.x migration non-trivial  (probably a separate future pull request on its own).

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
niteshmor: The current tests echo back the client address (IP address + client-port), which is a proxy for the connection itself. 
If the same connection is reused, then the (IP address + client-port) pair stays the same across the requests. 
If a new connection is established, the client-port changes. 
The current tests try to make a number of calls with a delay longer than the "max connection duration", and assert that a new (IP address + client-port) was used for each of them.

I now see how that is somewhat obtuse, and all this explanation probably needs to make it to the test itself as a comment.  What do you think?  Does that make sense and address the need for the tests?   

On why I had to choose this complicated roundabout way of verifying that the connections are terminated -- http clients (both apache http client and okhttp client) don't really have any public interface to give socket level information back to the user.  Both of them like to use internal mechanisms to maintain connection pools or attempt retries without necessarily exposing what's happening at the connection level back to the user. That's where I ended up conjuring up a server application that echoes back the (client IP + client-port) pair. 

I should also add a case when multiple requests are made all within the "max connection duration", and show that they indeed are using the same connection under the hood.

core/src/main/java/io/confluent/rest/filters/ConnectionDurationFilter.java
niteshmor: good point. will clarify with a documentation update

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
niteshmor: Interestingly, the HTTP requests don't fail if the connection closes explicitly. 
Both apache http client as well as okhttp client are smart enough to see that the TCP connection is closed and they attempt to open a new connection under the hood.  That's really showing that the "graceful termination" actually worked. 

In an earlier implementation attempt for this feature, I tried to close the network socket (by just calling `.close()`), which certainly was _not_ graceful termination and I saw that the client spewed out an error along the lines of "I tried to send a request but the server didn't respond". 

Also see -- https://github.com/confluentinc/rest-utils/pull/483#discussion_r1621935209

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
niteshmor: Apologies. It's a typo.  will filx

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
niteshmor: Apologies. It's a typo.  will filx

core/src/main/java/io/confluent/rest/RestConfig.java
msn-tldr: > The config is an absolute maximum limit for how long a single connection is allowed to be open, regardless of whether that's idle or not.

is this correct though? 

Jetty `Filter` is run on a request, or a response. So if connection is idle, i.e. no request or response, this filter won't run. So only for non-idle connection this config would work. That's why i suggested above(quoted below) that for idle connections `MAX_IDLE_TIMEOUT_MS` should be set.
> And add to the comment that for idle connections MAX_IDLE_TIMEOUT_MS should be configured appropriately.


re. naming, I am ok with the current name as naming is hard 😃 . As long the comment makes the right behaviour explicit.

core/src/main/java/io/confluent/rest/filters/ConnectionDurationFilter.java
msn-tldr: You can do that by not calling `filterChain.doFilter`, i.e. return early in if block at line 68.

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
msn-tldr: Merging the other comment https://github.com/confluentinc/rest-utils/pull/483/files#r1621942383 here as they are similar.

> I now see how that is somewhat obtuse, and all this explanation probably needs to make it to the test itself as a comment.

+1

> Both of them like to use internal mechanisms to maintain connection pools or attempt retries without necessarily exposing what's happening at the connection level back to the user.

Is it possible to set the http retries to 0 in http-clients? So that client only sends HTTP request once, max_duration exceeds -> connection closed, and request fails. As retries are 0, the error/exception is passed back to the test.

It would be gr8 to have the behaviour(connection disconnects -> test sees error) explicit in test, as that is much easier to reason in future.

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
trnguyencflt: shall we make the tests parameterized on http1.1 and http2

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
trnguyencflt: I think we have jetty http client code example across this repo. You might want to take a look.

core/src/test/java/io/confluent/rest/ConnectionDurationFilterTest.java
trnguyencflt: I think you can test this indirectly with client sending multiple requests for a span of time, and since the connection exceed the max time, the remaining requests will not complete.

niteshmor: Yeah setting retries =>0  sounds like a good idea.  
Let me rework the tests a bit incorporating all the feedback, thank you for all the suggestions. 

I also realized that the "ephemeral client" is not really an ephemeral client.  This snippet of code that I got from some other tests in the repo doesn't actually create ephemeral clients.  Setting that to "-1" essentially means "No suggestion on keepalive strategy" ([ref](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/conn/ConnectionKeepAliveStrategy.java#L42-L66)).  🤦 
```
private static CloseableHttpClient createEphemeralClient() {
    return HttpClients.custom()
        .setConnectionManager(new BasicHttpClientConnectionManager())
        .setKeepAliveStrategy((httpResponse, httpContext) -> -1)
        .build();
  }
```

core/src/main/java/io/confluent/rest/RestConfig.java
niteshmor: Ah yes, you are right on the caveat of idle connections.  My narrow-minded focus for adding this feature was specifically for very active connections that last days/weeks.  
A combination of "idle timeout" + this "max connection duration (reworded differently)" should go hand-in-hand for any config tuning. 
I will update the documentation to reflect that.

trnguyencflt:
@niteshmor thanks for the update 👍

PR #503, opened 2024-09-17
changing order for filter configuration [CPSEC-482]

ujjwalkalia:
If we add a new jetty filter holder to the filter chain through `applyCustomConfiguration` , it gets configured after the jersey servlet as the jersey servlet is configured first. Due to this the jetty filter is not triggered on incoming requests.

This PR swaps the order of configuring customConfiguration and jersey servlet, so that requests can be routed as expected.

### DIFF
--- a/core/src/test/java/io/confluent/rest/CustomInitTest.java
+++ b/core/src/test/java/io/confluent/rest/CustomInitTest.java
@@ -21,6 +21,7 @@ import static javax.ws.rs.core.Response.Status.OK;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.fail;
 
+import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


nsharma-git (approved):

ujjwalkalia:
> Could you please add a test to show that REST_SERVLET_INITIALIZERS_CLASSES_CONFIG is applied?

there is already a test for that :-
io/confluent/rest/CustomInitTest.java

trnguyencflt:
> > Could you please add a test to show that REST_SERVLET_INITIALIZERS_CLASSES_CONFIG is applied?
> 
> there is already a test for that :- io/confluent/rest/CustomInitTest.java

why the test was working before even with wrong order?

core/src/test/java/io/confluent/rest/CustomInitTest.java
trnguyencflt, line 1: I don't quite understand the test, should it be related to `REST_SERVLET_INITIALIZERS_CLASSES_CONFIG` config? Could you please help clarify this? Thanks 👍

ujjwalkalia: @trnguyencflt the flow of the test is :-
1. Configures a rest utils based application.
2. Injects `CustomRestInitializer` through `REST_SERVLET_INITIALIZERS_CLASSES_CONFIG`
3. `CustomRestInitializer` adds a jetty based filter to the context `CustomJettyFilter`
4.  The test validates whether `CustomJettyFilter` was triggered as part of the request flow through the application.

trnguyencflt: @ujjwalkalia thanks for the explanation, makes sense 👍

trnguyencflt (approved):

PR #478, opened 2024-04-01
Downgrade jetty package

shubh-ranade:
Downgrade jetty to `9.4.53`

Commons jetty version was upgraded to `9.4.54` recently: https://github.com/confluentinc/common/pull/582 which affects userId/tenant based DoS filters in this repo.

Dependencies after the downgrade:
```
❯ mvn dependency:tree -Dincludes=org.eclipse.jetty
[INFO] Scanning for projects...
[INFO] ------------------------------------------------------------------------
[INFO] Detecting the operating system and CPU architecture
[INFO] ------------------------------------------------------------------------
[INFO] os.detected.name: osx
[INFO] os.detected.arch: aarch_64
[INFO] os.detected.bitness: 64
[INFO] os.detected.version: 14.3
[INFO] os.detected.version.major: 14
[INFO] os.detected.version.minor: 3
[INFO] os.detected.classifier: osx-aarch_64
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Build Order:
[INFO]
[INFO] rest-utils-parent                                                  [pom]
[INFO] rest-utils                                                         [jar]
[INFO] rest-utils-test                                                    [jar]
[INFO] rest-utils-example                                                 [jar]
[INFO] rest-utils-package                                                 [pom]
[INFO] rest-utils-fips-tests                                              [jar]
[INFO]
[INFO] -------------------< io.confluent:rest-utils-parent >-------------------
[INFO] Building rest-utils-parent 7.8.0-0                                 [1/6]
[INFO]   from pom.xml
[INFO] --------------------------------[ pom ]---------------------------------
[INFO]
[INFO] --- dependency:3.3.0:tree (default-cli) @ rest-utils-parent ---
[INFO]
[INFO] ----------------------< io.confluent:rest-utils >-----------------------
[INFO] Building rest-utils 7.8.0-0                                        [2/6]
[INFO]   from core/pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[INFO]
[INFO] --- dependency:3.3.0:tree (default-cli) @ rest-utils ---
[INFO] io.confluent:rest-utils:jar:7.8.0-0
[INFO] +- org.eclipse.jetty.websocket:javax-websocket-server-impl:jar:9.4.53.v20231009:compile
[INFO] |  \- org.eclipse.jetty:jetty-annotations:jar:9.4.53.v20231009:compile
[INFO] |     +- org.eclipse.jetty:jetty-plus:jar:9.4.53.v20231009:compile
[...]

PR #505, opened 2024-10-28
bump: jetty servlets version

sumit-confluent:
This fixes [CONMON-10491](https://jira.confluentgov.com/browse/CONMON-10491)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,7 +58,7 @@
         <checkstyle.suppressions.location>checkstyle/suppressions.xml</checkstyle.suppressions.location>
         <io.confluent.rest-utils.version>8.0.0-0</io.confluent.rest-utils.version>
         <conscrypt.version>2.5.2</conscrypt.version>
-        <jetty.servlets.version>9.4.53.v20231009</jetty.servlets.version>
+        <jetty.servlets.version>9.4.54</jetty.servlets.version>
     </properties>
 
     <repositories>


pom.xml
```diff
@@ -58,7 +58,7 @@
         <checkstyle.suppressions.location>checkstyle/suppressions.xml</checkstyle.suppressions.location>
         <io.confluent.rest-utils.version>8.0.0-0</io.confluent.rest-utils.version>
         <conscrypt.version>2.5.2</conscrypt.version>
-        <jetty.servlets.version>9.4.53.v20231009</jetty.servlets.version>
+        <jetty.servlets.version>9.4.54.v20240208</jetty.servlets.version>
```
trnguyencflt, line 61: this is a bit weird why we need to pin jetty.servlets.version here, there is already a version from common https://github.com/confluentinc/common/blob/5eb23708ee9bd1e7c78204da53a8bd976b3b464e/pom.xml#L66 that we can use, I would suggest to remove the pin version here and use the one from common

sumit-confluent: removed the pinned version

pom.xml
```diff
@@ -196,7 +196,7 @@
             <dependency>
                 <groupId>org.eclipse.jetty</groupId>
                 <artifactId>jetty-servlets</artifactId>
-                <version>${jetty.servlets.version}</version>
+                <version>${jetty.version</version>
```
trnguyencflt, line 199: @sumit-confluent please fix the missing bracket 
```suggestion
                <version>${jetty.version}</version>
```

sumit-confluent: thanks for pointing out!! added the missing bracket

trnguyencflt:
@sumit-confluent unfortunately, there is a reason why we pin the version of jetty-servers, it is due to the newer version breaking DoSFilter https://github.com/confluentinc/rest-utils/pull/478, could you please raise this in kora-networking-sme for further action? Thanks 👍

PR #506, opened 2024-11-01
Log unexpected SNI headers in REST

bob-barrett:
Add support for a list of expected SNI headers, configured using `expected.sni.headers`. If this is present, and the SNI check is not enabled, it logs a warning when an unexpected SNI header is present on a connection. This can be used prior to enabling the SNI check to validate that the check will not be disruptive.

### DIFF
--- a/core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java
+++ b/core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java
@@ -1,0 +1,38 @@
+package io.confluent.rest.handlers;
+
+import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.server.handler.HandlerWrapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.List;
+
+import static org.eclipse.jetty.http.HttpStatus.Code.MISDIRECTED_REQUEST;
+
+public class ExpectedSniHandler extends HandlerWrapper {
+    private static final Logger log = LoggerFactory.getLogger(SniHandler.class);
+    private final List<String> expectedSniHeaders;
+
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java
```diff
@@ -0,0 +1,51 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.List;
+
+public class ExpectedSniHandler extends HandlerWrapper {
+  private static final Logger log = LoggerFactory.getLogger(SniHandler.class);
```
ehumber, line 31: ExpectedSniHandler, not SniHandler

core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  @Override
+  public void handle(String target, Request baseRequest,
+      HttpServletRequest request,
+      HttpServletResponse response) throws IOException, ServletException {
+    String sniServerName = SniUtils.getSniServerName(baseRequest);
+    if (sniServerName == null) {
+      log.warn("No SNI header present on request");
+    } else if (!expectedSniHeaders.contains(sniServerName)) {
```
ehumber, line 45: sniServerName is the "fully qualified DNS hostname of the server" according to the javadoc for SNIHostName. 

EXPECTED_SNI_HEADERS_DOC =
      "Comma-separated list of expected SNI headers for incoming connection
      
As far as I can see, you're checking the value of the hostname header, against a list of expected headers.  What am I missing? :)

bob-barrett: Yeah, that's right. This allows us to identify if a client is sending an SNI header that we don't expect, without actually blocking the the connection, purely for informational purposes.

ehumber (approved):

msn-tldr (approved):

PR #515, opened 2024-12-03
KNET-14907: Add DoSFilter from Jetty 9.4.53

mattwong949:
`jetty-servlets` package is currently pinned to 9.4.53 since there was a breaking change in the `DoSFilter`. We want to bump the servlets package to a higher version, so we can copy the DoSFilter from 9.4.53 as a short-term solution to allow the version bump w/o incurring the change in behavior.

There will be follow-up PR(s) to bump the jetty versions

### DIFF
--- a/core/src/main/java/io/confluent/rest/DoSFilter.java
+++ b/core/src/main/java/io/confluent/rest/DoSFilter.java
@@ -1,3 +1,7 @@
+/*
+ * Copyright 2024 Confluent Inc.
+ */
+
 package io.confluent.rest;
 
 import java.io.IOException;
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/confluent/rest/DoSFilter.java
trnguyencflt, line 1: Should we copy this into a separate package so it is easier to track

ehumber (approved):
Approved, but I agree with Truc's comment about a separate package

core/src/main/java/io/confluent/rest/DoSFilter.java
msn-tldr: +1

msn-tldr (approved):

ehumber (approved):

trnguyencflt (approved):

PR #516, opened 2024-12-05
KNET-16654: Handle jetty 11+ behavior of NetworkTrafficServerConnector

mattwong949:
There is a breaking change in NetworkTrafficServerConnector class from accepting multiple NetworkTrafficListener (jetty 9) to only accept a single listener (jetty 11+)

We can split the list of connectors into two to differentiate the connectors between metrics and rate limit listeners.

We will also need to change the API call from `addNetworkTrafficListener` to `setNetworkTrafficListener` when we bump the jetty version

References:
[jetty 9 doc](https://javadoc.jetty.org/jetty-9/org/eclipse/jetty/server/NetworkTrafficServerConnector.html)
[jetty 12 doc](https://javadoc.jetty.org/jetty-12/org/eclipse/jetty/server/NetworkTrafficServerConnector.html)

### DIFF
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -71,7 +71,8 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
 
   private static volatile int threadPoolRequestQueueCapacity;
 
-  private final List<NetworkTrafficServerConnector> connectors = new ArrayList<>();
+  private final List<NetworkTrafficServerConnector> metricsListenerConnectors = new ArrayList<>();
+  private final List<NetworkTrafficServerConnector> rateLimitListenerConnectors = new ArrayList<>();
   private final List<NamedURI> listeners;
 
   private static final Logger log = LoggerFactory.getLogger(ApplicationServer.class);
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -318,7 +321,8 @@ private void addConnectorForListener(HttpConfiguration httpConfiguration,
       connector.setName(listener.getName());
     }
 
-    connectors.add(connector);
+    metricsListenerConnectors.add(connector);
+    rateLimitListenerConnectors.add(connector);
```
trnguyencflt, line 325: I think this doesn't work because it is still one connector (i.e same connector in two list).

@mattwong949 another approach for this is to merge `MetricsListener` and `RateLimitNetworkTrafficListener` into one class, wdyt?

core/src/main/java/io/confluent/rest/RateLimitMetricsListener.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * With Jetty 11+, the NetworkTrafficServerConnector has been updated to only accept one
+ * NetworkTrafficListener per instance. This class merges the MetricsListener and
+ * RateLimitNetworkTrafficListener into one class to be used as a single NetworkTrafficListener
+ */
+public class RateLimitMetricsListener implements NetworkTrafficListener {
+
+  private final MetricsListener metricsListener;
+  private final RateLimitNetworkTrafficListener rateLimitNetworkTrafficListener;
```
trnguyencflt, line 31: Since both MetricsListener and RateLimitNetworkTrafficListener implements NetworkTrafficListener, 

we can make this class to be some name like CombineNetworkTrafficListener, that take a list of NetworkTrafficListener as argument for the constructor, stored in private property (delegates), this will make the implementation is more general so that we can extend it to more NetworkTrafficListener in the future, @mattwong949 wdyt?

trnguyencflt (approved):

PR #519, opened 2024-12-17
[DGS-18780] Ensure p99 metrics are correctly published

pritishatx:
We noticed that the percentile latency metrics in rest-utils (request-latency-95 and request-latency-99) are capped out at 10s (code [ref](https://github.com/confluentinc/rest-utils/blob/7.6.x/core/src/main/java/io/confluent/rest/metrics/MetricsResourceMethodApplicationListener.java#L325)). 

The PR makes PERCENTILE_MAX_LATENCY_IN_MS configurable in RestConfig, with 10s as the default value.

[Slack thread](https://confluent.slack.com/archives/C3226NBLJ/p1713808819756109)

### DIFF
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -194,8 +194,8 @@ public class RestConfig extends AbstractConfig {
       + " request latency meets or violates SLA";
   protected static final long METRICS_LATENCY_SLA_MS_DEFAULT = 50;
   public static final String PERCENTILE_MAX_LATENCY_MS_CONFIG = "percentile.max.latency.ms";
-  protected static final String PERCENTILE_MAX_LATENCY_MS_DOC = "The threshold (in ms) of percentile"
-          + " maximum latency";
+  protected static final String PERCENTILE_MAX_LATENCY_MS_DOC = "The threshold (in ms) of"
+          + " percentile maximum latency";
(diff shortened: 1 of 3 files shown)


mattwong949 (approved):
LGTM

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -761,6 +765,12 @@ private static ConfigDef incompleteBaseConfigDef() {
             METRICS_LATENCY_SLA_MS_DEFAULT,
             Importance.LOW,
             METRICS_LATENCY_SLA_MS_DOC
+        ).define(
+            PERCENTILE_MAX_LATENCY_MS_CONFIG,
+            Type.DOUBLE,
+            PERCENTILE_MAX_LATENCY_MS_DEFAULT,
```
trnguyencflt, line 771: [minor] we should have put the validation here to avoid non positive value

pritishatx: Sure, let me get back with a fix for this.

trnguyencflt: thanks @pritishatx

pritishatx: Hi @trnguyencflt, raised a [PR](https://github.com/confluentinc/rest-utils/pull/531) to add the validation.

PR #524, opened 2025-01-13
Backport http2 support changes to 7.5.x

rsrivastav31:
**Changes:** These changes address the Bouncy Castle issue described [here](https://confluentinc.atlassian.net/wiki/spaces/OAAC/pages/3950773390/Bouncy+Castle+test+failure+analysis).  

**Reason for Change:** The changes fix an ALPN Processor error that occurs when Bouncy Castle jars are used with HTTP/2. The fix was originally implemented [here](https://github.com/confluentinc/rest-utils/pull/477) and is now being backported to the 7.5.x branch in this PR.


trnguyencflt:
@rsrivastav31 you don't need to make the new fips-tests directory for the cherry-pick, the reason is that  the change isn't complete, so tests in fips-tests are not actually run. I would suggest to leave only the change in `core/` in this cherry-pick.

rsrivastav31:
> @rsrivastav31 you don't need to make the new fips-tests directory for the cherry-pick, the reason is that the change isn't complete, so tests in fips-tests are not actually run anyway, and it becomes a half-baked module. I would suggest to keep only the change in `core/` in this cherry-pick.

Makes sense. I'll remove the folder and push.

trnguyencflt (approved):
LGTM, thanks 👍

PR #545, opened 2025-01-31
DGS-19645: support to configure server with supplied thread pool

akhileshm1:
Allow supplying a thread pool when creating the application server.


trnguyencflt:
@akhileshm1 I'm not sure why we need this while we can configure the thread pool from RestConfig?

akhileshm1:
> @akhileshm1 I'm not sure why we need this while we can configure the thread pool from RestConfig?

you're right. this is not needed. thanks!

PR #546, opened 2025-02-07
Allow attaching NetworkTrafficListener to connectors of unnamed application

trnguyencflt:
One of the downstream projects uses a single application with null listener name (unnamed app), this allows them to configure the app with multiple unnamed listeners, for example `listeners="http://...:8080,https://...:443"` because the check for equality pass 

```
Objects.equals(connector.getName(), appListenerName)
```

However, [when they changed](https://github.com/confluentinc/cc-security-decisions/pull/1402/files) to use named listeners,`listeners="unsecured://...:8080,secure://...:443"`, this doesn't work any more because the application is unnamed why the listeners have name.

This change relaxes the rule for unnamed app.

**Note that:** using unnamed app needs to be aware that the configured NetworkTrafficListener are combined on all listeners, for example, if it is `MetricsListener`, then the metrics will be accounted for all of the listeners of the app.

### DIFF
--- a/core/src/main/java/io/confluent/rest/ApplicationServer.java
+++ b/core/src/main/java/io/confluent/rest/ApplicationServer.java
@@ -135,12 +135,16 @@ public final class ApplicationServer<T extends RestConfig> extends Server {
     return config.getBoolean(RestConfig.HSTS_HEADER_ENABLE_CONFIG);
   }
 
-  private void attachListener(RestConfig appConfig,
-                              String listenerName,
-                              Metrics metrics,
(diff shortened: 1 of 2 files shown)


dragonmushu (commented):
Thanks @trnguyencflt. Would we be able to back port this change to 7.9?

msn-tldr (approved):

trnguyencflt:
> Thanks @trnguyencflt. Would we be able to back port this change to 7.9?

yes, of course

PR #549, opened 2025-02-10
DGS-19675 Add config for using forward request customizer

kcorman0:
Add config option to use forwardRequestCustomizer. This will be used in IP filtering for SR as we have to propagate IP address when followers forward requests to leaders.


core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -540,6 +540,13 @@ public class RestConfig extends AbstractConfig {
       "Maximum buffer size for jetty request headers in bytes";
   protected static final int MAX_REQUEST_HEADER_SIZE_DEFAULT = 8192;
 
+  protected static final String NETWORK_FORWARDED_REQUEST_ENABLE_CONFIG =
+          "network.forwarded.request.enable";
```
akhileshm1, line 544:
```suggestion
          "network.forwarded.request.customizer.enable";
```

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -540,6 +540,13 @@ public class RestConfig extends AbstractConfig {
...
+  protected static final String NETWORK_FORWARDED_REQUEST_ENABLE_DOC =
+          "If true, forwarded requests will propagate the original IP address "
```
akhileshm1, line 546: nit: this is more about setting the remoteAddr correctly based on the request headers.

akhileshm1 (commented):
what about tests?

akhileshm1 (approved):

silvanli-confluent (approved):
LGTM

PR #552, opened 2025-03-06
[KNET-17653]Jetty migration 9 to 12: rest-utils

apurushotama:
This PR implements the changes for updating Jetty from 9 to 12. 
The PR uses Jakarta specs, JavaEE 10 and Jetty 12 as per the migration documents.

**Migration documents:** 
https://jetty.org/docs/jetty/12/programming-guide/migration/94-to-10.html 
https://jetty.org/docs/jetty/12/programming-guide/migration/11-to-12.html 

**Notes:** 
1. Since Jakarta EE 9 the jakarta namespace is introduced as a replacement for javax namespace from Java EE. Hence, I have used the Jakarta specs instead of the Javax ones ([source](https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest3x/migration.html)) 
2. Updated the following handler classes to use Jetty 12 API ([source](https://jetty.org/docs/jetty/12/programming-guide/migration/11-to-12.html#api-changes-handler)).
    -   core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java 
    -   core/src/main/java/io/confluent/rest/handlers/SniHandler.java
    -   core/src/main/java/io/confluent/rest/handlers/SniUtils.java

     Jetty 11 handler methods 
    `Handler.handle(String target, Request jettyRequest, HttpServletRequest request, HttpServletResponse response)` 
      have been changed in Jetty 12 to:
    `Handler.handle(Request request, Response response, Callback callback)`
3. DosFilter changes	
     -  https://github.com/confluentinc/rest-utils/pull/552#discussion_r1989663825 
     - We had our Jetty servlets pinned to v9.4.53 since the new version removed the external-id setting that our GlobalDosFilter relied on. 

5. AuthUtil.java changes	
Changes have been made to replicate the behavior of setAuthenticate, setRoles, setName used in Jetty 9 to ensure correct authorization levels are set for constraints based on different scenarios. ([source](https://javadoc.jetty.org/jetty-12/org/eclipse/jetty/security/Constraint.Authorization.html))
6. `SHUTDOWN_GRACEFUL_MS_DEFAULT` has been bumped to 5000 to pass the tests that were failing due to timeout when stopping the server.


core/pom.xml
```diff
@@ -17,7 +17,8 @@
     <dependencies>
         <dependency>
             <groupId>org.eclipse.jetty.websocket</groupId>
-            <artifactId>javax-websocket-server-impl</artifactId>
+            <artifactId>websocket-javax-server</artifactId>
+            <version>10.0.0</version>
```
apurushotama, line 21: Need to revisit

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -52,7 +52,7 @@
 import java.util.stream.Collectors;
 import javax.servlet.DispatcherType;
 import javax.servlet.Filter;
-import javax.servlet.ServletException;
+import jakarta.servlet.ServletException;
```
apurushotama, line 55: Either this or [QuietServletException.java](https://github.com/jetty/jetty.project/blob/jetty-12.0.x/jetty-ee10/jetty-ee10-servlet/src/main/java/org/eclipse/jetty/ee10/servlet/QuietServletException.java) if we want it to be less verbose

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -413,9 +412,9 @@ public Handler configureHandler() {
     configurePostResourceHandling(context);
     context.addServlet(defaultHolder, "/*");
 
-    RequestLogHandler requestLogHandler = new RequestLogHandler();
-    requestLogHandler.setRequestLog(requestLog);
-    context.insertHandler(requestLogHandler);
```
apurushotama, line 419: Since I moved the request log setup to server level, we no longer need to insert the RequestLogHandler into the context.

core/src/main/java/io/confluent/rest/JettyDosFilterMultiListener.java
```diff
@@ -17,8 +17,7 @@
 package io.confluent.rest;
 
 import java.util.List;
-import javax.servlet.http.HttpServletRequest;
-
+import org.eclipse.jetty.server.Request;
```
apurushotama, line 20: Why not use this directly?
import jakarta.servlet.http.HttpServletRequest;

core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java
```diff
@@ -37,8 +37,8 @@ public ExpectedSniHandler(List<String> expectedSniHeaders) {
 
   @Override
```
apurushotama, line 38: Fix this override to work with Handler.Wrapper

core/src/main/java/io/confluent/rest/handlers/SniHandler.java
```diff
@@ -19,21 +19,21 @@
...
+import org.eclipse.jetty.server.Handler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class SniHandler extends HandlerWrapper {
+public class SniHandler extends Handler.Wrapper {
   private static final Logger log = LoggerFactory.getLogger(SniHandler.class);
 
   @Override
   public void handle(String target, Request baseRequest,
```
apurushotama, line 34: Fix this override to work with Handler.Wrapper

core/src/main/java/io/confluent/rest/jetty/DoSFilter.java
```diff
@@ -22,39 +20,35 @@
...
-import org.eclipse.jetty.util.log.Logger;
+import org.eclipse.jetty.util.thread.AutoLock;
 import org.eclipse.jetty.util.thread.ScheduledExecutorScheduler;
 import org.eclipse.jetty.util.thread.Scheduler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
- * Denial of Service filter (copied from Jetty 9.4.53.v20231009)
+ * Denial of Service filter (copied from Jetty 12.0.x)
```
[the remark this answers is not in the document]
apurushotama: Interesting. Thanks for pointing this out.

pom.xml
```diff
@@ -177,40 +177,68 @@
...
-                <artifactId>jetty-jaas</artifactId>
+                <artifactId>jetty-security</artifactId>
                 <version>${jetty.version}</version>
             </dependency>
             <dependency>
                 <groupId>org.eclipse.jetty.websocket</groupId>
-                <artifactId>javax-websocket-server-impl</artifactId>
-                <version>${jetty.version}</version>
+                <artifactId>websocket-javax-server</artifactId>
+                <version>10.0.0</version>
```
msn-tldr, line 197: version are defined in `common` ex - `${jetty.version}`,  is there a reason to use an explicit version `10.0.0` here? 

This applies here.

core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java
```diff
@@ -36,17 +34,18 @@ public ExpectedSniHandler(List<String> expectedSniHeaders) {
...
+      log.warn("No SNI header present on request; request URI is {}", baseRequest.getHttpURI());
     } else if (!expectedSniHeaders.contains(sniServerName)) {
       log.warn("SNI header {} is not in the configured list of expected headers {}; "
-          + "request URI is {}", sniServerName, this.expectedSniHeaders, request.getRequestURI());
+          + "request URI is {}", sniServerName, this.expectedSniHeaders, baseRequest.getHttpURI());
     }
 
-    super.handle(target, baseRequest, request, response);
+    super.handle(baseRequest, response, callback);
+    return true;
```
msn-tldr, line 49: should this be `return super.handle(baseRequest, response, callback);`

core/src/main/java/io/confluent/rest/handlers/SniHandler.java
```diff
@@ -18,29 +18,30 @@
...
     String sniServerName = SniUtils.getSniServerName(baseRequest);
     if (sniServerName != null && !sniServerName.equals(serverName)) {
       log.debug("Sni check failed, host header: {}, sni value: {}", serverName, sniServerName);
-      baseRequest.setHandled(true);
-      response.sendError(MISDIRECTED_REQUEST.getCode(), MISDIRECTED_REQUEST.getMessage());
+      Response.writeError(baseRequest, response, callback,
+              MISDIRECTED_REQUEST.getCode(), MISDIRECTED_REQUEST.getMessage());
     }
-    super.handle(target, baseRequest, request, response);
+    super.handle(baseRequest, response, callback);
```
msn-tldr, line 43: should this be `return super.handle(baseRequest, response, callback);` ? callback will be handled by `super.handle`
see https://jetty.org/docs/jetty/12/programming-guide/server/http.html#handler-impl-filter

core/src/main/java/io/confluent/rest/handlers/SniUtils.java
```diff
@@ -28,9 +28,10 @@
 
 public class SniUtils {
   public static String getSniServerName(Request baseRequest) {
-    EndPoint endpoint = baseRequest.getHttpChannel().getEndPoint();
```
msn-tldr, line 31: @bob-barrett can you sanity check this logic? This might impact the v4 everywhere logic. 
Context: rest-utils is being moved to jetty 12 by @apurushotama, bunch of APIs are changing from jetty 9 to 12 versions.

PS - Looks alright at high level.

core/src/main/java/io/confluent/rest/jetty/DoSFilter.java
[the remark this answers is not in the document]
msn-tldr: +1 good catch.

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
```diff
@@ -141,16 +141,17 @@ private static ConstraintMapping createConstraint(
       final boolean authenticate,
       final String pathSpec
   ) {
-    final Constraint constraint = new Constraint();
-    constraint.setAuthenticate(authenticate);
+
+    final Constraint.Builder constraint = new Constraint.Builder();
+    //constraint.setAuthenticate(authenticate);
```
msn-tldr, line 146: I am unable to find any doc which tells if its ok to remove `setAuthenticate(true)`, did you come across any such?

msn-tldr: Looking around a bit, is the equivalent `Authorization::SPECIFIC_ROLE` to be set in the builder?
https://javadoc.jetty.org/jetty-12/org/eclipse/jetty/security/Constraint.Authorization.html#SPECIFIC_ROLE

core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -329,22 +329,23 @@ public Handler configureHandler() {
...
       log.info("Binding {} to all listeners.", this.getClass().getSimpleName());
     }
 
     ServletHolder defaultHolder = new ServletHolder("default", DefaultServlet.class);
     defaultHolder.setInitParameter("dirAllowed", "false");
 
-    ResourceCollection staticResources = getStaticResources();
-    if (staticResources != null) {
-      context.setBaseResource(staticResources);
+    Collection<Resource> staticResources = getStaticResources();
```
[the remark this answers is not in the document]
apurushotama: > Set the base resource to serve content from for this context, which must exist and be readable when the context is started. 

I assumed we need to set a single baseResource. 
Reference: https://javadoc.jetty.org/jetty-12/org/eclipse/jetty/server/handler/ContextHandler.html#setBaseResource(org.eclipse.jetty.util.resource.Resource)

core/src/main/java/io/confluent/rest/handlers/SniUtils.java
```diff
@@ -28,9 +28,10 @@
...
-    if (endpoint instanceof SslConnection.DecryptedEndPoint) {
-      SSLSession session = ((SslConnection.DecryptedEndPoint) endpoint)
+    SslConnection connection = (SslConnection) baseRequest.getConnectionMetaData().getConnection();
+    EndPoint endpoint = connection.getEndPoint();
+    if (endpoint instanceof SslConnection.SslEndPoint) {
+      SSLSession session = ((SslConnection.SslEndPoint) endpoint)
```
bob-barrett, line 34: I think we can just do connection.getSslEndPoint() here, since we already have an `SslConnection`. If we're worried about accounting for non-SSL connections, I think we need to check the type before casing to `SslConnection` on line 31.

bob-barrett: Also, once we have the `SslConnection`, do we even need to get the `EndPoint`? Looks like we just call `endpoint.getSslConnection()`, can we just do `SSLSession session = connection.getSslEngine().getSession()`?

core/src/main/java/io/confluent/rest/handlers/SniHandler.java
apurushotama: Yes, makes sense.

core/src/main/java/io/confluent/rest/handlers/SniHandler.java
```diff
@@ -18,29 +18,28 @@
...
 
   @Override
-  public void handle(String target, Request baseRequest,
-      HttpServletRequest request,
-      HttpServletResponse response) throws IOException, ServletException {
-    String serverName = request.getServerName();
+  public boolean handle(Request baseRequest,
+                     Response response,
+                     Callback callback) throws Exception {
+    String serverName = Request.getServerName(baseRequest);
```
apurushotama, line 36: String serverName = Request.getServerName(baseRequest);
Does this make sense?

core/src/main/java/io/confluent/rest/jetty/DoSFilter.java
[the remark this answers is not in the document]
mattwong949: I've pushed a commit to restore 9.4.53.v20231009 logic

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -95,7 +95,7 @@ public class RestConfig extends AbstractConfig {
   public static final String SHUTDOWN_GRACEFUL_MS_CONFIG = "shutdown.graceful.ms";
   protected static final String SHUTDOWN_GRACEFUL_MS_DOC =
       "Amount of time to wait after a shutdown request for outstanding requests to complete.";
-  protected static final String SHUTDOWN_GRACEFUL_MS_DEFAULT = "1000";
+  protected static final String SHUTDOWN_GRACEFUL_MS_DEFAULT = "5000";
```
mattwong949, line 98: we had a lot of tests failing on the `server.stop()` timeout, and bumping this higher seems to allow them to pass. I'm not feeling great about just bumping a timeout w/o understanding what's going on, but there are other logical issues in other tests to look at before we can revisit this.

msn-tldr: can you add a comment that lower value `1000` led to tests failing ? Probably create a followup if you run out of time to investigate this..

core/src/test/java/io/confluent/rest/ApplicationTest.java
```diff
@@ -180,11 +181,11 @@ public void testCreateSecurityHandlerWithNoRoles() {
 
     ConstraintSecurityHandler securityHandler = new TestApp(config).createBasicSecurityHandler();
     assertEquals(securityHandler.getRealmName(), REALM);
-    assertTrue(securityHandler.getRoles().isEmpty());
+    assertTrue(securityHandler.getConstraintMappings().get(0).getConstraint().getRoles().isEmpty());
     assertNotNull(securityHandler.getLoginService());
     assertNotNull(securityHandler.getAuthenticator());
     assertEquals(1, securityHandler.getConstraintMappings().size());
-    assertFalse(securityHandler.getConstraintMappings().get(0).getConstraint().isAnyRole());
+//    assertFalse(securityHandler.getConstraintMappings().get(0).getConstraint().isAnyRole());
```
msn-tldr, line 188: typo?

core/src/test/java/io/confluent/rest/ApplicationTest.java
```diff
@@ -212,14 +213,14 @@ public void testCreateSecurityHandlerWithSpecificRoles() {
...
+    assertFalse(securityHandler.getConstraintMappings().get(0).getConstraint().getRoles().isEmpty());
     assertNotNull(securityHandler.getLoginService());
     assertNotNull(securityHandler.getAuthenticator());
-    assertEquals(1, securityHandler.getConstraintMappings().size());
+    assertEquals(2, securityHandler.getConstraintMappings().get(0).getConstraint().getRoles().size());
     final Constraint constraint = securityHandler.getConstraintMappings().get(0).getConstraint();
-    assertFalse(constraint.isAnyRole());
-    assertEquals(constraint.getRoles().length, 2);
-    assertArrayEquals(constraint.getRoles(), new String[]{"roleA", "roleB"});
+//    assertFalse(constraint.isAnyRole());
```
msn-tldr, line 221: typo?

core/src/test/java/io/confluent/rest/ApplicationTest.java
```diff
@@ -232,11 +233,11 @@ public void testSetUnsecurePathConstraintsWithUnSecure() {
     final List<ConstraintMapping> mappings = securityHandler.getConstraintMappings();
     assertThat(mappings.size(), is(3));
     assertThat(mappings.get(0).getPathSpec(), is("/*"));
-    assertThat(mappings.get(0).getConstraint().getAuthenticate(), is(true));
+//    assertThat(mappings.get(0).getConstraint().getAuthenticate(), is(true));
```
msn-tldr, line 236: typo?

core/src/test/java/io/confluent/rest/ApplicationTest.java
```diff
@@ -232,11 +233,11 @@ public void testSetUnsecurePathConstraintsWithUnSecure() {
...
     assertThat(mappings.get(1).getPathSpec(), is("/path/1"));
-    assertThat(mappings.get(1).getConstraint().getAuthenticate(), is(false));
+//    assertThat(mappings.get(1).getConstraint().getAuthenticate(), is(false));
```
msn-tldr, line 238: typo?

core/src/test/java/io/confluent/rest/ApplicationTest.java
```diff
@@ -232,11 +233,11 @@ public void testSetUnsecurePathConstraintsWithUnSecure() {
...
     assertThat(mappings.get(2).getPathSpec(), is("/path/2"));
-    assertThat(mappings.get(2).getConstraint().getAuthenticate(), is(false));
+//    assertThat(mappings.get(2).getConstraint().getAuthenticate(), is(false));
```
msn-tldr, line 240: typo?

core/src/test/java/io/confluent/rest/ApplicationTest.java
```diff
@@ -251,8 +252,8 @@ public void testConstraintsWhenOptionsRequestNotAllowedAndSecurityOn() {
     assertThat(mappings.get(0).getPathSpec(), is("/*"));
     assertThat(mappings.get(0).getMethodOmissions(), is(new String[]{"OPTIONS"}));
     assertThat(mappings.get(1).getPathSpec(), is("/*"));
-    assertThat(mappings.get(1).getConstraint().getAuthenticate(), is(true));
-    assertThat(mappings.get(1).getConstraint().isAnyRole(), is(false));
+//    assertThat(mappings.get(1).getConstraint().getAuthenticate(), is(true));
```
msn-tldr, line 255: typo?

core/src/test/java/io/confluent/rest/ApplicationTest.java
```diff
@@ -271,8 +272,8 @@ public void testConstraintsWhenOptionsRequestNotAllowedAndSecurityOff() {
     final List<ConstraintMapping> mappings = securityHandler.getConstraintMappings();
     assertThat(mappings.size(), is(1));
     assertThat(mappings.get(0).getPathSpec(), is("/*"));
-    assertThat(mappings.get(0).getConstraint().getAuthenticate(), is(true));
-    assertThat(mappings.get(0).getConstraint().isAnyRole(), is(false));
+//    assertThat(mappings.get(0).getConstraint().getAuthenticate(), is(true));
```
msn-tldr, line 275: typo?

core/src/test/java/io/confluent/rest/CustomInitTest.java
```diff
@@ -198,12 +200,12 @@ public void configure(final Map<String, ?> config) {
     @Override
     public void accept(final ServletContextHandler context) {
       final List<String> roles = config.getList(RestConfig.AUTHENTICATION_ROLES_CONFIG);
-      final Constraint constraint = new Constraint();
-      constraint.setAuthenticate(true);
-      constraint.setRoles(roles.toArray(new String[0]));
+      final Constraint.Builder constraint = new Constraint.Builder();
+//      constraint.setAuthenticate(true);
```
msn-tldr, line 204: cleanup?

core/src/test/java/io/confluent/rest/HstsIntegrationTest.java
```diff
@@ -60,6 +62,18 @@ public class HstsIntegrationTest {
   private Properties props;
   private File clientKeystore;
 
+  HttpClient httpClient(SslContextFactory.Client sslContextFactory) {
```
msn-tldr, line 65: is there a reason to duplicate `httpClient()` in [ErrorHandlerIntegrationTest.java](https://github.com/confluentinc/rest-utils/pull/552/files/3f66161d8627842ec24f92c5ca181b3f1c43e602..ae462b74e51e58aecd263b8432f11bb1da40b819#diff-9de9ec813d1515d128281c576ab6dde4e5fc3f92e14757325294b2123cb558e4)? you an put the method as static func in utils class ex - `core/src/test/java/io/confluent/rest/TestUtils.java`?

core/src/test/java/io/confluent/rest/Http2Test.java
```diff
@@ -78,6 +80,19 @@ public class Http2Test {
   private static final String SSL_PASSWORD = "test1234";
   private static final String EXPECTED_200_MSG = "Response status must be 200.";
 
+  HttpClient httpClient(SslContextFactory.Client sslContextFactory, HTTP2Client http2Client) {
```
msn-tldr, line 83: is there a reason to duplicate httpClient() in [ErrorHandlerIntegrationTest.java](https://github.com/confluentinc/rest-utils/pull/552/files/3f66161d8627842ec24f92c5ca181b3f1c43e602..ae462b74e51e58aecd263b8432f11bb1da40b819#diff-9de9ec813d1515d128281c576ab6dde4e5fc3f92e14757325294b2123cb558e4)? you an put the method as static func in utils class ex - core/src/test/java/io/confluent/rest/TestUtils.java?

core/src/test/java/io/confluent/rest/RequestLogHandlerIntegrationTest.java
```diff
@@ -110,7 +110,7 @@ public void test_CustomRequestLog_registeredToCorrectListener() throws Exception
     verify(mockLogInternal, times(1)).log(requestCaptor.capture(), responseCaptor.capture());
     // check that external application never logs the request
     verify(mockLogExternal, never()).log(any(), any());
-    assertEquals("127.0.0.1", requestCaptor.getValue().getServerName());
+//    assertEquals("127.0.0.1", requestCaptor.getValue().getServerName());
```
msn-tldr, line 113: typo?

core/src/test/java/io/confluent/rest/SniHandlerIntegrationTest.java
```diff
@@ -67,6 +69,18 @@ public class SniHandlerIntegrationTest {
   private Properties props;
   private File clientKeystore;
 
+  HttpClient httpClient(SslContextFactory.Client sslContextFactory) {
```
msn-tldr, line 72: see comment above, dedup this method.

core/src/test/java/io/confluent/rest/auth/AuthUtilTest.java
```diff
@@ -122,7 +124,7 @@ public void shouldCreateGlobalConstraintWithAuthRequired() {
     final ConstraintMapping mapping = AuthUtil.createGlobalAuthConstraint(config);
 
     // Then:
-    assertThat(mapping.getConstraint().getAuthenticate(), is(true));
+//    assertThat(mapping.getConstraint().getAuthenticate(), is(true));
```
msn-tldr, line 127: typo? more of then in the file

fips-tests/src/test/java/io/confluent/rest/Http2FipsTest.java
```diff
@@ -75,6 +77,19 @@ class Http2FipsTest {
   private String SERVER_CERT;
   private String SERVER_KEY;
 
+  HttpClient httpClient(SslContextFactory.Client sslContextFactory, HTTP2Client http2Client) {
```
msn-tldr, line 80: dedup

core/src/main/java/io/confluent/rest/RestConfig.java
mattwong949: sounds good I'll add a comment in a later commit

core/src/main/java/io/confluent/rest/handlers/SniHandler.java
msn-tldr: Do the test pass with the change? Test should be the source of truth, otherwise looks ok from docs.

core/src/main/java/io/confluent/rest/handlers/SniUtils.java
msn-tldr: @bob-barrett reviewed separately, closing this.

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
apurushotama: Addressed in this commit. https://github.com/confluentinc/rest-utils/pull/552/commits/2803ca1b7cab8baa46cfd6a3aca9d054812ef153

core/src/test/java/io/confluent/rest/ApplicationTest.java
apurushotama: Addressed in this commit: https://github.com/confluentinc/rest-utils/pull/552/commits/2803ca1b7cab8baa46cfd6a3aca9d054812ef153#r1996324258
I had originally commented these out to revisit later

core/src/main/java/io/confluent/rest/errorhandlers/NoJettyDefaultStackTraceErrorHandler.java
```diff
@@ -1,41 +0,0 @@
...
-import org.eclipse.jetty.ee10.servlet.ServletContextRequest;
-import org.eclipse.jetty.http.HttpStatus;
-import jakarta.servlet.http.HttpServletRequest;
-import jakarta.servlet.http.HttpServletResponse;
-import org.eclipse.jetty.ee10.servlet.ErrorHandler;
-
-public class NoJettyDefaultStackTraceErrorHandler extends ErrorHandler {
-  public NoJettyDefaultStackTraceErrorHandler() {
-    super();
-    setShowServlet(false);
```
mattwong949, line 18: this `setShowServlet` is not used in Jetty 12 org.eclipse.jetty.server.handler.ErrorHandler . We replace it with  `setShowCauses(!suppressErrors);` and `setShowStacks(!suppressErrors);` in `StackTraceErrorHandler.java`

core/src/main/java/io/confluent/rest/errorhandlers/NoJettyDefaultStackTraceErrorHandler.java
```diff
@@ -1,41 +0,0 @@
...
-  }
-
-  @Override
-  protected void generateAcceptableResponse(ServletContextRequest baseRequest,
-                                            HttpServletRequest request,
-                                            HttpServletResponse response,
-                                            int code, String message) throws IOException {
-    // set Exception to null to avoid exposing stack trace to clients
-    request.setAttribute(RequestDispatcher.ERROR_EXCEPTION, null);
```
mattwong949, line 25: in Jetty 12 org.eclipse.jetty.server.handler.ErrorHandler , we don't need this because this is prevented by `setShowStacks` in `StackTraceErrorHandler.java`

core/src/test/java/io/confluent/rest/Http2Test.java
```diff
@@ -203,17 +203,17 @@ public void testHttp2AmbiguousSegment() throws Exception {
 
       // Just skip HTTP/2 for earlier than Java 11
       if (ApplicationServer.isJava11Compatible()) {
-        statusCode = makeGetRequestHttp2(HTTP_URI + "/test%2fambiguous%2fsegment");
+        statusCode = makeGetRequestHttp2(HTTP_URI + "/test?segment=%2Fambiguous%2Fsegment");
```
apurushotama, line 206: Based on [this](https://github.com/jetty/jetty.project/issues/11890#issuecomment-2156442947), we cannot pass a ambiguous / (encoded as either %2F or %252F) as part of the path section of the URI over HTTP any more with Servlet 6 (or anything in ee10+, including REST).
Based on Point 10. [Rejecting Suspicious Sequences](https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0#uri-path-canonicalization), any encoded URI of the form  /foo%2Fbar will be rejected with 400. 

I guess it is expected to throw a 400 if it is included in the path section of the URI.
However, the test is ensuring that URI-encoded / characters work in URIs in all variants, not sure if we want to use this workaround https://github.com/jetty/jetty.project/issues/11448#issuecomment-1969206031

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
```diff
@@ -141,16 +141,27 @@ private static ConstraintMapping createConstraint(
...
       final List<String> roles = restConfig.getList(RestConfig.AUTHENTICATION_ROLES_CONFIG);
-      constraint.setRoles(roles.toArray(new String[0]));
+      if (roles.isEmpty()) {
+        //equivalent of setting an empty <auth-constraint> if no setRoles(String []) is set,
+        // forbidding access
+        constraint.authorization(Constraint.FORBIDDEN.getAuthorization());
+      } else if (roles.size() == 1 && roles.get(0).equals("*")) {
+        constraint.authorization(Constraint.KNOWN_ROLE.getAuthorization());
+      } else {
+        constraint.roles(roles.toArray(new String[0]));
```
msn-tldr, line 157: is `Authorization::SPECIFIC_ROLE` implied implicitly since jetty has this as the default value? If so i would prefer to make it explicit by setting here.

msn-tldr (approved):

msn-tldr (changes requested):
Retracting review, as was mistakenly approved. I merely wanted to feedback without approval, as this is WIP.

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
msn-tldr: This look alright at a high level, but would prefer more eyes on this code as these deals with CP security.

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
apurushotama: Yes, it was by default set to Authorization::SPECIFIC_ROLE. 
Let me add that setting as well.

core/src/main/java/io/confluent/rest/handlers/SniHandler.java
apurushotama: Yes, the tests pass.

core/src/test/java/io/confluent/rest/RequestLogHandlerIntegrationTest.java
mattwong949: https://confluentinc.atlassian.net/browse/KNET-15387 note added

core/pom.xml
```diff
@@ -16,8 +16,19 @@
...
-            <groupId>org.eclipse.jetty.websocket</groupId>
-            <artifactId>javax-websocket-server-impl</artifactId>
+            <groupId>org.eclipse.jetty.ee10.websocket</groupId>
+            <artifactId>jetty-ee10-websocket-jakarta-server</artifactId>
+            <version>${jetty.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.websocket</groupId>
+            <artifactId>jakarta.websocket-api</artifactId>
+            <version>2.1.0</version>
```
brianbushree, line 26: Should these pinned versions go upstream? Into https://github.com/confluentinc/common ?

core/src/test/java/io/confluent/rest/ApplicationServerTest.java
```diff
@@ -26,16 +26,18 @@
...
-import javax.ws.rs.core.MediaType;
-import javax.ws.rs.core.Response;
-import javax.ws.rs.ext.ExceptionMapper;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.Configurable;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.ext.ExceptionMapper;
```
brianbushree, line 40: I assume this is not a compatible change for downstream consumers of this project? What's the testing plan? Have we tested it already?

core/src/main/java/io/confluent/rest/handlers/SniUtils.java
mattwong949:
```
SslConnection connection = (SslConnection) baseRequest.getConnectionMetaData().getConnection();
```
this would've failed bc the connection is actually an HttpConnection. I've updated logic for this in a recent commit

core/src/test/java/io/confluent/rest/ApplicationServerTest.java
apurushotama: Yes, this is going to break the downstream builds. Currently, we have pinned those repos to use the older versions. Once this is merged, we will revert those and the builds will break. Each of those need to be fixed.

brianbushree: Should we open an example PR downstream that points at this branch so we can visualize what all will need to change downstream?

apurushotama: We will get failures like this. https://semaphore.ci.confluent.io/jobs/0ba73dc2-cec6-48b0-a248-94005040a80b 
How do I point the secret-registry repo to use this branch without merging rest-utils?

core/pom.xml
apurushotama: Waiting on this: https://github.com/confluentinc/common/pull/766

core/src/test/java/io/confluent/rest/ApplicationServerTest.java
brianbushree: I suppose you could install the jars locally then use those local JARs while building a downstream service. The PR build wouldn't be relevant but at least we'd get a sense of the code changes in each downstream service.

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
```diff
@@ -141,16 +141,27 @@ private static ConstraintMapping createConstraint(
...
+
     if (authenticate) {
       final List<String> roles = restConfig.getList(RestConfig.AUTHENTICATION_ROLES_CONFIG);
-      constraint.setRoles(roles.toArray(new String[0]));
+      if (roles.isEmpty()) {
+        //equivalent of setting an empty <auth-constraint> if no setRoles(String []) is set,
+        // forbidding access
+        constraint.authorization(Constraint.FORBIDDEN.getAuthorization());
+      } else if (roles.size() == 1 && roles.get(0).equals("*")) {
+        constraint.authorization(Constraint.KNOWN_ROLE.getAuthorization());
```
ehumber, line 154: What was the equivalent of this in the old code?  I don't see the * anywhere there.

core/src/main/java/io/confluent/rest/jetty/DoSFilter.java
```diff
@@ -1238,7 +1233,7 @@ public OverLimit isRateExceeded(long now) {
             // rate limit is never exceeded, but we keep track of the request timestamps
             // so that we know whether there was recent activity on this tracker
             // and whether it should be expired
-            synchronized (this) {
+            try (AutoLock l = _lock.lock()) {
```
ehumber, line 1236: Why not stick with the synchronized block?

core/src/main/java/io/confluent/rest/jetty/DoSFilter.java
```diff
@@ -1025,7 +1016,10 @@ public boolean addWhitelistAddress(@Name("address") String address) {
 
     private boolean addWhitelistAddress(List<String> list, String address) {
         address = address.trim();
-        return address.length() > 0 && list.add(address);
+        if (address.length() <= 0)
```
ehumber, line 1019: Any reason to update this logic for the jetty change?

core/src/test/java/io/confluent/rest/SaslTest.java
```diff
@@ -93,7 +93,7 @@ public void setUp(@TempDir Path tmpFolder) throws Exception {
     Path loginPropertiesFile = tmpFolder.resolve("login.properties");
 
     String jaas = "c3 {\n"
-                  + "  org.eclipse.jetty.jaas.spi.PropertyFileLoginModule required\n"
+                  + "  org.eclipse.jetty.security.jaas.spi.PropertyFileLoginModule required\n"
```
ehumber, line 96: Are our users going to have to change this jaas string too in CP?

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
apurushotama: `KNOWN_ROLE` is equivalent to Servlet role "*".  
This would basically fit the AllRoles() case. You can find the in-linecomments in this [commit](https://github.com/confluentinc/rest-utils/pull/552/commits/2803ca1b7cab8baa46cfd6a3aca9d054812ef153) that explains this

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
```diff
@@ -141,16 +141,27 @@ private static ConstraintMapping createConstraint(
...
+    final Constraint.Builder constraint = new Constraint.Builder();
+
     if (authenticate) {
       final List<String> roles = restConfig.getList(RestConfig.AUTHENTICATION_ROLES_CONFIG);
-      constraint.setRoles(roles.toArray(new String[0]));
+      if (roles.isEmpty()) {
+        //equivalent of setting an empty <auth-constraint> if no setRoles(String []) is set,
+        // forbidding access
+        constraint.authorization(Constraint.FORBIDDEN.getAuthorization());
+      } else if (roles.size() == 1 && roles.get(0).equals("*")) {
```
prabhashkr, line 153:
```suggestion
      } else if (roles.size() == 1) {
        if (roles.get(0).equals("*")) {
          constraint.authorization(Constraint.KNOWN_ROLE.getAuthorization());
        } else if (roles.get(0).equals("**")) {
          constraint.authorization(Constraint.ANY_USER.getAuthorization());
        }
```

We are using "**" to bypass Jetty Authorization (as we depend on MDS for authorization.
@apurushotama can you please make the suggested changes. Thanks!

core/src/main/java/io/confluent/rest/jetty/RequestLogHandler.java
```diff
@@ -0,0 +1,4 @@
+package io.confluent.rest.jetty;
```
devhritwik, line 1: The build is failing as this doesn't have the copyright header at the start (checkstyle issue)

apurushotama: We don't need this file anymore.

core/src/test/java/io/confluent/rest/ApplicationServerTest.java
apurushotama: Thank! We built the pipeline locally to fix all the downstream repos.

core/src/main/java/io/confluent/rest/errorhandlers/StackTraceErrorHandler.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2022 Confluent Inc.
```
msn-tldr, line 2: nit: 2025

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
```diff
@@ -163,21 +163,38 @@ private static ConstraintMapping createConstraint(
     return mapping;
   }
 
+  private static void configureAuthentication(Constraint.Builder constraint,
+                                              RestConfig restConfig) {
+    final List<String> roles = restConfig.getList(RestConfig.AUTHENTICATION_ROLES_CONFIG);
```
msn-tldr, line 168: @apurushotama has this been tested that security functionality works?

msn-tldr:
@apurushotama shared feedback on testing required to merge changes. Need to validate security changes, make sure soak  & perf testing is OK.

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
ehumber: If I understand right, the CP code doesn't go via this path, the confluent-security-plugins overrides the security code https://confluent.slack.com/archives/C082S55LMKL/p1742407345260459?thread_ts=1742249229.647539&cid=C082S55LMKL

core/src/main/java/io/confluent/rest/auth/AuthUtil.java
prabhashkr: LGTM! from CP Security point of view. Thanks!

msn-tldr (approved):
LGTM, @apurushotama thanks for pushing on this, this was a big migration. @apurushotama confirmed offline this has gone through CP/CP-security verification as well.
Approving to unblock, lets merge when the CI is green and perf testing is done.

core/src/main/java/io/confluent/rest/jetty/DoSFilter.java
mattwong949: I think this must have been pulled from Jetty 9 changes https://github.com/jetty/jetty.project/commit/de862c558b1b2cccf03d1f17f72a811ef8e18734#diff-8cf68b792fa7e539e2ad0c49fe7bcdb78953fff6cdbf26529c0d122dd6816bf3R1097-R1104

functionally, it's the exact same. `list.add` always returns `true`

core/src/test/java/io/confluent/rest/SaslTest.java
mattwong949: yeah I think so.. we should make a note for the release

core/src/main/java/io/confluent/rest/jetty/DoSFilter.java
mattwong949: we were trying to find some overlap w/ the Jetty 12 utils. It seems Jetty 12 decided to use this AutoLock class throughout. This DosFilter class takes the base logic of the Jetty 9.4.53 version, but makes changes such that it is compatible w/ Jetty 12

aaddepalli7 (approved):

PR #553, opened 2025-03-07
Override jetty version to 9.4.57.v20241219

mattwong949:
https://github.com/confluentinc/common/commit/60322b8531dd93a1afc765080a43c3cb13392223 updated common repo to Jetty 12, our changes for rest-utils for Jetty 12 are not yet finished. Until they are, we will pin Jetty version to jetty 9 so that builds can succeed

our pom does not specify all of the jetty modules, so we need to override the ones we use. I scanned for them with
```
mvn dependency:tree | grep jetty | grep 12.0
```
and then added all of them w/ the overridden version to the pom in rest-utils-parent

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -195,6 +195,41 @@
             </dependency>
             <dependency>
                 <groupId>org.eclipse.jetty</groupId>
+                <artifactId>jetty-alpn-client</artifactId>
+                <version>${jetty.override.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.eclipse.jetty</groupId>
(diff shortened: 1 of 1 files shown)


apurushotama (approved):

apurushotama (approved):

mattwong949:
there is a failure downstream in ksqldb because they have a override in the pom using the common repo https://github.com/confluentinc/ksql/blob/master/ksqldb-rest-app/pom.xml#L153-L157

I will make a change there and inform the KSQL team. but this PR is good to merge

PR #555, opened 2025-03-20
DGS-19733 Add configs for SNI check and Implement PrefixSniHandler for prefix validation

zhaochun-ma:
1. Added new configuration option `prefix.sni.check.enabled` (default: false) to control tenant prefix validation for SNI, if enabled, it will take precedence over the existing [SniHandler](https://github.com/confluentinc/rest-utils/blob/b18f2abd7bed1d613214017037b9839f86958abf/core/src/main/java/io/confluent/rest/handlers/SniHandler.java#L54) when `sni.check.enabled` is also enabled
2. Implemented TenantPrefixSniHandler that:
   - Extracts tenant ID from SNI hostname (e.g., "lsrc-123" from "lsrc-123.us-east-1.aws.private.confluent.cloud")
   - Validates that the Host header starts with the same tenant ID
   - Returns 421 Misdirected Request if validation fails
3. Modified Application class to use TenantPrefixSniHandler when tenant.prefix.sni.check.enabled=true

4. Added new configuration option `sni.host.check.enabled` (default: true)  for SNI host checking in SecureRequestCustomizer.

Context:
https://confluentinc.atlassian.net/wiki/spaces/DG/pages/4119692883/Post+AWS+GA+CCN+Routing+for+SR+-+Design+Note#Data-Plane
CCN routing terminates TLS within the CCN network before forwarding the request to the regional ingress edge, while still operating at Layer 4 (L4). The Server Name Indication (SNI) is overwritten with the correct multi-tenant V4 endpoint, while the Layer 7 (L7) headers and metadata remain unchanged to ensure audit logging functions without modification. 

secureRequestCustomizer is from jetty and rest-util always [configure](https://github.com/confluentinc/rest-utils/blob/c19acb954e7b899ad91debed31d6117372e4b4c0/core/src/main/java/io/confluent/rest/ApplicationServer.java#L275-L283) it on the https listener. Due to SNI rewrite, requests routed from CCN will not pass the SNI validation in secureRequestCustomizer running on PSRC.
```
curl  https://lsrc-stgcp1pp3k-netstgc39j420.us-west-2.aws.glb.stag.cpdev.cloud/
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1"/>
<title>Error 400 Host does not match SNI</title>
</head>
<body><h2>HTTP ERROR 400 Host does not match SNI</h2>
<table>
<tr><th>URI:</th><td>/</td></tr>
<tr><th>STATUS:</th><td>400</td></tr>
<tr><th>MESSAGE:</th><td>Host does not match SNI</td></tr>
</table>

</body>
</html>
```

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -30,7 +30,7 @@ import io.confluent.rest.exceptions.JsonParseExceptionMapper;
 import io.confluent.rest.extension.ResourceExtension;
 import io.confluent.rest.filters.CsrfTokenProtectionFilter;
 import io.confluent.rest.handlers.SniHandler;
-import io.confluent.rest.handlers.TenantPrefixSniHandler;
+import io.confluent.rest.handlers.PrefixSniHandler;
 import io.confluent.rest.metrics.Jetty429MetricsDosFilterListener;
 import io.confluent.rest.metrics.JettyRequestMetricsFilter;
 import io.confluent.rest.metrics.MetricsResourceMethodApplicationListener;
@@ -419,11 +419,9 @@ public abstract class Application<T extends RestConfig> {
     context.insertHandler(requestLogHandler);
 
     // Add SNI validation handler if enabled
-    boolean sniEnabled = config.getSniCheckEnable();
-    boolean tenantPrefixEnabled = config.getTenantPrefixSniCheckEnable();
-    if (tenantPrefixEnabled) {
-      context.insertHandler(new TenantPrefixSniHandler());
-    } else if (sniEnabled) {
+    if (config.getPrefixSniCheckEnable()) {
+      context.insertHandler(new PrefixSniHandler());
+    } else if (config.getSniCheckEnable()) {
       context.insertHandler(new SniHandler());
     }
 
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -93,7 +93,6 @@ import org.glassfish.jersey.server.validation.ValidationFeature;
 import org.glassfish.jersey.servlet.ServletContainer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import io.confluent.rest.RestConfig;
 
 /**
  * A REST application. Extend this class and implement setupResources() to register REST
@@ -421,7 +420,8 @@ public abstract class Application<T extends RestConfig> {
 
     // Add SNI validation handler if enabled
     boolean sniEnabled = config.getBoolean(RestConfig.SNI_CHECK_ENABLED_CONFIG);
-    boolean tenantPrefixEnabled = config.getBoolean(RestConfig.TENANT_PREFIX_SNI_CHECK_ENABLED_CONFIG);
+    boolean tenantPrefixEnabled = config.getBoolean(
+        RestConfig.TENANT_PREFIX_SNI_CHECK_ENABLED_CONFIG);
     if (tenantPrefixEnabled) {
       context.insertHandler(new TenantPrefixSniHandler());
     } else if (sniEnabled) {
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -417,7 +418,13 @@ public Handler configureHandler() {
     requestLogHandler.setRequestLog(requestLog);
     context.insertHandler(requestLogHandler);
 
-    if (config.getSniCheckEnable()) {
+    // Add SNI validation handler if enabled
+    boolean sniEnabled = config.getBoolean(RestConfig.SNI_CHECK_ENABLED_CONFIG);
```
GunalKupta, line 422: Nit: Why not use `config.getSniCheckEnable()`?

karthikeyanas: We can use getTenantPrefixSniCheckEnable() below.

core/src/main/java/io/confluent/rest/handlers/TenantPrefixSniHandler.java
```diff
@@ -0,0 +1,73 @@
...
+      // Extract the tenantID from the sniServerName, which is always the first segment before '.'
+      // Example: "lsrc-123.us-east-1.aws.private.confluent.cloud" → "lsrc-123"
+      String tenantID = getFirstPart(sniServerName);
+      // The logical cluster ID should also appear at the start of the hostHeader.
+      // It may be followed by either a dot (.) or a dash (-).
+      // hostHeader format examples:
+      // - "lsrc-123-$dom<slug>.us-east-1.aws.glb.confluent.cloud"
+      // - "lsrc-123.$dom<slug>.us-east-1.aws.aws.confluent.cloud"
+      if (tenantID == null
+              || !(hostHeader.startsWith(tenantID + DOT_SEPARATOR)
```
karthikeyanas, line 54: can't we just do hostHeader.startsWith(tenantID)?

core/src/main/java/io/confluent/rest/handlers/TenantPrefixSniHandler.java
```diff
@@ -0,0 +1,73 @@
...
+                  || hostHeader.startsWith(tenantID + DASH_SEPARATOR))) {
+        log.debug("SNI prefix check failed, host header: {}, sni tenantId: {}, full sni: {}",
```
karthikeyanas, line 56: can be a warn()

karthikeyanas (commented):
Will do another round reviewing the unit tests again.

core/src/main/java/io/confluent/rest/handlers/TenantPrefixSniHandler.java
zhaochun-ma: I added separate to disallow,  say `lsrc-123.us-east-1.aws.private.confluent.cloud` to match `lsrc-1234.$dom<slug>.us-east-1.aws.aws.confluent.cloud` in case different Lsrc ID have same prefix

wcheng92:
The PR LGTM. Left a minor comment. Thanks!

core/src/main/java/io/confluent/rest/handlers/TenantPrefixSniHandler.java
```diff
@@ -0,0 +1,73 @@
...
+        baseRequest.setHandled(true);
+        response.sendError(MISDIRECTED_REQUEST.getCode(), MISDIRECTED_REQUEST.getMessage());
+        return;
+      }
+    }
+    super.handle(target, baseRequest, request, response);
+  }
+
+  private static String getFirstPart(String hostname) {
+    if (hostname == null) {
```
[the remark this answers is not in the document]
zhaochun-ma: hostname.split("\\.") would create a new array of Strings. I feel it causes overhead when I just want the part before the first dot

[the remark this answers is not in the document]
wcheng92: Yeah I agree. I was mainly thinking to combine the conditions that we want to return `null` but the current code also works.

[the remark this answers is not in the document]
wcheng92: Resolving this conversation. Feel free to change it or leave it as it is now. Thanks!

wcheng92 (approved):

trnguyencflt:
@zhaochun-ma please also do pint-merge once you merge the PR, thanks 👍

karthikeyanas (approved):

PR #563, opened 2025-04-07
Add configuration hsts.header.enable to control whether we response w…

lei-tianyu:
…ith HSTS header (#491)

* Add configuration hsts.header.enable to control whether we response with HSTS header

* Add integration test to show that HSTS header exists in response

* Update doc to false by default

merge this change from master to branch 7.7.x


trnguyencflt (approved):
LGTM, please also do pint-merge from 7.7.x to master after this

PR #568, opened 2025-04-21
Remove zk notice for cp 8.0.x release

silvanli-confluent:


shubh-ranade:
Should we remove the ZooKeeper reference in https://github.com/confluentinc/rest-utils/blob/master/licenses/NOTICE.confluent-common.txt ?

PR #574, opened 2025-05-20
Carl/s2 s 3829

carlhejiayu:


core/src/test/java/io/confluent/rest/SpireTest.java
carlhejiayu, line 1: This file is NOT fully READY for CI, for now, the file can only be ran if we do the following
1. Locally spin up docker spire server/agent
2. Locally spin up the helloWorldApplication.
In order for this file to be fully ready, we need to automate the two step above into the test file.

core/src/test/java/io/confluent/rest/SpireTest.java
```diff
@@ -0,0 +1,74 @@
...
+        String response = callServer("https://localhost:8080/hello?name=fff");
+
+        
+        assertTrue(response.contains("Hello, fff!"));
+    }
+
+    private String callServer(String serverUrl) throws Exception {
+        DefaultX509Source.X509SourceOptions x509SourceOptions = DefaultX509Source.X509SourceOptions
+                .builder()
+                .spiffeSocketPath("tcp://127.0.0.1:31523")
```
carlhejiayu, line 35: Please ignore the details of this test file; once we agree , then i will make this properly.

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -95,10 +100,32 @@ static boolean isJava11Compatible() {
...
+
+      DefaultX509Source.X509SourceOptions x509SourceOptions = DefaultX509Source.X509SourceOptions
+              .builder()
+              .spiffeSocketPath(spiffeSocketPath)
+              .svidPicker(list -> list.get(list.size() - 1))
+              .build();
+      return DefaultX509Source.newSource(x509SourceOptions);
+    } catch (Exception e) {
+      throw new RuntimeException("Failed to initialize SPIFFE X509 source", e);
+    }
```
carlhejiayu, line 125: if we initialize in applicationServer, we have to think about multiple apps ; how will they interact/use this X509 Source? Or should we move to app level?

core/src/main/java/io/confluent/rest/SslFactory.java
```diff
@@ -159,6 +157,27 @@ public static SslContextFactory createSslContextFactory(SslConfig sslConfig) {
...
   }
 
+  private static void configureSpiffeSslContext(
+      SslContextFactory.Server sslContextFactory,
+      X509Source x509Source,
+      boolean isMtls) {
+    SpiffeSslContextFactory.SslContextOptions options = SpiffeSslContextFactory.SslContextOptions
+        .builder()
+        .x509Source(x509Source)
+        .acceptAnySpiffeId()
```
carlhejiayu, line 167: Right now, I configured to `acceptAnySpiffeId` . The authorization implementation is not part of this PR, it can be done in the following PR.

core/src/main/java/io/confluent/rest/ApplicationServer.java
```diff
@@ -95,10 +99,18 @@ static boolean isJava11Compatible() {
...
   public ApplicationServer(T config) {
-    this(config, createThreadPool(config));
+    this(config, createThreadPool(config), null);
+  }
+
+  public ApplicationServer(T config, X509Source x509Source) {
+    this(config, createThreadPool(config), x509Source);
+  }
+
+  public ApplicationServer(T config,ThreadPool threadPool) {
```
silvanli-confluent, line 109: nit: whitespace

core/pom.xml
```diff
@@ -15,6 +15,23 @@
...
     <dependencies>
+        <!-- SPIFFE Dependencies -->
+        <dependency>
+            <groupId>io.spiffe</groupId>
+            <artifactId>java-spiffe-core</artifactId>
+            <version>0.8.11</version>
+        </dependency>
+        <dependency>
+            <groupId>io.spiffe</groupId>
+            <artifactId>grpc-netty-macos-aarch64</artifactId>
```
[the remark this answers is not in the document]
trnguyencflt: can you check what has been done in ce-kafka? There might already a better way for this

core/src/main/java/io/confluent/rest/ApplicationServer.java
carlhejiayu: thx. I made this [commit](https://github.com/confluentinc/rest-utils/pull/574/commits/6e48e15ceb60867f03773d610e1f4c3c684564e8) to fix the white space

akhileshm1:
@carlhejiayu, is it possible to make this PR against 7.9.x? we can pint-merge the changes to master. the reason being SR is currently on 7.9.x in CC.

carlhejiayu:
> @carlhejiayu, is it possible to make this PR against 7.9.x? we can pint-merge the changes to master. the reason being SR is currently on 7.9.x in CC.

Yes @akhileshm1 , i created a brand new PR that is based in 7.9.X. see here: https://github.com/confluentinc/rest-utils/pull/583

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -206,6 +206,11 @@ public class RestConfig extends AbstractConfig {
       + " runtime request tags in global stats.";
   protected static final boolean METRICS_GLOBAL_STATS_REQUEST_TAGS_ENABLE_DEFAULT = false;
 
+  public static final String SSL_IS_SPIRE_ENABLED_CONFIG = "ssl.spire.enabled";
```
dragonmushu, line 209: Would named listeners be supported here. Basically we need to also be able to open up multiple ssl ports within same application. One for spire and another maybe that is just tls.

core/src/main/java/io/confluent/rest/SslFactory.java
```diff
@@ -103,27 +106,22 @@ private static FileWatcher.Callback onFileChangeCallback(SslConfig sslConfig,
...
-        Path watchLocation = Paths.get(sslConfig.getReloadOnKeyStoreChangePath());
-        try {
-          FileWatcher.onFileChange(watchLocation,
-              onFileChangeCallback(sslConfig, sslContextFactory));
-          log.info("Enabled SSL cert auto reload for: " + watchLocation);
-        } catch (java.io.IOException e) {
-          log.error("Cannot enable SSL cert auto reload", e);
-        }
-      }
+      configureKeyStore(sslContextFactory, sslConfig);
```
dragonmushu, line 124: Both shouldn't be configured right?

core/src/main/java/io/confluent/rest/SslFactory.java
dragonmushu: Can we not configure a regex pattern here

dragonmushu (commented):
Thanks added some comments and also is there a way we can log mTLS failures

core/src/main/java/io/confluent/rest/SslFactory.java
carlhejiayu: If you look at the native library [here](https://github.com/spiffe/java-spiffe/blob/a6059bdaebccfc118870893aee96af1fa028d991/java-spiffe-provider/src/main/java/io/spiffe/provider/SpiffeSslContextFactory.java#L150), they don't natively support regex. If we want to have more complex pattern, we have to define one more layer abstraction to do this functionality, which is why it won't be this PR scope.

core/src/main/java/io/confluent/rest/SslFactory.java
carlhejiayu: if we have `spire` enabled,  you are right that they won't be needed to be configured; However, even configuring won't do much anything because internally the server will go to talk spire agent to grab all the trust bundle and stuff into cache; it will never have a need to reach the KeyStore even Keystore was configured. I already provided the comment [here](https://github.com/confluentinc/rest-utils/blob/6e48e15ceb60867f03773d610e1f4c3c684564e8/core/src/main/java/io/confluent/rest/RestConfig.java#L211)

core/src/main/java/io/confluent/rest/RestConfig.java
dragonmushu: https://github.com/confluentinc/cc-security-decisions/pull/1402/files

carlhejiayu:
> @carlhejiayu , @moe-omar do we not want to interface with the SRJ standalone module here: https://confluentinc.atlassian.net/wiki/x/HAId6w instead of the raw java library
> 
> Instead of passing in an X509 source can we pass in the SpireX509Client.

Hi @dragonmushu , I would not recommend do this for the following the 2  reason:

1. `SpireX509Client` is an SRJ concept (SRJ is confluent internal repo), we should not import confluent dependency into this public repo `rest-util`; in fact, it is not allowed.
2. I believe SRJ is on process of using `rest-util` to do their implementation. I believe that `X509Source` should be the way to go unless you have a strong reason we should keep `SpireX509Client`

carlhejiayu:
> Also @carlhejiayu one thing I called out in the requirements was observability around the
> 
> > @carlhejiayu , @moe-omar do we not want to interface with the SRJ standalone module here: https://confluentinc.atlassian.net/wiki/x/HAId6w instead of the raw java library
> > Instead of passing in an X509 source can we pass in the SpireX509Client.
> 
> Sorry for these late reviews. It would be good to unify these two libraries because clients are recommended to use the SRJ library to configure client SDKs and so if we can somehow just wrap the SRJ module here for the server side it would be nice. Also we would only try to have one connection to the agent per service.

I believe SRJ would start to rely on `rest-util`

core/src/main/java/io/confluent/rest/RestConfig.java
carlhejiayu: Hey @dragonmushu , in the example link you provided, it looks you have 2 port , 1 is TLS, 1 is unsecured. I am not seeing we have 2 ports for 2 TLS.

PR #583, opened 2025-06-09
S2S-3829 Enable Spire in Rest-Util -base 7.9.x

carlhejiayu:
Description
Enable Server using Spire to do S2S authentication.

Detailed Changes

Introduce new settings ssl.spire.enabled, by default it's false and everything function exactly the same as before; it will only try to do s2s through spire if this is turned on.
Modify application and server constructor to take in X509Source. This argument is optional for non-spire setting. If it's configured to use spire, this argument became Mandatory.
HelloWorldSpireApplication is the sample application with spire setting. We can take this as an example for all library users to follow.
Testing
Most of the testing are done through unit test where we mock X509Source to provide certain bundle and test against the server behaviour with its certain configuration. All the tests are written in SpireTest.java

Server Onboarding Spire Doc
Please look at the following doc to see how existing rest-util server can make config changes to onboard spire
https://confluentinc.atlassian.net/wiki/spaces/SECENG/pages/4470801894/Rest-Util+Server+onboarding+to+Spire

DEMO Video
Here is a demo video on how we can turn on certain setting and make our application doing spire S2S in our local dev machine
https://drive.google.com/file/d/1BXudIXuEIy0ck1QckdoapAMHcox9jU2D/view?usp=sharing

TODO: we can setup the container testing strategy where can spin up spire server/agent in the CI do the testing. But this is also not super trival, wont' be included in the scope of this PR.


core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -199,6 +199,11 @@ public class RestConfig extends AbstractConfig {
       + " runtime request tags in global stats.";
   protected static final boolean METRICS_GLOBAL_STATS_REQUEST_TAGS_ENABLE_DEFAULT = false;
 
+  public static final String SSL_IS_SPIRE_ENABLED_CONFIG = "ssl.spire.enabled";
```
trnguyencflt, line 202: [nit] `SSL_IS_SPIRE_ENABLED` -> `SSL_SPIRE_ENABLED`

core/src/main/java/io/confluent/rest/SslFactory.java
```diff
@@ -159,6 +157,23 @@ public static SslContextFactory createSslContextFactory(SslConfig sslConfig) {
     return sslContextFactory;
   }
 
+  private static void configureSpiffeSslContext(
+      SslContextFactory.Server sslContextFactory,
+      X509Source x509Source) {
+    SpiffeSslContextFactory.SslContextOptions options = SpiffeSslContextFactory.SslContextOptions
+        .builder()
+        .x509Source(x509Source)
+        .acceptAnySpiffeId()
```
trnguyencflt, line 166: what is the implication of this, will there be any security risk?

core/src/main/java/io/confluent/rest/SslFactory.java
```diff
@@ -178,4 +193,33 @@ private static void configureSecurityProvider(Server sslContextFactory, SslConfi
...
+      sslContextFactory.setKeyManagerFactoryAlgorithm(
+          sslConfig.getKeyManagerFactoryAlgorithm());
+    }
+
+    if (sslConfig.getReloadOnKeyStoreChange()) {
+      configureKeyStoreReload(sslContextFactory, sslConfig);
+    }
+  }
+
+  private static void configureKeyStoreReload(
```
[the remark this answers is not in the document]
carlhejiayu: This [comment](https://github.com/confluentinc/rest-utils/pull/574#discussion_r2140487685)  should answered your question

core/src/main/java/io/confluent/rest/RestConfig.java
carlhejiayu: Yes, fix. thank you

core/src/main/java/io/confluent/rest/SslFactory.java
trnguyencflt: Can we please add a JIRA ticket to address this? 

Also, has this design passed the security review?

carlhejiayu: Yeah, it's good idea to add JIRA ticket: https://confluentinc.atlassian.net/browse/S2S-3920

Just want to clarify we are not Introducing any NEW vulnerability here. Based on my understanding, this authorization capabilities are NEW after we introduce Spire-MTLS. In the traditional MTLS, rest-util only did authentication, and I am not seeing any authorization schema in this library (correct me if i am wrong). The reason behind is that `SPIRE` give each services/clients SVID (which is an identity). Because of this, server now can choose to let whom to talk to.

However,  as I previously mentioned, natively it doesn't support complex pattern matching, like regex. If we add to the config to accept `a list of acceptable spiffeID` , which require clients to know in advance what the ID is going t be looking like. In the case SDS, they may not every single identity in advance. This simple authorization schema is probably going to create more troubles than ever. Our intern @jocelyndrean is working on middleware to do this authorization specifically, I strongly suggest that we should wait for the concrete design and then proceed.

core/src/main/java/io/confluent/rest/SslFactory.java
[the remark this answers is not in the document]
trnguyencflt: I just want to make sure that the logic to automatically reload the keystore should still work.

SPIFFE certificate will eventually expire, so there must be a logic in there that reload with new cert.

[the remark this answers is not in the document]
trnguyencflt: fair enough, thanks 👍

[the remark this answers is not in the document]
trnguyencflt: fair enough, could you please add the above explanation to the comment of the code? Thanks 👍

trnguyencflt (approved):
Approved under the condition that https://github.com/confluentinc/rest-utils/pull/583#discussion_r2152568177 is done similarly for other services and there is plan to improve on that later.

trnguyencflt (approved):
LGTM, thanks 👍

PR #593, opened 2025-07-29
KNET-10736: Add tenant rate limiting to alleviate noisy neighbor problem

nabeelnauman:
Please first read through the JIRA (and associated docs) [here](https://confluentinc.atlassian.net/browse/KNET-10736) to see motivation for this change. 

This implementation the suggestion in the JIRA to add tenant rate limiting by creating a new DoS Filter that overrides `extractUserId` with the tenant ID. Note that for both the V3 and V4 network, I was able to get the tenant ID from the path and extracting it from the hostname was not required for V4. But I've left the logic for hostname extraction as a fallback so V4 extraction will still work even if the path somehow doesn't have it.

Some implementation details:
- The order of the rate limiters currently goes from most most granular to least granular as I believe this makes the most sense: IP-based -> tenant-based -> global. Ideally the values of each should also be <= to the next one in the chain. Setting IP-based = tenant-based would allow us to give maximum req/sec available to the tenant, and they could handle specific IP rate limiting on their end. At that point, IP-based rate limiting is more of an individual tenant problem and not relevant to us. 
- For tenant ID extraction, I implemented a simple string matching solution that returns as soon as a match is found. Prior to this, I was taking a regex approach, but given tenant ID parsing is called on each request, performance is important and simple string matching should be more efficient.
- In the event a tenant ID is not recognized, we return null from the `extractUserId` function. That results in the fallback logic [here](https://github.com/confluentinc/rest-utils/blob/97439b296a3546e8e1ad0d5a8d3864a05e06505b/core/src/main/java/io/confluent/rest/jetty/DoSFilter.java#L532). So we fallback to session-based rate limiting if available or IP-address based rate limiting otherwise. I.e. the behavior is similar to what we would have without tenant rate limiting, which I think is the best we can do when tenant identification fails for whatever reason.
- Added config for the actual rate limit and flag to enable/disable tenant rate limiting. Kept the default rate limit at 25 to match the default IP and global rate limits. When tenant rate limiting is disabled, it should behave exactly as it did prior to this change. Will make a subsequent PR to make these both available as LD flags.

This code has been tested to be working in both V3 and V4 clusters. I ran a test to simulate the NNP problem (as described [here](https://confluentinc.atlassian.net/wiki/spaces/~712020196c4f65f87449a1834c0930c2dc26fe/pages/4608655417/Noisy+Neighbor+Problem+-+Repro)). Namely, one tenant from multiple IPs bombarded the cluster with reqs, while a second well-behaved tenant sent a low number of requests. Prior to this change, it was observed that tenant 2 was hit with constant `429`'s in addition to tenant 1. After the change, only tenant 1 experienced `429`'s, thus confirming the tenant rate limiting is working as expected.

I've added unit tests for both the path-based and hostname-based tenant ID extraction logic. I've also added an int test for the new tenant rate limiter, alongside the existing IP-based and global rate limiter tests.

### DIFF
--- a/core/src/main/java/io/confluent/rest/TenantDosFilter.java
+++ b/core/src/main/java/io/confluent/rest/TenantDosFilter.java
@@ -44,7 +44,6 @@ public class TenantDosFilter extends DoSFilter {
     }
 
     HttpServletRequest httpRequest = (HttpServletRequest) request;
-
     String tenantId = TenantUtils.extractTenantId(httpRequest);
     
     if (tenantId.equals(UNKNOWN_TENANT)) {
@@ -52,15 +51,11 @@ public class TenantDosFilter extends DoSFilter {
       // This results in the base DoSFilter to fall back to session-based (if enabled) or
       // IP-based rate limiting as this is the best we can do in this scenario.
       log.warn("Skipping tenant-based rate limiting for unidentified tenant. "
-              + "Request: {} '{}' (Host: '{}')",
+          + "Request: {} '{}' (Host: '{}'), falling back to IP-based rate limiting",
           httpRequest.getMethod(), httpRequest.getRequestURI(), httpRequest.getServerName());
       return null;
     }
 
-    log.debug("Final result - tenant ID: '{}' for request: {} '{}' (Host: '{}')",
-        tenantId, httpRequest.getMethod(), httpRequest.getRequestURI(),
-        httpRequest.getServerName());
-    
     return tenantId;
   }
 }
--- a/core/src/main/java/io/confluent/rest/TenantDosFilter.java
+++ b/core/src/main/java/io/confluent/rest/TenantDosFilter.java
@@ -60,5 +60,10 @@ public class TenantDosFilter extends DoSFilter {
 
     return tenantId;
   }
+
+  @Override
+  protected boolean isTenantBasedTracking() {
+    return true;
+  }
 }
 
--- a/core/src/main/java/io/confluent/rest/TenantDosFilter.java
+++ b/core/src/main/java/io/confluent/rest/TenantDosFilter.java
@@ -38,6 +38,11 @@ public class TenantDosFilter extends DoSFilter {
 
   @Override
   protected String extractUserId(ServletRequest request) {
+    // IMPORTANT: If we can't identify the tenant (or get a bad request), return null to skip
+    // tenant-based rate limiting.
+    // This results in the base DoSFilter to fall back to session-based (if enabled) or
+    // IP-based rate limiting, as this is the best we can do in this scenario.
+
     if (!(request instanceof HttpServletRequest)) {
       log.warn("Request is not an HttpServletRequest, cannot extract tenant ID");
       return null;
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/confluent/rest/TenantUtils.java
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Utility class for extracting tenant IDs from HTTP requests.
+ * Supports both V3 and V4 network for tenant identification.
+ */
+public final class TenantUtils {
+
+  private static final Logger log = LoggerFactory.getLogger(TenantUtils.class);
+
+  // V3 pattern: /kafka/v3/clusters/lkc-xxx/...
+  // TODO: Regex matching is slow, optimize if performance becomes an issue
```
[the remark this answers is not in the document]
nabeelnauman: Will optimize this later as noted in the `TODO` comment and ideally get rid of regex. Just tyring to get it to work for now

core/src/main/java/io/confluent/rest/TenantDosFilter.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    HttpServletRequest httpRequest = (HttpServletRequest) request;
+
+    String tenantId = TenantUtils.extractTenantId(httpRequest);
+    
+    if (tenantId.equals(UNKNOWN_TENANT)) {
+      // If we can't identify the tenant, return null to skip tenant-based rate limiting.
+      // This results in the base DoSFilter to fall back to session-based (if enabled) or
+      // IP-based rate limiting as this is the best we can do in this scenario.
```
mattwong949, line 53: nit: this comment is helpful in understanding the DoSFilter and how to use it by overloading extractUserId. I like how you mention that returning null is the expected case when we want to fall back to session-based rate tracking, but can we put that info at the top of the function? This is because we also return null on L43. I don't want anyone to miss the description of this part of the behavior, since it's pretty important IMO

core/src/test/java/io/confluent/rest/JettyDosFilterMultiListenerIntegrationTest.java
```diff
@@ -60,7 +60,7 @@
  * This test makes sure when DosFilter rejects requests, then configured dosfilter-listeners are
  * run, including the mandatory Jetty429MetricsDosFilterListener.
  */
-@Tag("IntegrationTest")
+// @Tag("IntegrationTest") - Temporarily removed for testing
```
mattwong949, line 63: is this test working? I see in the semaphore build that each of the 4 tests passed all 5 of their runs

mattwong949 (commented):
everything LGTM except for a comment in the JettyDosFilter test. Are we still working on it or is it just a leftover comment?

core/src/test/java/io/confluent/rest/JettyDosFilterMultiListenerIntegrationTest.java
nabeelnauman: Ah thanks, I commented out this test before I updated it with the new tenant logic. This is just a leftover from that. And yeah it seems it's already running on semaphore even without the tag

PR #598, opened 2025-08-06
KNET-19517: Add dry-run mode for tenant rate limiting

nabeelnauman:
This builds on the previous PR (#593) to add a dry-run mode for tenant rate limiting. I'm creating a separate PR for this as this is temporary code just to be used for the initial testing (as mentioned [here](https://confluentinc.atlassian.net/wiki/spaces/~712020196c4f65f87449a1834c0930c2dc26fe/pages/4636246473/Noisy+Neighbor+Problem+in+Multi-Tenant+Clusters#Deployment-%2F-Rollout-Considerations)) and I don't want to obscure the logic of the main PR. 

When in dry-run mode, we don't do any actual tenant rate limiting, but simply observe the results of tenant extraction. This is important to verify in production so we can catch any potential edge cases that result in failed tenant extraction.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -827,11 +827,12 @@ public abstract class Application<T extends RestConfig> {
   // TODO: This is temporary code to be removed after tenant rate limit testing
   private void configureTenantDryRunFilter(ServletContextHandler context) {
     TenantDryRunFilter dryRunFilter = new TenantDryRunFilter();
-    FilterHolder filterHolder = new FilterHolder(dryRunFilter);
+    String tenantLimit = String.valueOf(config.getDosFilterTenantMaxRequestsPerSec());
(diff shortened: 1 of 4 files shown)


mattwong949 (approved):
Nice idea for the dry run filter to pass it through but log the result

nabeelnauman:
Manually tested to be working. Namely, verified that this correctly does NOT apply tenant-based rate limiting (noisy neighbor problem was still observed), that a warning message is emitted for tenants that would have exceeded their tenant-based rate limits (note that this DOESN'T include tenants that were subject to NNP and didn't actually exceed their tenant rate limits but still got 429'd), and that unknown tenant ID logs are emitted when parsing fails.

Log showing noisy tenant exceeding their rate limit (but were not actually tenant-rate-limited in the dry-run):
image

Log showing a bad tenant ID I passed in that correctly logs a warning:
image

Verifying that well-behaved tenants that experience NNP do not get misclassified as those exceeding tenant rate limits:
image

PR #600, opened 2025-08-12
temp: Add a feature flag to disable response size metrics collection

devhritwik:
With this change, the `response size` is registered as 0, but the sensor will still detect the count of the response. 

This is temporary and should be done better using a `Writer Interceptor`, handled in this JIRA here - https://confluentinc.atlassian.net/browse/KNET-19746


nabeelnauman (approved):
LGTM, just left a minor nit

nabeelnauman (approved):

PR #679, opened 2026-04-28
Add config to enforce required SNI headers

bob-barrett:
This PR adds a new config, `reject.invalid.sni.headers`. When this config is set to true, `ExpectedSniHandler` will reject connections with an HTTP 400 error if the SNI header of the incoming connection is missing or does not match the configured list of valid hostnames.

### DIFF
--- a/core/src/main/java/io/confluent/rest/Application.java
+++ b/core/src/main/java/io/confluent/rest/Application.java
@@ -460,7 +460,8 @@ public abstract class Application<T extends RestConfig> {
     } else if (config.getSniCheckEnable()) {
       context.insertHandler(new SniHandler());
     } else if (!expectedSniHeaders.isEmpty()) {
-      context.insertHandler(new ExpectedSniHandler(expectedSniHeaders));
+      context.insertHandler(
+          new ExpectedSniHandler(expectedSniHeaders, config.getRejectInvalidSniHeaders()));
     }
 
     Sequence handlers = new Sequence();
--- a/core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java
+++ b/core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java
@@ -16,6 +16,8 @@
 
 package io.confluent.rest.handlers;
 
+import static org.eclipse.jetty.http.HttpStatus.Code.BAD_REQUEST;
+
 import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.Response;
 import org.eclipse.jetty.server.Handler;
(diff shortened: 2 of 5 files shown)


core/src/main/java/io/confluent/rest/Application.java
```diff
@@ -460,7 +460,8 @@ public Handler configureHandler() {
     } else if (config.getSniCheckEnable()) {
       context.insertHandler(new SniHandler());
     } else if (!expectedSniHeaders.isEmpty()) {
-      context.insertHandler(new ExpectedSniHandler(expectedSniHeaders));
+      context.insertHandler(
+          new ExpectedSniHandler(expectedSniHeaders, config.getRejectInvalidSniHeaders()));
     }
```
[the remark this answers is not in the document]
bob-barrett: This is fine, if no expected headers are configured the rejection config is intended to be a no-op

core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java
```diff
@@ -27,10 +29,18 @@
 
 public class ExpectedSniHandler extends Handler.Wrapper {
   private static final Logger log = LoggerFactory.getLogger(ExpectedSniHandler.class);
+  private static final String INVALID_SNI_MESSAGE = "Invalid SNI";
```
devhritwik, line 32: Is this message sufficient or should we be more descriptive in the error?

devhritwik (approved):

core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java
```diff
@@ -43,6 +53,13 @@ public boolean handle(Request baseRequest,
     } else if (!expectedSniHeaders.contains(sniServerName)) {
       log.warn("SNI header {} is not in the configured list of expected headers {}; "
           + "request URI is {}", sniServerName, this.expectedSniHeaders, baseRequest.getHttpURI());
+    } else {
+      return super.handle(baseRequest, response, callback);
+    }
+    if (rejectInvalidSniHeaders) {
+      Response.writeError(baseRequest, response, callback,
+          BAD_REQUEST.getCode(), INVALID_SNI_MESSAGE);
+      return true;
```
devhritwik, line 67: [TIOLI] This is a bit hard to read, needed me to use a notebook and pen to follow :P 

Is something like this better? 
```
  boolean invalid = sniServerName == null || !expectedSniHeaders.contains(sniServerName);                                                                                                                                                            
  if (!invalid) { 
    return super.handle(baseRequest, response, callback);                                                                                                                                                                                            
  }
  if (sniServerName == null) {                                                                                                                                                                                                                       
    log.warn("No SNI header present on request; request URI is {}", baseRequest.getHttpURI());
  } else {                                                                                                                                                                                                                                           
    log.warn("SNI header {} is not in the configured list of expected headers {}; "
        + "request URI is {}", sniServerName, expectedSniHeaders, baseRequest.getHttpURI());                                                                                                                                                         
  }                                                                                                                                                                                                                                                  
  if (rejectInvalidSniHeaders) {                                                                                                                                                                                                                     
    Response.writeError(baseRequest, response, callback,                                                                                                                                                                                             
        BAD_REQUEST.getCode(), INVALID_SNI_MESSAGE);
    return true;                                                                                                                                                                                                                                     
  }
  return super.handle(baseRequest, response, callback);  
```

core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java
bob-barrett: I think this is fine, at the end of the day if someone hits this they're probably going to open a support ticket regardless of how descriptive we are. A quick internet search shows "Invalid SNI" as a message used by other systems: https://peterobrien.blog/2024/02/29/invalid-sni-what-is-it-and-how-to-fix-it/. So this is at least not totally unprecedented.

core/src/main/java/io/confluent/rest/handlers/ExpectedSniHandler.java
bob-barrett: I don't like the duplication of `super.handle`, but I tried to clean this up a bit, let me know what you think.

devhritwik (approved):

akhileshchg (approved):

PR #693, opened 2026-05-04
[CPBR-3705] Migrate packaging files to nightly-branch pattern (7.5.x)

saurabhcnf:
#### Summary
Migrate RPM, DEB, and archive packaging files from separate `origin/rpm`, `origin/debian`, and `origin/archive` branches into a unified `debian/` directory on the code branch, following the nightly-branch pattern used by schema-registry, common, and other components.

This is a prerequisite for s390x multi-arch RPM packaging support. This will be point-merged till master.

This is the 7.5.x version of the same migration that was on 7.4.x PR in #677 which was cancelled. Cherry-picked from the corresponding `74x-syadav-branch-migration` branch — same end state, rebased onto `7.5.x`. `rest-utils` has version-agnostic packaging (no cross-version differences in `origin/rpm`, `origin/debian`, `origin/archive`), so the 7.4.x commit applies cleanly.

#### What changed
- Added `debian/Makefile` — unified Makefile combining RPM, DEB, and archive targets with `PACKAGE_TYPE`-based PREFIX switching and `debian-control` target for generating `debian/control` from `control.in`
- Added `debian/confluent-rest-utils.spec.in` — RPM spec template (from origin/rpm)
- Added `debian/control.in` — Debian control template with `##VERSION##` placeholder (from origin/debian)
- Added `debian/compat`, `debian/changelog`, `debian/copyright`, `debian/gbp.conf`, `debian/rules`, `debian/source/format` — standard Debian packaging metadata
- Added `debian/patches/series` and `debian/patches/standard_build_layout.patch`

#### Key design decisions
- Install logic from `create_archive.sh` inlined into Makefile `install:` target
- RPM spec generation uses inline `sed` instead of `create_spec.sh`
- Uses `control.in` (not static `control`) with `debian-control:` target to substitute version at build time
- Follows the same unified Makefile pattern as common and schema-registry

### Testing
Companion packaging PR (7.5.x): https://github.com/confluentinc/packaging/pull/2383

#### Artifacts comparison and verification
https://confluentinc.atlassian.net/browse/CPBR-3673?focusedCommentId=5050099

Artifact comparison between migration build and baseline build (both on 7.4.x) confirmed zero functional differences:
- All jar-level differences are explained by ZIP metadata timestamps, MANIFEST.MF Commit-ID, Scala compiler non-determinism, and Maven build timestamps
- RPM, DEB, and archive packages are functionally identical

Re-verification on 7.5.x is not required: this PR cherry-picks the exact same commit from 7.4.x. The packaging file content is identical, and `rest-utils` has no cross-version differences in its legacy packaging branches. Since the 7.4.x verification proved the relocated files produce byte-equivalent packages, the same conclusion holds for 7.5.x.

### Issue
https://confluentinc.atlassian.net/browse/CPBR-3673


rahejaprince (approved):

debian/changelog
```diff
@@ -0,0 +1,5 @@
+confluent-rest-utils (0.1~SNAPSHOT-1) unstable; urgency=low
```
shayelkin, line 1: We currently don't bundle changelog in our deb packages (at least in `confluent-rest-utils_8.2.0-1_all.deb` which I just checked).

If we to start, we should have a real changelog here (or keep without, same as we do now).

debian/Makefile
```diff
@@ -0,0 +1,149 @@
+# Dependencies you'll probably need to install to compile this: make, curl, git,
+# zip, unzip, patch, java7-jdk | openjdk-7-jdk, maven.
+
+SHELL=/bin/bash
+
+ifndef VERSION
+# Note that this is sensitive to this package's version being the first
+# <version> tag in the pom.xml
+VERSION=$(shell grep version pom.xml | head -n 1 | awk -F'>|<' '{ print $$3 }')
```
shayelkin: This is broken: the first match would be the `<?xml version…` header of the pom, not the package version.

debian/Makefile
```diff
@@ -0,0 +1,149 @@
...
+LIBPATH=$(PREFIX)/share/$(PACKAGE_TITLE)
+DOCPATH=$(PREFIX)/share/doc/$(PACKAGE_TITLE)
+
+INSTALL=install -D -m 644
+INSTALL_X=install -D -m 755
+
+install: build
+	rm -rf $(DESTDIR)$(PREFIX)
+	mkdir -p $(DESTDIR)$(PREFIX)
+	mkdir -p $(DESTDIR)$(BINPATH)
```
shayelkin: We don't install anything to `$BINPATH`.

debian/patches/standard_build_layout.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/Makefile b/Makefile
```
shayelkin, line 1: Why do we need this patch?

saurabhcnf: The packaging pipeline (`build-rpm.py`, `build-deb.py`, `build-archive.py` in `confluentinc/packaging`) invokes make <target> from the repo root, but our Makefile lives at `debian/Makefile`. This patch creates a thin root-level Makefile that proxies any target to `debian/Makefile`, so make rpm / make deb / make archive work from the repo root during package builds without committing a Makefile to the working tree (which would confuse Maven/IDEs during normal development). 

It's byte-identical to the same patch used by other components on the nightly-branch pattern. The 2-line diff vs legacy `origin/debian:debian/patches/standard_build_layout.patch` is intentional: we dropped the cp `debian/create_archive.sh` / `rm create_archive.sh` lines because the script no longer exists (its logic moved into debian/Makefile's install: target).

debian/changelog
saurabhcnf: This file is byte-identical to what's on legacy `origin/debian:debian/changelog` (we migrated it as-is) and matches the same placeholder pattern other components on the nightly-branch pattern use. The file is required by `dpkg-buildpackage` to read package version/maintainer metadata at build time but, as you noted, isn't bundled inside the .deb. This isn't something introduced by the migration change.

debian/Makefile
saurabhcnf: You're right — the regex matches `<?xml version="1.0"?>` first, so this fallback would produce an empty `VERSION` if it ever ran. However, this block is gated by ifndef `VERSION`, and the packaging pipeline always exports VERSION from `CP_VERSION` in packaging-run-config.yml before invoking make — so the fallback never executes in CI. The same buggy regex exists identically in legacy `origin/rpm:Makefile` and `origin/debian:debian/Makefile`, and in other components' nightly-pattern Makefiles. 

Will push a fix commit for this.

debian/Makefile
saurabhcnf: This mkdir creates an empty /usr/bin/ because rest-utils ships no scripts there. I checked `debian/confluent-rest-utils.spec.in` %files (lists only `/usr/share/java/rest-utils` and `/usr/share/doc/rest-utils`) and `debian/rules` (uses dh which only ships installed files), so the empty dir does not end up in the .rpm or .deb. 

The line is inherited from legacy `origin/debian:debian/create_archive.sh` (and the `rpm/archive` equivalents) which did `mkdir -p ${DESTDIR}${BINPATH}` identically — so artifacts are unchanged either way. Safe to drop the `$BINPATH`. 

Will push a fix that removes the `mkdir -p $(DESTDIR)$(BINPATH)`

debian/Makefile
```diff
@@ -0,0 +1,148 @@
+# Dependencies you'll probably need to install to compile this: make, curl, git,
+# zip, unzip, patch, java7-jdk | openjdk-7-jdk, maven.
+
+SHELL=/bin/bash
+
+ifndef VERSION
+# Note that this is sensitive to this package's version being the first
+# <version> tag in the pom.xml
+VERSION=$(shell grep '<version>' pom.xml | head -n 1 | awk -F'>|<' '{ print $$3 }')
```
shayelkin: I think the comment in L8-L9 doesn't hold, as the first `<version>` [tag in the pom.xml](https://github.com/confluentinc/rest-utils/blob/master/pom.xml#L11) is the version for the parent (`common`), which is a range. 

Mind testing this Makefile again (making sure `VERSION` isn't defined, of course), just to be certain it works as expected?

I tried to write an regexp that would do slightly better matching,and came up with such a horrible one, when testing it I got 
```
grep: ../rest-utils/pom.xml: exceeded PCRE's backtracking limit
```
So I would highly recommend just using maven if you can:
```
mvn help:evaluate -Dexpression=project.version -q -DforceStdout
```
should output the actual version for the package, avoiding the regex issue all together.

XKCD#1171

debian/Makefile
saurabhcnf: The earlier change had made it worse failed when tested locally. Switched to the suggestion you provided offline to use 
`mvn help:evaluate -Dexpression=project.version -q -DforceStdout` 
It now correctly returns projects own version. Tested in local without VERSION set.

shayelkin (approved):
Thank you!

PR #714, opened 2026-06-11
DGS-24490 - blanket request timeout for SR

tmalikconfluent:
Description
Implement a blanket timeout for requests at the Jetty layer and return a HTTP 500 response to customers if a request takes more than 10 minutes to complete. Also free up the  Jetty thread in this scenario.
This change is off by default and turned on in the SR repo, and should only effect SR. 
PR on the SR side : https://github.com/confluentinc/schema-registry/pull/4373

### DIFF
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -387,6 +387,16 @@ public class RestConfig extends AbstractConfig {
           + "the timeout.";
   public static final long REQUEST_TIMEOUT_MS_DEFAULT = 0;
 
+  public static final String REQUEST_TIMEOUT_INTERRUPT_ENABLE_CONFIG =
+          "request.timeout.interrupt.enable";
+  public static final String REQUEST_TIMEOUT_INTERRUPT_ENABLE_DOC =
+          "When a request exceeds " + REQUEST_TIMEOUT_MS_CONFIG + ", also interrupt the worker "
+          + "thread handling it. This can reclaim a thread blocked on an interruptible operation "
+          + "(e.g. network I/O), but has no effect on CPU-bound work that does not check the "
+          + "interrupt status, nor on non-interruptible blocking calls. Only has an effect when "
+          + REQUEST_TIMEOUT_MS_CONFIG + " is greater than 0.";
+  public static final boolean REQUEST_TIMEOUT_INTERRUPT_ENABLE_DEFAULT = false;
+
   public static final String THREAD_POOL_MIN_CONFIG = "thread.pool.min";
   public static final String THREAD_POOL_MIN_DOC =
           "The minimum number of threads will be started for HTTP Servlet server.";
@@ -1044,6 +1054,12 @@ public class RestConfig extends AbstractConfig {
             REQUEST_TIMEOUT_MS_DEFAULT,
             Importance.LOW,
             REQUEST_TIMEOUT_MS_DOC
+        ).define(
+            REQUEST_TIMEOUT_INTERRUPT_ENABLE_CONFIG,
+            Type.BOOLEAN,
+            REQUEST_TIMEOUT_INTERRUPT_ENABLE_DEFAULT,
+            Importance.LOW,
+            REQUEST_TIMEOUT_INTERRUPT_ENABLE_DOC
         ).define(
             THREAD_POOL_MIN_CONFIG,
             Type.INT,
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -379,6 +379,13 @@ public class RestConfig extends AbstractConfig {
           "The number of milliseconds to hold an idle session open for.";
   public static final long IDLE_TIMEOUT_MS_DEFAULT = 30_000;
 
+  public static final String REQUEST_TIMEOUT_MS_CONFIG = "request.timeout.ms";
+  public static final String REQUEST_TIMEOUT_MS_DOC =
+          "Maximum time in milliseconds a single HTTP request is allowed to take before the "
+          + "server aborts it and returns HTTP 504 Gateway Timeout. Guards against requests that "
+          + "hold a worker thread indefinitely. A value <= 0 (the default) disables the timeout.";
+  public static final long REQUEST_TIMEOUT_MS_DEFAULT = -1;
+
   public static final String THREAD_POOL_MIN_CONFIG = "thread.pool.min";
   public static final String THREAD_POOL_MIN_DOC =
           "The minimum number of threads will be started for HTTP Servlet server.";
@@ -1030,6 +1037,12 @@ public class RestConfig extends AbstractConfig {
             IDLE_TIMEOUT_MS_DEFAULT,
             Importance.LOW,
             IDLE_TIMEOUT_MS_DOC
+        ).define(
+            REQUEST_TIMEOUT_MS_CONFIG,
+            Type.LONG,
+            REQUEST_TIMEOUT_MS_DEFAULT,
+            Importance.LOW,
+            REQUEST_TIMEOUT_MS_DOC
         ).define(
             THREAD_POOL_MIN_CONFIG,
             Type.INT,
--- a/core/src/main/java/io/confluent/rest/RestConfig.java
+++ b/core/src/main/java/io/confluent/rest/RestConfig.java
@@ -383,8 +383,9 @@ public class RestConfig extends AbstractConfig {
   public static final String REQUEST_TIMEOUT_MS_DOC =
           "Maximum time in milliseconds a single HTTP request is allowed to take before the "
           + "server aborts it and returns HTTP 504 Gateway Timeout. Guards against requests that "
-          + "hold a worker thread indefinitely. A value <= 0 (the default) disables the timeout.";
-  public static final long REQUEST_TIMEOUT_MS_DEFAULT = -1;
+          + "hold a worker thread indefinitely. A value of 0 (the default) or less disables "
+          + "the timeout.";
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/confluent/rest/handlers/RequestTimeoutHandler.java
```diff
@@ -0,0 +1,93 @@
...
+
+  @Override
+  public boolean handle(Request request, Response response, Callback callback) throws Exception {
+    // Shared between the timeout task and the wrapped callback so that whichever fires first
+    // wins and the request is completed exactly once.
+    final AtomicBoolean done = new AtomicBoolean(false);
+
+    final Scheduler.Task timeoutTask = request.getComponents().getScheduler().schedule(() -> {
+      if (done.compareAndSet(false, true)) {
+        log.warn("Request to {} exceeded the configured timeout of {} ms; returning 504",
```
DhirajSuri27: Nit: Can we use `HttpStatus.GATEWAY_TIMEOUT_504` in the log?

core/src/main/java/io/confluent/rest/RestConfig.java
```diff
@@ -379,6 +379,13 @@ public class RestConfig extends AbstractConfig {
           "The number of milliseconds to hold an idle session open for.";
   public static final long IDLE_TIMEOUT_MS_DEFAULT = 30_000;
 
+  public static final String REQUEST_TIMEOUT_MS_CONFIG = "request.timeout.ms";
+  public static final String REQUEST_TIMEOUT_MS_DOC =
+          "Maximum time in milliseconds a single HTTP request is allowed to take before the "
+          + "server aborts it and returns HTTP 504 Gateway Timeout. Guards against requests that "
+          + "hold a worker thread indefinitely. A value <= 0 (the default) disables the timeout.";
+  public static final long REQUEST_TIMEOUT_MS_DEFAULT = -1;
```
DhirajSuri27: Nit: Can the default be `0` in that case?

core/src/test/java/io/confluent/rest/RequestTimeoutIntegrationTest.java
```diff
@@ -0,0 +1,154 @@
...
+      latch = new CountDownLatch(1);
+    }
+
+    static void release() {
+      latch.countDown();
+    }
+
+    @GET
+    @Path("/slow")
+    public String slow() throws InterruptedException {
```
DhirajSuri27, line 197: Nit: Can we have a test to simulate and validate behavior in cases of non-blocking operations that block the server thread (e.g., an I/O bound operation that takes too long to complete)?

DhirajSuri27 (approved):

core/src/main/java/io/confluent/rest/handlers/RequestTimeoutHandler.java
```diff
@@ -0,0 +1,93 @@
...
+            request.getHttpURI(), timeoutMs);
```
DhirajSuri27: Nit: Is there any way to record the actual request duration?

DhirajSuri27 (approved):

DhirajSuri27 (approved):

DhirajSuri27 (approved):

core/src/main/java/io/confluent/rest/handlers/RequestTimeoutHandler.java
```diff
@@ -0,0 +1,112 @@
+/*
+ * Copyright 2024 Confluent Inc.
```
trnguyencflt: 2026

core/src/test/java/io/confluent/rest/RequestTimeoutIntegrationTest.java
```diff
@@ -0,0 +1,227 @@
+/*
+ * Copyright 2024 Confluent Inc.
```
trnguyencflt: 2026

core/src/main/java/io/confluent/rest/handlers/RequestTimeoutHandler.java
```diff
@@ -0,0 +1,112 @@
...
+ * the stuck worker thread (Java has no safe way to do so). The thread is reclaimed only once its
+ * downstream call returns or its own socket/idle timeout trips.
+ *
+ * <p>When {@code interruptOnTimeout} is enabled, the handler additionally interrupts the worker
+ * thread on timeout. This can reclaim a thread blocked on an <em>interruptible</em> operation
+ * (e.g. network I/O such as leader forwarding, {@code sleep}, lock/wait); it has <em>no</em>
+ * effect on CPU-bound work that never checks the interrupt status, nor on non-interruptible
+ * blocking calls. Enable it only where the long-running work is known to be interruptible.
+ */
+public class RequestTimeoutHandler extends Handler.Wrapper {
```
trnguyencflt, line 52: [minor] is there any existing Handler in jetty library that can do this?

core/src/main/java/io/confluent/rest/handlers/RequestTimeoutHandler.java
```diff
@@ -0,0 +1,112 @@
...
+  public boolean handle(Request request, Response response, Callback callback) throws Exception {
+    // Shared between the timeout task and the wrapped callback so that whichever fires first
+    // wins and the request is completed exactly once.
+    final AtomicBoolean done = new AtomicBoolean(false);
+    final long startNanos = System.nanoTime();
+    // The handler chain runs synchronously on this thread, so it is the worker thread that will
+    // execute (and potentially block in) the downstream request handling.
+    final Thread handlingThread = Thread.currentThread();
+
+    final Scheduler.Task timeoutTask = request.getComponents().getScheduler().schedule(() -> {
```
trnguyencflt: is the handling thread part of the request.getComponents().getScheduler()?

If so we might be in a scenario where the task is scheduled into the handling thread which is currently blocking processing the request.

trnguyencflt: this might also get schedule rejection exception, how do we deal with this?

core/src/main/java/io/confluent/rest/handlers/RequestTimeoutHandler.java
```diff
@@ -0,0 +1,112 @@
...
+  private final long timeoutMs;
+  private final boolean interruptOnTimeout;
+
+  public RequestTimeoutHandler(long timeoutMs, boolean interruptOnTimeout) {
+    this.timeoutMs = timeoutMs;
+    this.interruptOnTimeout = interruptOnTimeout;
+  }
+
+  @Override
+  public boolean handle(Request request, Response response, Callback callback) throws Exception {
```
trnguyencflt, line 65: please also write unit tests for this

core/src/main/java/io/confluent/rest/handlers/RequestTimeoutHandler.java
```diff
@@ -0,0 +1,112 @@
...
+    final Scheduler.Task timeoutTask = request.getComponents().getScheduler().schedule(() -> {
+      if (done.compareAndSet(false, true)) {
+        long elapsedMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startNanos);
+        log.warn("Request to {} exceeded the configured timeout of {} ms (elapsed {} ms); "
+                + "returning {}{}",
+            request.getHttpURI(), timeoutMs, elapsedMs, HttpStatus.GATEWAY_TIMEOUT_504,
+            interruptOnTimeout ? " and interrupting the worker thread" : "");
+        if (interruptOnTimeout) {
+          // Nudge the worker to abort; only effective for interruptible operations.
+          handlingThread.interrupt();
```
trnguyencflt: if this throw exception then we will not be able to write 504 response back.

core/src/main/java/io/confluent/rest/handlers/RequestTimeoutHandler.java
tmalikconfluent: Jetty's built-in timeout mechanisms don't cover this case. a worker thread blocked deep in a synchronous resource method never returns to the container, so the async timeout never starts counting

core/src/main/java/io/confluent/rest/handlers/RequestTimeoutHandler.java
tmalikconfluent: ScheduledExecutorScheduler delegates to its own ScheduledExecutorService backed by a dedicated scheduler thread (the class holds its own Thread), completely separate from the QueuedThreadPool worker threads. So the timeout task fires on the scheduler thread even while the worker is blocked, whivch is the intended design

core/src/main/java/io/confluent/rest/handlers/RequestTimeoutHandler.java
tmalikconfluent: Moved Response.writeError(...) to run before handlingThread.interrupt(), and wrapped the interrupt in a try/catch that logs instead of propagating. Now nothing can prevent the timeout response from reaching the client.

core/src/main/java/io/confluent/rest/handlers/RequestTimeoutHandler.java
tmalikconfluent: added

core/src/main/java/io/confluent/rest/handlers/RequestTimeoutHandler.java
tmalikconfluent: Wrapped the scheduler.schedule(...) call in a try/catch. If the scheduler rejects the task, it logs a warning and falls back to super.handle(request, response, callback). The request is handled normally without timeout protection rather than failing outright.

trnguyencflt (approved):
LGTM

PR #728, opened 2026-08-18
Skip TLS validation for non-SPIRE client certs

DhirajSuri27:
Skip TLS validation for non-SPIRE client certs

### DIFF
--- a/core/src/main/java/io/confluent/rest/SpireOptionalTrustManager.java
+++ b/core/src/main/java/io/confluent/rest/SpireOptionalTrustManager.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright 2026 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/confluent/rest/SpireOptionalTrustManager.java
```diff
@@ -0,0 +1,168 @@
...
+    if (chain == null || chain.length == 0) {
+      return false;
+    }
+    Collection<List<?>> subjectAlternativeNames;
+    try {
+      subjectAlternativeNames = chain[0].getSubjectAlternativeNames();
+    } catch (CertificateParsingException e) {
+      log.debug("Failed to parse SAN entries while checking for a SPIFFE URI SAN; "
+          + "treating certificate as non-SPIFFE", e);
+      return false;
```
tobiogunbi, line 146: nit: should we be swallowing on parse failures like this?

tobiogunbi (approved):
lgtm but minor nit

xuliang2019 (approved):

PR #733, opened 2026-09-16
Add SPIRE SPIFFE-ID allowlist trust manager for full-SPIRE listeners

nloyapally-cflt:
## Summary

Adds an optional SPIFFE-ID allowlist for SPIRE-enabled TLS listeners. On a full-SPIRE
listener (`ssl.spire.enabled=true`, `ssl.spire.trust.only.enabled=false`) configured with a
non-empty `ssl.spire.accepted.spiffe.id.patterns`, the SPIFFE trust manager is wrapped to
additionally require the client SVID's `spiffe://` URI to fully match one of the configured
regex patterns, and per-listener validation counters are recorded. The allowlist defaults to
empty, in which case behavior is unchanged (any SVID that chains to the SPIRE bundle is
accepted), so existing listeners are unaffected.

## What changed

- **`RestConfig`** — new config `ssl.spire.accepted.spiffe.id.patterns` (`LIST`, default empty).
- **`SslConfig`** — `getAcceptedSpiffeIdPatterns()` accessor.
- **`SslFactory`** — when the allowlist is non-empty, build the `SSLContext` explicitly and
  wrap the SPIFFE trust manager in `SpireSpiffeAllowlistTrustManager`; empty allowlist keeps the
  existing java-spiffe path untouched.
- **`ApplicationServer`** — build one `SslContextFactory` per listener and pass lazily-resolved,
  per-listener metrics suppliers (resolved at first handshake, after applications register).
- **`SpireSpiffeAllowlistTrustManager`** (new) — enforces chain-to-bundle (delegated) plus the
  SPIFFE-ID regex allowlist; fails closed; records four counters in the `spire` group:
  `spire-handshake-success-total`, `spire-handshake-failure-total`,
  `spire-x509-source-unavailable-total`, `spiffe-allowlist-reject-total`.

## Compatibility

Default-empty allowlist means no behavior change for any existing listener. The new trust
manager is only installed on a full-SPIRE listener that configures a non-empty allowlist.

## Testing

- Unit tests (JDK 17): `SpireSpiffeAllowlistTrustManagerTest`, `SslConfigTest`, `SslFactoryTest`,
  `RestConfigTest`, `ApplicationServerTest` — covering SAN/SPIFFE-ID extraction, full-match
  allowlist logic, the validation flow and failure logging, counter recording (including the
  first-handshake case), lazy metrics resolution, config parsing/defaults, and the SslFactory
  allowlist branch.
- `checkstyle` clean on `src/main`.


core/src/main/java/io/confluent/rest/SslFactory.java
```diff
@@ -181,30 +208,65 @@ public static SslContextFactory createSslContextFactory(
...
+        SSLContext sslContext = SpiffeSslContextFactory.getSslContext(options);
+        sslContextFactory.setSslContext(sslContext);
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+      return;
+    }
 
+    // Allowlist configured: server SVID from SPIRE + SPIFFE trust wrapped with allowlist + metrics.
+    log.info("SPIRE SSL allowlist enforcement enabled ({} accepted SPIFFE-ID pattern(s))",
```
Zhongyin-Zhang: log.debug?

DhirajSuri27 (commented):
LGTM from an cross-service impact perspective. This will not affect SPIRE-enabled default behavior, or SPIRE Trust-only behavior at all. So, we should be good on that front 👍

core/src/main/java/io/confluent/rest/SpireSpiffeAllowlistTrustManager.java
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  // ---------------------------------------------------------------------------
+  // Helpers
+  // ---------------------------------------------------------------------------
+
+  /** Reject the caller unless its SVID's SPIFFE ID matches the configured allowlist. */
+  private void enforceSpiffeAllowlist(X509Certificate[] chain) throws CertificateException {
+    if (acceptedSpiffeIdPatterns.isEmpty()) {
+      // No allowlist configured: chain-to-bundle validation only.
```
Zhongyin-Zhang, line 218: NIT, lets add a debug level log here to record enforceSpiffeAllowlist is passed because the acceptedSpiffeIdPatterns is empty

core/src/main/java/io/confluent/rest/SpireSpiffeAllowlistTrustManager.java
```diff
@@ -0,0 +1,367 @@
...
+    }
+    String spiffeId = (chain != null && chain.length > 0) ? spiffeIdOf(chain[0]) : null;
+    if (!spiffeIdAllowed(spiffeId, acceptedSpiffeIdPatterns)) {
+      record(() -> allowlistRejectSensor);
+      // WARN: a validly-chained SVID that fails the allowlist is a misconfiguration or an
+      // unexpected/unauthorized caller -- worth alerting on. Do not log the whole certificate.
+      log.warn("Rejecting client: SPIFFE ID [{}] is not in the accepted allowlist", spiffeId);
+      throw new CertificateException(
+          "SPIFFE ID " + spiffeId + " is not in the accepted-identity allowlist");
+    }
```
Zhongyin-Zhang, line 230: NIT: can we add a debug level log here to record the which spiffeId is passed validation?

Zhongyin-Zhang (approved):
Thanks @nloyapally-cflt the PR overall LGTM. Leave couple NIT comments.

core/src/main/java/io/confluent/rest/SpireSpiffeAllowlistTrustManager.java
nloyapally-cflt: added

core/src/main/java/io/confluent/rest/SpireSpiffeAllowlistTrustManager.java
nloyapally-cflt: added

Zhongyin-Zhang (approved):

